PR #6, opened 2017-03-21
Add .editorconfig file & Improve some stuffs for contributions

emagnier:
Hi @pejulian 

I'm plan to send a PR with French translations very soon, + some bugfixes.
But first I suggest a PR containing some codebase improvement.

First commit:
- Added an [.editorconfig](http://editorconfig.org/) file (EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs).
- Trim the trailing spaces, this generally help a lot to compare different versions of codebase, and when multiple persons are editing the same part of code.
- No code is modified.

Second commit:
- Fix a failing test. Probably because of an update of the moment library (or its sub components). By making the strings lowercase, I think it will avoid similar cases in the future.

Thanks for this awesome module!
It's well written and there is a lot of tests 🍺


pejulian:
Hi @emagnier,

Thank you for your contributions.
It has been some time since I have actively worked on it and an update is definitely welcome.

.editorconfig
```diff
@@ -0,0 +1,11 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
tukusejssirs: Is there any need for final newline insertion? I think that they should be removed and this setting should be set to `false`. If anyone agrees, I can create a PR and then we could merge it.

emagnier: I think that one is very subjective.
The example on editorconfig.org has this option set to `true`, and I have been working on a few project with it, without any inconvenience.

So I don't see the advantage of changing this option to `false`.
Furthermore changing that will bring an update of almost all code files, which is IMHO not necessary when things are already consistent.

tukusejssirs: Okay, I don’t argue, though I personally prefer keep it `false`. And I know that [editorcong.org](editorconfig.org] has it that way and IMO it is the main reason everyone default to that. And I don’t any advantage of keeping a final newline either.

Anyway, keep it the way it is. :)

PR #8, opened 2017-04-25
Compute correctly the week numbers in Lent

emagnier:
During lent, weeks was starting on Wednesday instead of Sunday.

This fix remove the first 4 days of lent (Ash Wednesday to Saturday) when computing the week number.


emagnier:
The travis build is failing because it use an old version of Node that don't recognise the latest `es6` features.

@pejulian, what do you think about updating the `.travis.yml` with a node version `>=` 6 ?

pejulian:
Hi @emagnier, I've checked the code and saw the problem mentioned. The weeks were being incremented incorrectly. I've merged your fix and updated `travis.yml` to include later node versions. Thank you.

PR #87, opened 2019-12-01
Update HISTORY.md and bump to v.1.2.4

emagnier:


HISTORY.md
```diff
@@ -1,10 +1,22 @@
 # Version history
 
-## 1.2.4 /2017-03-22
-* Added French translations for France, Canada and Belgium.
-* Updated French Calendar
+## 1.2.4
```
tukusejssirs: @emagnier, did you really modify old history? Why? ;)

tukusejssirs: Were we not bumping to ~`v1.3.4`~ `v1.3.1`?

emagnier: Because there was a mistake, the French translations has been added in `v1.2.3`, not `v1.2.4`.

And we are here on `master`, not on `romcal-es6`. It's important to release/publish the last things done on `master` before merging soon `romcal-es6` which gonna be a big update.

tukusejssirs: Okay then, but I think we should try to unify the histories of both branches and these _ad hoc_ commits to `master` (without committing to `romcal-es6`) are not the way to go.

emagnier: Nope, it is the way to go :) Commits here are not related to the work done on `romcal-es6`.
It's precisely my goal to stabilize `romcal-es6` (including fixing the critical bug related to locales) so we can rebase / merge it to `master`, publish a new version, and delete `romcal-es6`.

tukusejssirs: Well, I know there are many Git workflows, but I am still inclined to prefer the workflow described [here](https://nvie.com/posts/a-successful-git-branching-model/) (see image below), which I call the _standard_ Git workflow.

Git workflow

emagnier: We already had discussion about that and I still think it's too overkill for this sized project. Lean on git tags and release management from npmjs.org is more than enough.

tukusejssirs: Okay.

PR #94, opened 2019-12-04
Locales for Italy

emagnier:
With the permission of @gtortone, I'm opening here a commit with his big work of translation 🇮🇹.
Thanks @gtortone for this amazing work! 🎉

This PR is open on the `romcal-es6` branch, since there is currently some refactors in progress on locales and keys. It will be easier to take care of the new Italian locales in the same time.

I've also added a commit, to fix some duplicated keys. But this highlight that we need to find a more standard way to write them. @tukusejssirs you suggested somewhere to remove saint/blessed to simplify key names (that would make also reading easier in the locales files). It's something that we should consider and move forward on that.


gtortone:
Hi @emagnier,
I just realized the problem of duplicated keys is true because I started from a translation that did not include these saints... Thanks !

emagnier:
@gtortone you are right, new keys have been added in our working branch. No problem :) And thanks!

Issue #88, opened 2019-12-02
Rename `weekday` to `feria` (rollback)


tukusejssirs:
I suggest to replace `week day` by `feria`, as `feria` is proper name for this. `WEEKDAY` is currently used in `Types.js`.

This issue is somewhat related to #73.

emagnier:
This request has been addressed and merged. Thanks!

tukusejssirs:
It might have been a wrong decision to rename `weekday` to `feria`. I’ve just found the [_General Instruction of the Roman Missal_](https://www.nlo.org.nz/assets/Uploads/NZ-GIRM-Final.pdf) (on the website of National Liturgy Office website of the Catholic Church in Aotearoa New Zealand) from 2010 Roman Missal and it never uses the term _feria_ and always uses the term _weekday_.

emagnier:
This is the same from this doc: https://www.liturgyoffice.org.uk/Resources/GIRM/Documents/GIRM.pdf

As long as romcal v2 is still not out, this is still the good time to improve and change the things.

emagnier:
Addressed on https://github.com/romcal/romcal/pull/250.

PR #100, opened 2019-12-05
Rename weekday to feria

emagnier:
This is basically a find and replace in the codebase.

- Feria is more relevant than Weekday, and is the the exact term in the Catholic liturgy.
- It will also avoid possible confusion, since Feria is a day of the week other than Sunday, and a weekday is (by definition) any day on a week.

This will address #88.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,9 +237,9 @@ Each date in the liturgical calendar is assigned a type. romcal defines these ty
 6. `MEMORIAL`
 7. `OPT_MEMORIAL`
 8. `COMMEMORATION`
-9. `WEEKDAY`
+9. `FERIA`
 
-Where the importance or rank of the celebration is in descending order (Solemnity being of highest importance and weekday being the lowest).
+Where the importance or rank of the celebration is in descending order (Solemnity being of highest importance and feria being the lowest).
 
 Types play an important role in determining which celebration should take precendence over another when two or more celebrations coincide on the same date. Certain celebration types will also have different liturgical colors applied to them.
 
--- a/src/constants/Types.js
+++ b/src/constants/Types.js
@@ -7,6 +7,6 @@ const Types = [
   "MEMORIAL",           // [5]
   "OPT_MEMORIAL",       // [6]
   "COMMEMORATION",      // [7]
-  "WEEKDAY"             // [8]
+  "FERIA"               // [8]
 ];
 export default Types;
--- a/src/lib/Calendar.js
+++ b/src/lib/Calendar.js
@@ -299,7 +299,7 @@ const _applyDates = ( options, dates ) => {
         }
 
         //------------------------------------------------------------------
-        // MEMORIAL or OPT_MEMORIAL that fall on a WEEKDAY
+        // MEMORIAL or OPT_MEMORIAL that fall on a FERIA
         // in the SEASON OF LENT are reduced to a COMMEMORATION
         //------------------------------------------------------------------
         else if (
@@ -312,20 +312,20 @@ const _applyDates = ( options, dates ) => {
         }
 
         //------------------------------------------------------------------
-        // MEMORIAL or OPT_MEMORIAL that fall on a WEEKDAY
-        // outside LENT or the Easter Octave will replace the general WEEKDAY
+        // MEMORIAL or OPT_MEMORIAL that fall on a FERIA
+        // outside LENT or the Easter Octave will replace the general FERIA
         //------------------------------------------------------------------
         else if (
-          _.eq( date.type, _.last( Types ) ) // If the current date is of type weekday
-          && !_.eq( date.data.season.key, LiturgicalSeasons.LENT ) // And this weekday is not in Lent
+          _.eq( date.type, _.last( Types ) ) // If the current date is of type feria
+          && !_.eq( date.data.season.key, LiturgicalSeasons.LENT ) // And this feria is not in Lent
           && ( _.eq( candidate.type, Types[5] ) || _.eq( candidate.type, Types[6] ) ) // And the candidate is either a memorial or optional memorial
         ) {
-          replace = true; // Then the candidate is fit to replace the weekday
+          replace = true; // Then the candidate is fit to replace the feria
         }
 
         //------------------------------------------------------------------
-        // A non prioritized FEAST can only replace WEEKDAY \
-        // - weekday must not be in the octave of EASTER
+        // A non prioritized FEAST can only replace FERIA \
+        // - feria must not be in the octave of EASTER
         //------------------------------------------------------------------
         else if (
           _.eq( candidate.type, Types[4] )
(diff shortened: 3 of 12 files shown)


emagnier:
After merged this one, I'm wondering if `FERIAL` might be better than `FERIA`.
These 2 terms are similar, but Ferial seems most used than Feria in English (and if I'm not wrong, feria is simply Latin for weekday/ferial)

emagnier:
On second thought, and after some readings, I think Ferial (english) is a more appropriate word than Feria (latin). And this will keep the codebase consistent with other english terms.
@tukusejssirs if you are ok I'll open a new PR soon to adjust the wording.

tukusejssirs:
#### Introduction

I have found some contradictions on the Internet. I don’t say that any of the following links is a primary source we need to adhere.

#### Pro _feria_

- [This site](https://en.wikipedia.org/wiki/Ferial) says that _ferial_ is ‘[p]ertaining to [f]eria days’ (i.e. it is adjective create from _feria_).
- [This site](https://en.wikipedia.org/wiki/Feria) uses _feria_ or _ferial day_.
- [This site](https://www.lexico.com/en/definition/ferial) says that _ferial_ means ‘[d]enoting an ordinary weekday, as opposed to one appointed for a festival or fast’ and even gives the following example how it is to be used: _These are usually referred to as festal responses, ferial responses keeping the chant in the treble and being harmonized in a simple fashion._ (again, _ferial_ is used as an adjective). Still, it says that it originates from mediaeval Latin _ferialis_ and that its late Latin form _feria_ came to mean ‘ordinary weekday’.

#### Pro _weekday_

[This website](https://en.wikipedia.org/wiki/General_Roman_Calendar) uses (though only once) the term _weekday_. I presume that this term is more widely used in the US and therefore in the US English, but I might be wrong.
Even the [LiturgyOffice.co.uk](http://www.liturgyoffice.org.uk/Calendar/2020/index.shtml) uses this term.

#### Conclusion

So my thoughts on this:

1. _Ferial_ word usage as you suggest is definately wrong, as this word is an adjective. The term _feria_ is a noun (substantivum), therefore from these two it is the only correct for our use case.
2. _Weekday_ is another correct term. I have no idea which one (_feria_ or _weekday_) is more widely used. I simply _prefer_ the term _feria_, as the more original a term/usage/whatever is, the more I love it (this is also the reason why I prefer UK English to US English).
3. Again I just _think_ that the reason lays in the **International Commission on English in the Liturgy (ICEL)** ([wiki](https://en.wikipedia.org/wiki/International_Commission_on_English_in_the_Liturgy), [official website](http://www.icelweb.org/)) and the change they made to the liturgical books, esp the _[Roman Missal](https://en.wikipedia.org/wiki/International_Commission_on_English_in_the_Liturgy#Roman_Missal)_. I believe they substituted the word _feria_ with _weekday_ in the latest translation of _Missale romanum, editio typica tertia_ (the 3rd typical edition). I don’t own a copy of the new translation (released in 2010 or 2011), so I cannot check upon it.
4. Therefore, If this would depend on me exclusively, I would keep the _feria_ wording. Not only because I prefer it, but also because it (1) is understood also by non-native English speakers and also (2) it won’t be ambiguous for English speakers that have not mastered the English in such depth (for them, _weekday_ might be a bit ambiguous).
5. @emagnier, could you send me some links where did you link the _ferial_ is _a more appropriate word_ to use?
6. Finally, I am definatelly sure that generally in a text (i.e. not in a code, variables, etc) the words _feria_, _ferial_ and _weekday_ should be written with lowercase initial letter, unless they are at the beginning of a sentence or in a title.

emagnier:
You are right, using "ferial" alone is incomplete, the complete term is "Ferial day", and the suggested type could be `FERIAL_DAY`.

I like "ferial day" because it have a bit of the words "feria" and "weekday" together.
I also prefer it over weekday, especially for non-native English speakers (like you mentioned).

I'm not a specialist, but I notice that "Ferial day" is generally most used and more in the common speech over "feria" (which a latin word and sounding more traditional).

I can not found all the articles I read speaking about ferial days, but here just a few:
https://episcopalchurch.org/library/glossary/feria-or-ferial-day
http://www.newadvent.org/cathen/06043a.htm
http://www.catholiclane.com/feria/

This is the reason I would prefer "ferial day" over "ferial" or "weekday".
What do you think?

emagnier:
An old english missal (before the usage of weekday) could help to do the choice. 
In the meantime it might be better to stay on "feria", since it mean the same thing, and we are just talking  about language preference in english.

tukusejssirs:
Well, in my opinion, the term _ferial day_ is far less used in English that the other two options. I personally would not even consider this option.

Here is what the Google has to say about the words usage on the Internet. I don’t say that this is the best way to count which one is used the most, but it gives some numbers at least. Note that all results are using Google Search set to search in English only. Also note that results without _liturgy_ keyword (might) include much wider sense of the terms.

|  |  |  |
|---|---|---|
|  |  |  |

> I notice that "Ferial day" is generally most used and more in the common speech over "feria"

I don’t know much about you personally (where you currently live or whom / what nationalities/natives you meet during your day), but to me _ferial day_ sounds a bit more troublesome since it is composed of two words. A bit off topic, but once a native British asked me, why they use the phrase _piece of cake_ instead of (for example) _piece of pizza_. His answer was that _cake_ is easier/faster to pronounce. I would apply same principle in this case (English[-speaking] people are lazy to pronounce anything.)

Still, I have no idea why you capitalise the _F_ in _Feria(l)_. There is no need when other rule(s) are not applied (like sentence case or title case). :)

> https://episcopalchurch.org/library/glossary/feria-or-ferial-day

I don’t know how much you know about [Christian denominations](https://en.wikipedia.org/wiki/Christian_denomination), but I don’t think that [Episcopal Church](https://episcopalchurch.org/who-we-are) should make the final call on which term we should use. On the other hand, I don’t say we should fight againt other Churches. All I want to say that different Churches (in the sense of denominations) could (and have) different (frequency of) usage of different words (although this might not be the case).

> http://www.newadvent.org/cathen/06043a.htm
> http://www.catholiclane.com/feria/

These are Catholic (I believe Roman Catholic) websites. For our quite _narrow_ field of use case, I would consider these sources as the ones to adhere, rather than the one from the Episcopal Church.

---

I think you are right that the _feria_ is more traditional term. However, I think that ICEL (and others) might prefer _workday_ (for some, to me unknown reason). I don’t say the _ferial day_ is not used at all, but it takes two words (and therefore it is longer) in contrary to _feria_.

All in all, I personally would vote for _feria_, but I think that ICEL (which somewhat unites all the English liturgical books for most of the English-speaking world) prefers/suggests _workday_.

tukusejssirs:
> An old [E]nglish missal (before the usage of weekday) could help to do the choice.

I might try to ask some of my friends, priests in the UK. I am not sure if and when they might answer.

> In the meantime it might be better to stay on "feria", since it means the same thing, and we are just talking about language preference in [E]nglish.

I agree on sticking to _feria_. :)

emagnier:
Thank you for your comments, and ok for `FERIA` 👍 

It was just a thought. I like to challenge ideas to move forward 😉

PR #1, opened 2015-03-16
Corrected typos in localization and add Polish strings

jarosz:
Hi!

I have corrected some typos in localization.json, especially those coming from transliterating Polish character 'ł' as 't' instead of 'l'.

I have also localized Polish strings according to Missale Romanum for Polish dioceses.

Issue #90, opened 2019-12-03
Shrove Monday and Tuesday


emagnier:
It seems these 2 dates are rather traditional and cultural, than a real liturgical celebration. 
I wonder if these elements really have their place in the liturgical calendar (for exemple in France we have also nothing about that in the liturgical calendar and they need to be removed).

tukusejssirs:
Well, I think so too. The current General Roman Calendar does not have these, but some particular calendars might include them as a _heritage_ from the historical calendars (before the Second Vatical Council).

If you want them to be removed, I suggest to keep the keys (for future use in historical calendars, #23, and the Extra-Ordinary Calendar #51). I agree on removing the definitions in `general.js` (and `drop` definitions in particular calendars).

emagnier:
Yes, this is a thing I'll do.

emagnier:
PR merged.

Issue #23, opened 2018-03-16
question / feature request: add historical liturgical calendar


tukusejssirs:
I have another suggestion that would require a bit of coding. It is even less important (at least to me) than issue #20, but if we are in the future about to implement it, all the coding from now on should take this in mind.

I propose to add a ’historical calendar.’ What I mean is we would:

- add options for
    - **forced Novo Ordo Missæ calendar**
    - **forced Traditional (Tridentine) Mass calendar**
    - **automatic** detection of calendar (Tridentine calendar between Tridentine and II. Vatican Councils; after II. Vatican Council Novo Ordo calendar)

Nice feature also would be to implement the day of addition of a celebration. What I mean is that a celebration (e.g. a saints day of celebration) is added to the calendar (either general or national) is added continually, not at the beginning. Just remember St John Paul II—before his canonisation there no celebration in the liturgy of this great saint. This would be included in all three ‘options’ listed above, of course, what was added only after II. Vatican Council, should not be added in the Tridentine calendar.

tukusejssirs:
I think that for beginning we should add two things.

Firstly, add possiblity to change calendar type. For that we need to create `const Calendars` in `src/constants/Calendars.mjs` (you might want to change anything in it):
```javascript
const Calendars = {
"tridentine": "Tridentine Calendar",  // https://en.wikipedia.org/wiki/Tridentine_Calendar
"cal1954": "General Roman Calendar of 1954",  // https://en.wikipedia.org/wiki/General_Roman_Calendar_of_1954
"cal1955": "General Roman Calendar of Pope Pius XII",  // https://en.wikipedia.org/wiki/General_Roman_Calendar_of_Pope_Pius_XII
"cal1960": "General Roman Calendar of 1960",  // https://en.wikipedia.org/wiki/General_Roman_Calendar_of_1960; this is currently autherised as extraordinary form of the Roman Rite (‘Tridentine Mass’ that might be currently celebrated)
"cal1969": "General Roman Calendar of 1969 (current)"  // https://en.wikipedia.org/wiki/Mysterii_Paschalis; this is the current GRC
};
export default Calendars;
```

Then, we need to have an option choose a calendar in the following method (`calendar: 'cal1969'`):
```javascript
romcal.calendarFor({
    year: 2018,
    calendar: 'cal1969',
    country 'unitedStates',
    locale: 'plPl',
    christmastideEnds: 't|o|e',
    epiphanyOnJan6: true|false,
[...]

PR #102, opened 2019-12-07
Drop shrove monday and tuesday from the general calendar

emagnier:
Shrove monday and tuesday are removed from the general calendar (since the current General Roman Calendar does not have these).

It is still possible to reintroduce them in particular calendars if needed.

Related to #90.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
 * Locale files have been renamed to kebab-case with capitalized regions to follow the IETF Language Codes standards and be close to the general conventions in software development for i18n.
 * In locale files, the `general` and `national` parts are now merged in a new `sanctoral` part. This will avoid potential duplicated or missing keys between these 2 parts, particularly when a celebration is move from the general to the national calendar. Also, keys in celebrations and the new sanctoral have been sorted alphabetically. ([#97](https://github.com/romcal/romcal/pull/97))
 * Italian translation has been added in romcal ([#94](https://github.com/romcal/romcal/issues/94)).
+* Drop Shrove Monday and Tuesday from the general calendar ([#90](https://github.com/romcal/romcal/issues/90)).
 * Potential breaking changes:
     - The `WEEKDAY` type is renamed to `FERIA` ([#88](https://github.com/romcal/romcal/issues/88))
 * Fixes:
--- a/README.md
+++ b/README.md
@@ -620,10 +620,6 @@ Usually, this means excluding a celebration defined in `src/calendars/general.js
 }
 ```
 
-An example of this can be seen in the national calendar of Slovakia (`src/calendars/slovakia.js`) where the celebrations of Shrove Monday and Shrove Tuesday are not relevant in the calendar output.
-
-Therefore, the `src/calendars/slovakia.js` redefines these 2 celebration keys with only one property, `drop` with the value of `true` so that they are excluded in the `calendarFor` output.
-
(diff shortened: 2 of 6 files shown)


tukusejssirs:
> Shrove monday and tuesday are removed from the general calendar (since the current General Roman Calendar does not have these).
> It is still possible to reintroduce them in particular calendars if needed.

I am okay with this change.

> Corresponding locales keys in French have been removed since these 2 celebrations are not defined in any liturgical calendar of french speaking countries.

I disagree with this. I know, you have already removed this from the current version of [your comment](https://github.com/romcal/romcal/pull/102#issue-350300926), but I want to emphasise that although the current French-speaking countries does not use this in their (current) calendars, this would make those strings (keys) untranslated (and thus displayed in English) whech a different calendar (one for different country) that celebrates those celebrations, in French locale. Just try to output Slovak (or Italian) calendar in French–as not all the keys are translated, most of the keys are displayed in English. And the problem will be even more complicated when we introduce more and more languages, and when even `histcal` (#23) will be implemented.

> Related to #90.

Actually, it is not only related to #90, but this PR (when merged) _fixes_ it. :smiley:

HISTORY.md
```diff
@@ -7,6 +7,7 @@
 * Locale files have been renamed to kebab-case with capitalized regions to follow the IETF Language Codes standards and be close to the general conventions in software development for i18n.
 * In locale files, the `general` and `national` parts are now merged in a new `sanctoral` part. This will avoid potential duplicated or missing keys between these 2 parts, particularly when a celebration is move from the general to the national calendar. Also, keys in celebrations and the new sanctoral have been sorted alphabetically. ([#97](https://github.com/romcal/romcal/pull/97))
 * Itallian translation has been added in romcal ([#94](https://github.com/romcal/romcal/issues/94)).
+* Drop shrove monday and tuesday from the general calendar ([#90](https://github.com/romcal/romcal/issues/90)).
```
tukusejssirs: Shrove Monday and Tuesday

emagnier:
> Corresponding locales keys in French have been removed since these 2 celebrations are not defined in any liturgical calendar of french speaking countries.

I've removed that after reading again your comment on #90. I know sometime I'm reading or do things very quicly, but I generally try to take time before do a merge or approve something, to have second thought on an idea or on what was made. I like to challenge ideas and intuitions (including mine), I'm like that. Here I agreed with what you already wrote and it's not necessary to give me more explainations 😄  

> Actually, it is not only related to #90, but this PR (when merged) fixes it.

Your are playing with words. I didn't used "fix" since it's not really a bug.

---

Thanks for the review 👍

tukusejssirs:
> Your are playing with words. I didn't used "fix" since it's not really a bug.

I don’t _play_ with the words (though it might seem so), but I use them in a _formal_ or _punctual_ way. In this case I also wanted to note (implicitly) that GitHub closes issues when some special keywords are used for example i  commit messages, like `fixes #1` ([more info](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords)).

And in my opinion, though that issue is not a bug, it is still an _issue_ (as in _GitHub issue_) and thus needs to be fixed/closed/resolved. One might prefer one keyword over other, but I seem to prefer _one word for all [cases]_. Therefore: I am sorry for eome of my notes and remarks. :smiley:

emagnier:
You're forgiven 😉 

Thanks for the GitHub keywords.

PR #112, opened 2019-12-11
Add the Feast of the Exaltation of the Holy Cross to general.js

### LINKED ISSUE #11: feast name: The Exaltation of the Holy Cross or The Triumph of the Cross
You used the latter in `locales/en.json`, while the [wikipedia](https://en.wikipedia.org/wiki/Feast_of_the_Cross) states that the former is the translation of the Roman Missal (does not states the edition—probably should be checked in _editio typica tertia_) and the latter is the translation from Roman Missal (1973). Consider the change to the Exaltation of the Holy Cross.

I think the English translation from Roman Missal editio typica tertia (2011) should be used for any and all of the feasts. On the other hand, we in Slovakia have editio typica altera only, therefore I’d correct my previous statement (sentence) to this: alway used the translation of feast names from the latest Roman Missal translation in a particular language.

tukusejssirs:
This issue fixes #11.


jarosz:
I'd rather merge romcal-es6 to master first, and ask you to create this request again against new master branch.

tukusejssirs:
I’d rather do it in reverse order. This won’t break the API, therefore it should be merged before `v1.3.0` release. Also I think that any and all merges to `master` should be a new release (at least a minor).

cc: @emagnier

src/locales/fr.js
```diff
@@ -63,7 +63,7 @@ export default {
     "peterAndPaulApostles": "Saint Pierre et Saint Paul, Apôtres",
     "presentationOfTheLord": "Présentation de Jésus au Temple",
     "sacredHeartOfJesus": "Sacré-Cœur de Jésus",
-    "theExaltationOfTheHolyCross": "The Exaltation of the Holy Cross",
+    "theExaltationOfTheHolyCross": "Exaltation de la Sainte-Croix",
```
emagnier: After verification in a missal and the [francophone bishop association](https://www.aelf.org/calendrier/romain/2019/09) website, I found that in French we don't literally translate it, but uses instead `La Croix Glorieuse`.

tukusejssirs:
@emagnier, I have fixed the localisation of `theExaltationOfTheHolyCross` in `fr.js`.

I have also added another commit (in case you’d like to omit it) with replacing all apostrophes with the typographic form of it (i.e. ’).

Issue #110, opened 2019-12-11
Votive Masses


tukusejssirs:
@emagnier, I have no idea if I should add the celebration of Sts Cyril and Methodius to 14 February in `slovakia.js`.

The official liturgical calendar (called _Direktórium_) issue the by Liturgical Commission at the Conference of Slovak Bishops each year, says the following:
- 14 Feb: votive Mass of Sts Cyril, Monk, and Methodius, Bishop, Patrons of Europe;
- 5 July: Solemnity of Sts Cyril and Methodius, Slavic Missionaries.

The problem lays in the _votive Mass_ part. I know that _votive Mass_ is not a liturgical rank, but on the other hand it is still included in the official liturgical calendar.

Should I include it in #46? Or should I skip it until _some release after `v2.0.0`_, when we would deal with this by adding a special rank `VOTIVE_MASS`?

jarosz:
https://en.wikipedia.org/wiki/General_Roman_Calendar#Slovakia - it states Feb 14 is a feast in Slovakia (as it's in Poland and I believe should be in all European countries).

tukusejssirs:
@jarosz, I know what it states, but it is wrong. If you can understand a bit of Slovak, you can check out https://lc.kbs.sk/?den=20190214 and https://lc.kbs.sk/?den=20190705, which fully corresponds with the official Slovak Directory (liturgical calendar).

Anyway, I wanted to fix it there, but there is to much politics in the community about how it _should_ be done, so I was discouraged.

emagnier:
Yes I saw that recently somewhere else for the Votive Mass of the Holy Spirit, and was wondering how this could be integrated.

Can we consider that `VOTIVE_MASS` could have the same rank/rules as an `OPT_MEMORY`?
I already played a bit with all these types in an experiment branch, but we need to rethink a bit first the way type dictionary (and all other constants) are managed in the code base (partially related to #109).

Anyway I'm in favor to include it soon. And depending on how we to the refactor, it will not break things or the API since it's just an addition. If we don't have time to include them in the v1.3.0, maybe this might be possible in v1.3.1.

jarosz:
In feria per annum you can celebrate any votive mass, but not any saint memorial (only optional memorial and those saints, who are in Martyrology for that particular day), so it's not exactly the same.

[...]

PR #114, opened 2019-12-13
Refactor rank types of celebrations

### LINKED ISSUE #109: Convert Types Array to Dictionary in types.js
Issue #38 was fixed, but if Types was a Dictionary it would improve code readability.

So Types[6] would change into Types.OPT_MEMORIAL.

emagnier:
* Improve code readability for rank types of celebrations.
* eslint: better es6 and Babel support.

resolve #109 

this will help to move forward on #110 

@jarosz, I worked on it last night, and when I opened a PR today I discovered your work [here](https://github.com/jarosz/romcal/commit/92ecc8ba14a99d7cc81ca71bb8fc6d8099bf790f). Our contributions was very close (extend an Array instead of using a Map object, to offer syntactic sugar). And I've include some of your code comments and the way you declare new type properties 👍 (with the `writable: false`). 
Thank you very much for your contribution. I've added you as a co-author of this commit 😉

### DIFF
--- a/src/calendars/chile.js
+++ b/src/calendars/chile.js
@@ -9,19 +9,19 @@ let dates = year => {
   let _dates = [
     {
       "key": "blessedLauraVicunaVirgin",
-      "type": Types[6],
+      "type": Types.OPT_MEMORIAL,
       "moment": moment.utc({ year: year, month: 0, day: 22 }),
       "data": {}
     },
     {
       "key": "blessedPiusIxPope",
-      "type": Types[6],
+      "type": Types.OPT_MEMORIAL,
       "moment": moment.utc({ year: year, month: 1, day: 7 }),
       "data": {}
     },
     {
       "key": "ourLadyOfLourdes",
-      "type": Types[5],
+      "type": Types.MEMORIAL,
       "moment": moment.utc({ year: year, month: 1, day: 11 }),
       "data": {
         "meta": {
@@ -31,7 +31,7 @@ let dates = year => {
     },
     {
       "key": "saintsPhilipAndJamesApostles",
-      "type": Types[4],
+      "type": Types.FEAST,
       "moment": moment.utc({ year: year, month: 4, day: 4 }),
       "data": {
         "meta": {
@@ -41,7 +41,7 @@ let dates = year => {
     },
     {
       "key": "saintTeresaOfLosAndesVirgin",
-      "type": Types[4],
+      "type": Types.FEAST,
       "moment": moment.utc({ year: year, month: 6, day: 13 }),
       "data": {
         "meta": {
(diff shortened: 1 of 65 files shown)


tukusejssirs:
I’ll review this in the evening.

However, I’d suggest that we should _always_ assign an issue to ourselves to eliminate duplicate work, and even probably note this in a comment.

emagnier:
Good point @tukusejssirs, we have to use the assignments feature to avoid these situations again.

src/calendars/chile.js
```diff
@@ -52,13 +52,13 @@ let dates = year => {
     // Commented because the key and rank is the same as the definition in general.js
```
tukusejssirs: I believe the problem lays in the addition of Optional Memorial of St Henry, Bishop of Martyr ([src](https://en.wikipedia.org/wiki/General_Roman_Calendar#Chile)), as on 13 July (when St Henry is celebration in `general.js`) in `chile.js` is Saint Teresa of Los Andes, Virgin (Feast) is celebrated.

I think you should correct this in this PR (and then remove the comment on line 52).

src/calendars/poland.js
```diff
@@ -666,26 +666,26 @@ let dates = year => {
...
       "data": {}
     },
     {
       "key": "saintBarbaraVirginAndMartyr",
-      "type": Types[6],
+      "type": Types.OPT_MEMORIAL,
       "moment": moment.utc({ year: year, month: 11, day: 4 }),
       "data": {}
     },
     // [pejulian] Needs confirmation: In Poland this day is a feast
```
tukusejssirs: @jarosz, could you confirm this please?

src/calendars/chile.js
emagnier: I hadn't even seen it. What should we do here, remove all this commented block?

tukusejssirs: You should do this:
1. Remote that comment block.
2. Create a new key (localise it in `en.js`): `"saintCamillusDeLellisPriestOrSaintHenryBishopAndMartyr": "Saint Camillus de Lellis, priest, or Saint Henry, Bishop and Martyr"` (`sk.js`: `Svätého Kamila de Lellis, kňaza, alebo svätého Henricha`).
3. Define `saintCamillusDeLellisPriestOrSaintHenryBishopAndMartyr` celebration in `chile.js`.

emagnier: Ok, this isn't very related to this PR, but I can do it as a separated commit.

src/calendars/poland.js
tukusejssirs: @emagnier, this one might also be a separate issue then—what do you think?

emagnier: yes same things

src/calendars/chile.js
tukusejssirs: Moved to issue #116.

src/calendars/poland.js
tukusejssirs: Moved to issue #116.

Issue #48, opened 2018-05-21
Calendar rankings (inheritance)


tukusejssirs:
In order to fix issue #13, we should create something like calendar ‘ranks’, i.e. calendars for more area types.

Currently, we only support General Roman Calendar and so-called ‘national’ calendars. The problem is that officially, there’s no such thing as national calendar (in the sense we use it); there are ~~calendars of dioceses~~ particular calendars. Nevertheless, I would leave the national calendars and besides dioceasan ones I would add continental a parish calendars, too. Here’s a summary:

- general
- continental (Africa, Asia, Australia and Oceania, Europe, North America, South America)
   - this would contain all continent specific celebrations (like patron saints of the particular continent) upon generation,
   - it would be built with General Roman Calendar as bass
- national
    - this would be a calendar for the particular country (like the UK, the US, Malaysia, Slovakia, etc)
    - it would be built upon continental calendar (and through it upon the General Roman)
- diocesan
    - liturgical calendar of diocese
    - built upon national (and throuth it upon continental and general)
- parish
    - this one usually does not exist, but still I think it (once) would be a nice feature;
    - it would contain celebrations that would be celebrated only in that particular parish, like the church patronage
    - later on, when/if we bring the ‘directory-style output’, we might include some other special days … for exampoe, in our parish, on the 18th of each month we pray for our parish
    - built upon diocesan calendar.

Note: There might be all/some dioceses in one country grouped together (like in Slovakia or the UK), or even countries (like Austria, Germany and Switzerland). In my opinion, it would be nice to have an option to generate these grouped calendars (either national, diocesan or parish) as grouped calendars, but also separately (although they would be same) in case the generator (person) won‘t know they are grouped.

emagnier:
And this might be related to #73. If we go ahead on this, the `national` directory will have no senses.

I've similar needs on my side, and I'll try to expose soon how it could be implemented.

emagnier:
What the purpose of grouping calendar together? I think in France we don't have this need, so I'm not sure what this addition could bring.
Can we consider it as a combination, like combine the french calendar and the slovak calendar to get a new combined calendar?
But we will get an issue with celebrations on both calendars that might have different properties (date, rank type, meta...).

tukusejssirs:
> What the purpose of grouping calendar together? I think in France we don't have this need, so I'm not sure what this addition could bring.

[...]

PR #124, opened 2019-12-15
Add celebrations for Chinese region (fixes #118)

### LINKED ISSUE #118: Add (group) calendar for China, Hong Kong and Taiwan
We could add this calendar based on [Wiki](https://en.wikipedia.org/wiki/General_Roman_Calendar#China,_Hong_Kong_and_Taiwan).

jarosz:
Based on Wikipedia article: https://en.wikipedia.org/wiki/General_Roman_Calendar#China,_Hong_Kong_and_Taiwan


tukusejssirs:
@emagnier, what do you think? Should we create a _group calendar_ (see #48) like this?

I don’t say that @jarosz’ work is bad—indead it’s awesome!—but we must look at the issues from higher perspective.

So, my question is: what should we do?
1. Should we wait with this PR/issue until we implement the _group calendars_?
2. Should we create three separate calendars for China, Hong Kong, and Taiwan?
3. Should we merge this PR as is (after reviews)?

jarosz:
I have no way to confirm validity of the Wiki article. For sure it needs clarification with Augustine Zhao Rong solemnity and memorial. Even with these inaccuracies, I think we can treat it as a reference, because it has a wider audience than romcal. You can correct Wiki with your findings.

Translation: I can do Polish myself, there's obviously no official translation.

emagnier:
If this calendar cover the China, Hong-Kong and Taiwan regions, we might rename `china.js` to `chinaHongKongTaiwan.js` to make it more explicit.

@tukusejssirs When #48 will be implemented, we will be able to leverage this calendar at that time.

Translations other than English are only a nice to have, so it's not necessary to add them in this PR (but @jarosz you still can add Polish translations here if you want to do them).

Except the file to rename, for me it's a LGTM.

jarosz:
I don't know if you're familiar with situation of the Church in China. Let's say it's not certain whether China, Taiwan and Hong-Kong should be treated as a whole or separate entities. I'd stick with china.js (maybe chinese.js, since the data is taken from Chinese Regional Bishops' Conference of Taiwan).

tukusejssirs:
Okay, let’s keep it the way it currently is.

@emagnier, I know that it is nice to have all keys localised, but I do my best to cover all the keys. Just for demonstration, the number of localised keys in each _main_ locale is as follows (note that this counts also keys like `season` or `day`):

```bash
$ wc -l [a-z][a-z].js
     5 cs.js
   724 en.js
   298 fr.js
   256 it.js
   306 pl.js
   723 sk.js  # This one will contain 724 (as `en.js`) when PR #127 will be merged)
```

So, I think that when you two, @emagnier and @jarosz, will have some time, you could search the Internet for the remaining keys. I did this last year for `sk.js`, it took some time, but I look forward to look at your ‘full’ locale translation. Again: I know that this is not the priority number one. :smiley:

emagnier:
> I don't know if you're familiar with situation of the Church in China.

@jarosz: No, I'm not too familiar with this situation. Ok, thanks for your suggestion.

> I know that it is nice to have all keys localised

@tukusejssirs: I was just talking about English locales (by suggesting what you already wrote elsewhere). Personally, I'm trying to find enough time to work on romcal presently, so for now I'll not be able to do more on localizing keys in French, unfortunately. If a calendar has already all its keys localized in the speaking language of its country, I prefer then to focus on features instead of taking time to translating everything.

PR #127, opened 2019-12-15
Additional improvements of the Slovak locale

tukusejssirs:
- replace double spaces with single spaces in en.js and sk.js (but kept the indentation);

- remove saintNicholas and saintPatrick keys from sk.js;

- improve the translation of guardianAngelOfPortugal;

- add translation of blessedIstvanSandorMartyr, findingOfTheHolyCross and ourLadyOfKnock.


tukusejssirs:
I have just found out that the number of keys in `en.js` and `sk.js` are not matching. After this PR is merged, this would be fixed and the number of the keys is equal and the keys are exactly the same. :smiley:

jarosz:
> I have just found out that the number of keys in `en.js` and `sk.js` are not matching. After this PR is merged, this would be fixed and the number of the keys is equal and the keys are exactly the same. 😃

Could you share the script you use to check it?

tukusejssirs:
@jarosz, I use Linux, so I use generally available Linux programs/utilities. I you need anything else, just ask. :smiley:

```bash
# Count lines in the main locales (while in `src/locales/`, obviously)
wc -l [a-z][a-z].js

# List all keys in a locale (e.g. `en.js`) unordered, non-unique
# Note that this outputs also keys like `season` or `day`
grep -Po '^\s*"\K[^"]*(?=": ")' en.js

# List all keys in a locale without selected keys (`after`, `before`, `day`, `day_after_ash_wed`, `feria`, `octave`, `season`, `sunday`) unordered, non-unique
grep -Po '^\s*"\K[^"]*(?=": ")' en.js | sed '/^after$\|^before$\|^day$\|^day_after_ash_wed$\|^feria$\|^octave$\|^season$\|^sunday$/d'

# List all keys in a locale without selected keys, ordered and without duplicates
grep -Po '^\s*"\K[^"]*(?=": ")' en.js | sed '/^after$\|^before$\|^day$\|^day_after_ash_wed$\|^feria$\|^octave$\|^season$\|^sunday$/d' | sort | uniq

# Get the difference between one locale and another locale
diff -y --suppress-common-lines <(grep -Po '^\s*"\K[^"]*(?=": ")' en.js | sed '/^after$\|^before$\|^day$\|^day_after_ash_wed$\|^feria$\|^octave$\|^season$\|^sunday$/d' | sort | uniq) <(grep -Po '^\s*"\K[^"]*(?=": ")' pl.js | sed '/^after$\|^before$\|^day$\|^day_after_ash_wed$\|^feria$\|^octave$\|^season$\|^sunday$/d' | sort | uniq)
```

jarosz:
git on Windows is bundled with a working bash shell and these tools :) Thank you.

tukusejssirs:
If you use MS Windows 10, I’d suggest to use Windows Subsystem Linux 2 (WSL2). :smiley:

PR #141, opened 2019-12-26
Add St John Nepomune to Mexico, Slovenia, Germany, Austria, Switzerland (fixes #30)

### LINKED ISSUE #30: Add `saintJohnNepomucenePriestAndMartyr` to `mexico.mjs`, `slovenia.mjs`, (`austria.mjs`, `switzerland.mjs`)
This Optional Memorial on 16th May [Mexico](https://en.wikipedia.org/wiki/General_Roman_Calendar#Mexico), [Slovenia](https://en.wikipedia.org/wiki/General_Roman_Calendar#Slovenia), [Germany+Austria+Switzerland](https://en.wikipedia.org/wiki/General_Roman_Calendar#Austria,_Germany,_Switzerland)—for these three countries see the note in the link [we should probably create something like I suggested for locales in issue #24 (3), i.e. create a main national calendar for German speaking countries and sub-calendars for the particular countries]}.

tukusejssirs:


tukusejssirs:
I have added that celebration to Germany, Austria, Switzerland separately, although we should create a super-national calendar for these three calendars, therefore when #48 will be implemented, these three has to be revised.

PR #147, opened 2020-01-13
Fix installation script when romcal is installed as a node dependency

emagnier:
Use `prepublish` (instead of `install`): Run BEFORE the package is packed and published, as well as on local `npm install` without any arguments.

- For romcal contributors or when publishing a new version of romcal, it will change nothing. 
- But this fix an issue when romcal is installed as a node dependency, from another node module (the `install` script from romcal was executed when it was not necessary, and it caused an error)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "romcal",
-  "version": "1.3.0-rc1",
+  "version": "1.3.0-rc2",
   "description": "Utility library that outputs the Liturgical Calendar used by the Roman Rite (Western Church)",
   "main": "dist/index.js",
   "module": "src/index.js",
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "prepack": "npm run build && npm test",
-    "install": "npm run build",
+    "prepublish": "npm run build",
     "build": "npm run build:babel && npm run build:webpack",
     "build:babel": "babel src --out-dir dist",
     "build:webpack": "webpack-cli --config webpack.config.js",
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,9 @@
 # Version history
 
-## 1.3.0-rc.1 /2020-12-01
+## 1.3.0-rc.2 /2020-01-13
+* Fix installation script when romcal is installed as a node dependency
+
+## 1.3.0-rc.1 /2020-01-12
 * Drop Shrove Monday and Tuesday from the general calendar (#90).
 * Refactor rank types of celebrations (#109).
 * Various bug fixes and improvements (#105 #133 #129 #44).


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "prepack": "npm run build && npm test",
-    "install": "npm run build",
+    "prepublish": "npm run build",
```
tukusejssirs: @emagnier, do I understand this right that `npm install` would after merging this PR fail as there would be so such script? Does `npm` not require the `install` script to be defined?

And anyway, how does this fix this issue? You have just changed the _name_ of the script (from `install` to `prepublish`), not the _script itself_ (which remained `npm run build`), therefore the outcome should be the same. What do you think?

Anyway, I think should still have `install` script and create a new one (e.g. `install_prepublish`) that would fix the issue you have described. Or even better: the script should have check that when it fails, it catches the error and runs the other script (e.g. `install_prepublish`).

emagnier: In this script section, the names here are important, because they correspond to specific npm handlers. You can't name them as you want.

FYI: https://docs.npmjs.com/misc/scripts

emagnier: And to manually run the `install` script from this section, you had to do `npm run install` which is not the same script as `npm install`.

tukusejssirs: Good to know! I resolve this conversation, but read on please. :smiley:

```

Anyway, I suggest that the PR description should be either a list of changes (like I did in [#146](https://github.com/romcal/romcal/pull/146#issue-362122454)) or (when a change/issue is more complex) a short description followed by list of what needs to be paid heed to (like in this PR, how romcal should be build) and other remarks. The PR title is the merge subject (first line of the merge message) and the PR description should be the merge description (the rest of the merge message). PR description (merge description) is not required when the particular PR change is simple and is described in the PR title / merge subject. But when there is the description, there needs to be a blank line between merge subject and merge description in order to be properly parsed by `git` and GitHub (this is automatically done by GitHub when you merge PRs on GitHub website).

Therefore, I suggest to change the description of this PR as follows:


```markdown
When romcal is installed as a node dependency, from another node module (the `install` script from romcal was executed when it was not necessary and it caused an error).

- Use `prepublish` (instead of `install`), which is run before the package is packed and published, as well as on local `npm install` without any arguments.
- For romcal contributors or when publishing a new version of romcal, this will change nothing. 
```

emagnier: Ok ok, you are very perfectionist. On the romcal project, I spending more time on things like that, and to explain things, instead of improving romcal itself... This is very time consuming for me (particularily to do it in english) and during this time I'm not working on my other projects that allow me to earn my living.

tukusejssirs: You’re right. Sorry for wasting your time. I’ll do my to best to value your time much more :disappointed:

tukusejssirs:
When running `npm install`, it warns that `prepublish` scripts are deprecated.

```bash
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
```

tukusejssirs:
I think you should amend this merge. :smiley:

emagnier:
I didn't seen that. The NPM documentation could have indicated that it's deprecated. I will update it. Thanks.

tukusejssirs:
Anyway, the `package-lock.json` lists romcal version `v1.3.0-rc1`, but when it is build, it is changed to `v1.3.0-rc2` and thus marks tha `package-lock.json` file as modified. Could it be changed everytime a new romcal version out?

PR #152, opened 2020-01-25
Pre-release tidy-up

tukusejssirs:
- Bump romcal version in `package.json`;
- Fix Git URLs in `package.json`;
- Fix the link to `README.md`;
- Fix the casing of _Localizing celebration names_ in the TOC of the `README.md`.
- Add specific link names to sections in `HISTORY.md`;
- Add blank lines between headings and the following paragraph make the `HISTORY.md` be more in conformity with Markdown standard syntax;
- Fix indentation in `HISTORY.md` (change 4 spaces to 2 spaces);
- Remove `cs.js`, en-BO.js and en-PH.js as there is no localised keys in there;
- Add list of contributors to `package.json`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 
 ## 1.3.0 /2020-01-25 <a name="1.3.0"></a>
 
-This release changed a lot since [1.2.4](#1.2.4), therefore we list here only the main changes and small changes were combined. For more thorough list of changes, check the release notes for [1.3.0-beta.1](#1.3.0-beta.1), [1.3.0-beta.2](#1.3.0-beta.2), [1.3.0-beta.3](#1.3.0-beta.3), [1.3.0-rc.1](#1.3.0-rc.1) and [1.3.0-rc.2](#1.3.0-rc.2).
+romcal changed a lot since [1.2.4](#1.2.4), therefore we list here only the main changes and small changes were combined. For more thorough list of changes, check the release notes for [1.3.0-beta.1](#1.3.0-beta.1), [1.3.0-beta.2](#1.3.0-beta.2), [1.3.0-beta.3](#1.3.0-beta.3), [1.3.0-rc.1](#1.3.0-rc.1) and [1.3.0-rc.2](#1.3.0-rc.2).
 
 ### General changes
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -19,14 +19,14 @@ romcal changed a lot since [1.2.4](#1.2.4), therefore we list here only the main
 
 ### Calendar and celebrations
 
-- Added national calendars for Czech Republic and Slovakia.
+- Added national calendars for Czech Republic, Italy and Slovakia.
 - Drop Shrove Monday and Tuesday from the General Roman Calendar ([issue #90](https://github.com/pejulian/romcal/issues/90)).
 - Epiphany is always celebrated on Jan 6 in Slovakia ([issue #29](https://github.com/pejulian/romcal/issues/29)).
 - Apply changes in the General Roman Calendar made by the Holy See.
 
 ### Localization
 
-- Add Slovak locale files.
+- Add Italian and Slovak locale files.
 - Some localization keys where renamed in order to confirm with official English names of the celebrations.
 
 ### Build process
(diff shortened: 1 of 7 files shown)


HISTORY.md
```diff
@@ -1,107 +1,159 @@
...
+
+- Add Slovak locale files.
+- Some localization keys where renamed in order to confirm with official English names of the celebrations.
+
+### Build process
+
+- Update and simplify the build process.
+- Add automated build process to transpile the code after installing and before publishing.
+- Distribution files are now not included anymore in the codebase.
+- In addition to the `romcal.bundle.min.js` file (for browsers), source code is also transpiled by Babel in the `dist/` directory. This makes it easier to include romcal in any node.js package.
```
emagnier: ...is also transpiled by Babel and Webpack...

~~This makes it easier to include romcal in any node.js package.~~ That new feature is to allow using romcal directly in the browser (which is new), so not only in a node.js package or environment (babel and webpack allow to have the same codebase for backend/node.js and frontend/browser).

HISTORY.md
```diff
@@ -1,107 +1,159 @@
...
+### Calendar and celebrations
+
+- Added national calendars for Czech Republic and Slovakia.
+- Drop Shrove Monday and Tuesday from the General Roman Calendar ([issue #90](https://github.com/pejulian/romcal/issues/90)).
+- Epiphany is always celebrated on Jan 6 in Slovakia ([issue #29](https://github.com/pejulian/romcal/issues/29)).
+- Apply changes in the General Roman Calendar made by the Holy See.
+
+### Localization
+
+- Add Slovak locale files.
```
emagnier: And Italian.

HISTORY.md
```diff
@@ -1,107 +1,159 @@
...
+- Refactor rank types of celebrations ([issue #109](https://github.com/pejulian/romcal/issues/109)).
+
+### New features
+
+- Introduce the new `drop` property that allows national calendars to omit unwanted celebrations from their output ([issue #14](https://github.com/pejulian/romcal/issues/14)).
+- Introduce new `christmastideIncludesTheSeasonOfEpiphany` flag to allow output to include or exclude the "season" of Epiphany from being merged into Christmastide.
+
+### Calendar and celebrations
+
+- Added national calendars for Czech Republic and Slovakia.
```
emagnier: And Italian.

HISTORY.md
tukusejssirs: What do you think about the following?

> - In addition to the `romcal.bundle.min.js` file (for browsers), source code is also transpiled by Babel and Webpack in the `dist/` directory. This feature makes it possible to use romcal directly in the browser, so not only in a node.js package or environment (babel and webpack allow to have the same codebase for backend/node.js and frontend/browser).

emagnier:
Nice work. You also need to update the `locales/index.js` (remove ref to `cs.js`) to make Travis happy.

I'm still wondering what `en-BO` and `en-PH` are doing here but this might be not the goal of this PR.

tukusejssirs:
> I'm still wondering what `en-BO` and `en-PH` are doing here but this might be not the goal of this PR.

@pejulian created those to localise that key differently for those regions/countries. In order to comply with the decisions we two made, I think we should remove them. Still, when we separate the person celebrations (those of saints/blesseds), we will deal with this differently.

Therefore I delete those two files, too. (And also from the `index.js`. :smiley:)

tukusejssirs:
And I think we should delete also `en-CA` file.

emagnier:
`en-CA` might stay, because the translation seems different for this francophone saint for that country (however I'm not sure how theses translations are accurate).

tukusejssirs:
> `en-CA` might stay, because the translation seems different for this francophone saint for that country (however I'm not sure how theses translations are accurate).

Well, in [`en-CA.js`](https://github.com/romcal/romcal/blob/master/src/locales/en-CA.js), there is only an addition of `Secondary Patrons of Canada` title, otherwise the key value is same as in `en.js`. I think this is more of a _calendar-specific_ thing (I have suggested that saint/blessed titles should be defined in the calendars).

But if you think we should really keep keep, just say so. :smiley:

emagnier:
Also, keeping the `en-CA` will give an exemple for potential new contributors.
There is also a different translation of Saint Jean / Saint John.

The actual usage of `en-PH` and `en-BO` was wrong because it's a kind of hack (to add another saint on a same/existing celebration).

tukusejssirs:
Okay, I’ll re-add it. But I still think that in the future, it will be removed altogether. As for the example for potential new contributors: I’d rather have a page in a Wiki for that. :smiley:

tukusejssirs:
Anything else you’d like me to change? Should I squash the changes or will you do that at merge time?

emagnier:
Great! let's merge it and release the `1.3.0` 🎆 🎉 🎉 
A big thank you for all your hard work!

tukusejssirs:
Will you release at `npm` too? :smiley:

emagnier:
yes ! 🚀

Issue #79, opened 2019-11-21
Create the community standard files


tukusejssirs:
I suggest we should create files that is considered as _community standard_.
- [ ] `contributing.md` (#69, #77);
- [ ] `code_of_conduct.md` (#59);
- [ ] issue template; read https://help.github.com/articles/about-issue-and-pull-request-templates/;
- [ ] pull request template; read https://help.github.com/articles/about-issue-and-pull-request-templates/;
- [ ] `readme.md`, `readme.[lang].md` (#61);
- [ ] `authors.md`; e.g. https://github.com/sinatra/sinatra/blob/master/AUTHORS.md;
- [ ] `contributors`;
- [ ] `history` → `changelog`;
- [ ] (`maintainance`);
- [ ] `version` (#60);
- [ ] `support`; e.g. https://github.com/atom/atom/blob/master/SUPPORT.md;
- [ ] `style_guide`; modify [Angular](https://angular.io/guide/styleguide) to our needs; probably should inporporated in `contributing.md`;
- [ ] `vision.md`; create a list of features to be added in the future (or vision / roadmap); let there be features divided into sub-features (which compose the features);

pejulian:
Hello @tukusejssirs and @emagnier,

In respect to infrastructure enhancements, I've done some work on the deployment process of `romcal` to automate the npm publishing process.

The first step was to create a machine user that can manage automated release activities:

1. Created a new "system" user with the email address (romcalproject@gmail.com)
2. Created a new "machine user" account in GitHub
3. Created a new TravisCI (.org) account with the machine user above.
    - We should only need this travis instance moving forward for running tests, publishing and so on
4. Created a GitHub personal access token to allow TravisCI to push updates to `romcal/romcal`
5. Created an `npm` token to allow TravisCI to publish new packages to `npm`
6. Encrypted the tokens above to be used by travis during deployments
    - These tokens will only be supplied to the build if the travis instance above is used

The second step was to enhance the existing `travis.yml` file:

- Updated `node_js` to support the latest stable release and all LTS releases only
- Run unit tests in the above versions of node 
- Deploy a new version of `romcal` using `semver` with the following constraints
   - The push event is on a branch (not a pull request)
   - The push event is not on a fork of `romcal/romcal`
   - The publish only happens once per build on the latest stable version of node
   - The updated version is pushed back to the branch that the build originated from 

To top it off, I also:

[...]

PR #164, opened 2020-02-01
Streamline and automate the versioning and publishing of romcal

pejulian:
As described in [my comments](https://github.com/romcal/romcal/issues/79#issuecomment-581043070) on #79, this PR aims to make life easier for contributors of `romcal` by handling:

- semantic versioning
- distribution tagging
- publishing to npm

Once enacted, we should be able to rely on the automated infrastructure to handle all the above for us.

The machine user saves contributors from:
- needed to have access to Travis and NPM for build and publishing
- having to manually keep track of semantic versioning numbers and formats
- having to set up their machines to publish new versions of the module
- suffer setbacks from a non responding human party with all the needed administrative rights to publish the module (i.e. me)


emagnier:
Great! I'll try to do a review tonight or tomorrow on it.
Also you mention some remaining todos in this [comment](https://github.com/romcal/romcal/issues/79#issuecomment-581043070), do you plan to add them here or we can already merge this PR before theses future additions?

scripts/bump-version.sh
```diff
@@ -0,0 +1,70 @@
...
+
+    echo "LATEST_TAG_VERSION is $LATEST_TAG_VERSION"
+
+    tokens=($(echo $LATEST_TAG_VERSION | tr "-" "\n"))
+    SEMVER=${tokens[0]};
+    PREID=${tokens[1]};
+
+    # If tag version is higher than the 'latest' version in npm
+    # then it means that no new version has been published to master yet and
+    # we shuld continue incrementing the preid based on the tag version
```
tukusejssirs:
```bash
    # If the tag version is higher than the "latest" version in npm,
    # then no new stable version has been published from master yet
    # and we should continue incrementing the "preid" based on the tag version 
```

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+###
+# semantic version comparition using semver specification http://semver.org/
```
tukusejssirs: I believe that `comparition` is misspelled `comparison`. Still, I think it might sound better when re-worded.

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
...
+# This bash script compares pre-releases alphabetically as well
+#
+# returns 1 when A greater than B
+# returns 0 when A equals B
+# returns -1 when A lower than B
+#
+# Usage
+# chmod +x semver.sh
```
tukusejssirs: Why is `chmod`-ing required? Git repos keep the permissions, therefore _if you commit_ this file with `+x` (or even `a+x`), it will be executable when cloned and checked out on any machine.

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
...
+# ./semver.sh 1.0.0 v1.0.0-rc.0
+# --> 1
+#
+# Author Ariel Rodriguez
+# License MIT
+###
+semver_compare() {
+  local version_a version_b pr_a pr_b
+  # strip word "v" and extract first subset version (x.y.z from x.y.z-foo.n)
```
tukusejssirs: As I see it, you have copied this script from somewhere. Still, I’d capitalise the first letters of all the comments.

And add the definite article (`the`) before ordinal number `first`.

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
...
+    pr_b=$(echo "$2" | awk -F'-' '{print $2}')
+
+    ####
+    # Return 0 when A is equal to B
+    [ "$pr_a" \= "$pr_b" ] && echo 0 && return 0
+
+    ####
+    # Return 1
+
+    # Case when A is not pre-release
```
tukusejssirs: I’d remove `Case` from the beginning. `When A is not a pre-release` is suffiecient (also note the indefinite article `a` before `pre-release`).

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
...
+    if [ -z "$pr_a" ]
+    then
+      echo 1 && return 0
+    fi
+
+    ####
+    # Case when pre-release A exists and is greater than B's pre-release
```
tukusejssirs: I’d go with `When pre-release A exists and is greater then pre-release B`.

scripts/semver.sh
```diff
@@ -0,0 +1,71 @@
...
+
+    # extract numbers -rc.x --> x
```
tukusejssirs:
```
Extract `rc` numbers
````

File: scripts/semver.sh [hunk 8]
```diff
@@ -0,0 +1,71 @@
...
+    number_a=$(echo ${pr_a//[!0-9]/})
+    number_b=$(echo ${pr_b//[!0-9]/})
+    [ -z "${number_a}" ] && number_a=0
+    [ -z "${number_b}" ] && number_b=0
+
+    [ "$pr_a" \> "$pr_b" ] && [ -n "$pr_b" ] && [ "$number_a" -gt "$number_b" ] && echo 1 && return 0
+
+    ####
+    # Retrun -1 when A is lower than B
```
tukusejssirs: `Retrun` → `Return`

tukusejssirs (changes requested):

pejulian:
> Great! I'll try to do a review tonight or tomorrow on it.
> Also you mention some remaining todos in this [comment](https://github.com/romcal/romcal/issues/79#issuecomment-581043070), do you plan to add them here or we can already merge this PR before theses future additions?

Hey @emagnier,

I just added an enhancement to the script where we can use the identifier `[USE_CUSTOM_SEMVER]` in a commit message to skip the auto incrementing mechanism to prefer whatever value the user supplied in the `version` field of `package.json`. As with all other human operations, human error is non avoidable and the build will fail if an incompatible version was supplied (i.e. lower version that what is already published). As such, it will be the user's responsibility to check the latest version published on the given tag(s) and supply a higher (and valid semantic) version.

I think what we have now is enough to put into general use.
The other "todo's" for git tagging and auto change log can be merged in at some later point when I have them ready.

scripts/bump-version.sh
pejulian: Hi @tukusejssirs, I've amended the comments here as per your recommendation.

scripts/semver.sh
pejulian: Hi @tukusejssirs,

For this and the comments below in respect to `semver.sh`, I think I'll leave the code and comments as-is because I did not write it. I could have but it's some pretty tedious stuff! So I found a code written by someone else that worked and decided to use that instead.

Further to this, I've also just updated the content of `semver.sh` with a more complete library that I found: 
https://github.com/fsaintjacques/semver-tool

scripts/semver.sh
pejulian: I'm nor sure why but I ran into some permission issues on TravisCI without explicitly adjusting file permissions.

Then I found this:
https://stackoverflow.com/questions/39666585/does-git-store-the-read-write-execute-permissions-for-files

It could be a safety precaution to adjust permissions when needed to avoid relying on Git entirely.

pejulian:
You may have noticed that CI checks for this PR have been failing. That is so because I had previously set up a travis-ci.com (pro) account that triggered builds for this repository. The builds fail on the deploy step because the new `travis.yml` and corresponding bash scripts require environment variables that are now only available via the machine user travis-ci.org instance. This mechanism ensures that only 1 user is ever only responsible for build and publishing related activities for `romcal`. I have deactivated the old travis-ci.com instance on `romcal`. Future CI reports should now only come from the machine user travis instance.

emagnier (approved):
For me it's a LGTM. As soon as @tukusejssirs approve this PR, we can merge it.
I'm waiting to merge other pending PRs, I'll merge them when that one is merged. So we can test this new automated process.

tukusejssirs:
> [@pejulian] As with all other human operations, human error is non avoidable and the build will fail if an incompatible version was supplied (i.e. lower version that what is already published). As such, it will be the user's responsibility to check the latest version published on the given tag(s) and supply a higher (and valid semantic) version.

I think that the custom version can (and should) be checked against the latest released version of the particular ‘release type’ (i.e. stable, beta, canary). Even more: I think that the user could just specify that they want to release the next major release and the `semver` should auto-calculate the next major version number (curretly that would be `2.0.0`). This could apply to minor and patch numbers, as well as beta/canary numbers.

tukusejssirs (approved):

pejulian:
> > [@pejulian] As with all other human operations, human error is non avoidable and the build will fail if an incompatible version was supplied (i.e. lower version that what is already published). As such, it will be the user's responsibility to check the latest version published on the given tag(s) and supply a higher (and valid semantic) version.
> 
> I think that the custom version can (and should) be checked against the latest released version of the particular ‘release type’ (i.e. stable, beta, canary). Even more: I think that the user could just specify that they want to release the next major release and the `semver` should auto-calculate the next major version number (curretly that would be `2.0.0`). This could apply to minor and patch numbers, as well as beta/canary numbers.
> 
> **Update:** However, my suggestion above should not be taken as merge blocker, i.e. the script might be improved at a later stage (in that case, this should be moved into the `backlog`).

These are great suggestions @tukusejssirs 

It should be possible to expand the regex that matches the `[USE_CUSTOM_SEMVER]` substring to include a custom `semver` value itself (i,e. `[USE_CUSTOM_SEMVER:major|minor|patch|...]`) which will then be checked to ensure that it is a valid `semver` value before being applied to the package.

I'll look into this when I get some time.

PR #165, opened 2020-02-01
Fix misspelled word POrtugal in `portugal.js`, `en.js` and `sk.js`

tukusejssirs:


src/calendars/portugal.js
```diff
@@ -127,7 +127,7 @@ let dates = year => {
       }
     },
     {
-      "key": "blessedSanchaAndMafaldaVirginsOrBlessedTheresaOfPOrtugalReligious",
+      "key": "blessedSanchaAndMafaldaVirginsOrBlessedTheresaOfPrtugalReligious",
```
pejulian: Could there be another typo here?

`blessedSanchaAndMafaldaVirginsOrBlessedTheresaOfPrtugalReligious`
might have intended to be
`blessedSanchaAndMafaldaVirginsOrBlessedTheresaOfPortugalReligious`
instead

tukusejssirs: You’re right. I made a mistake there. :smiley:

PR #166, opened 2020-02-03
Fix wrong translations in French for Epiphany and Palm Sunday

emagnier:
In french:
- `epiphany` had the localized title of `palmSunday`.
- `palmSunday` had an incomplete title.


src/locales/fr.js
```diff
@@ -49,7 +49,7 @@ export default {
     "corpusChristi": "Saint-Sacrement",
     "divineMercySunday": "Dimanche de la Divine Miséricorde",
     "easter": "Dimanche de Pâques",
-    "epiphany": "Dimanche des Rameaux et de la Passion",
+    "epiphany": "L'Épiphanie du Seigneur",
```
tukusejssirs: @emagnier, as I have already replaced all `'` with `’`, could you replace it in here too please? :smiley:

`L’Épiphanie du Seigneur`

emagnier: Done.

Issue #174, opened 2020-03-01
`Saint` vs `Saints` in celebration names that include `companions`. Simplifying the l10n key nanes


tukusejssirs:
The inconsistency lays in the word `saint(s)` or `blessed(s)` (the _canonisation level_).

For `blessed(s)`, we use it consistently: in singular. For `saint(s)`, we use `saint` 18 times and `saints` 11 times.

Now, we need to decide how will we name the celebrations that include `companions`. Which option is (grammatically) correct?

Take for example these two options (currectly, these are duplicates we need to fix; these keys made me notice this issue):
1. singluar: _Saint Augustine Zhao Rong, Priest and Companions, Martyrs_;
2. plural: _Saints Augustine Zhao Rong, Priest, and Companions, Martyrs_.

In the official Sloval liturgical calendar (_Direktorium_), we use the plural version, which might be _grammatically_ more correct, as we (may) want to say that not only _Augustine Zhao Rong_ is saint, but also _his companions_. But this might not be actually _true_, because not all the companions _must_ be saints (some of them might be blesseds).

As part of the _saints metadata separation process_ (#154), I want to list all the companions (and other _collective names of the celebrated persons_, like the _Vietnamese martyrs_) and this _trueness dilemma_ might be solved by checking if all the _non-named persons_ are of the same canonisation level as the _named ones_: if true, use plural option; if false, use singular option unless we name two or more persons who have the same canonisation level.

Any ideas, @pejulian, @emagnier?

emagnier:
FYI In the official francophone proper calendars, they use the singular version.
And I think it's better because we separate better each entities/parts:
1. Saint Augustine Zhao Rong, Priest
2. Companions
3. Martyrs (<= the common titles for entities 1 and 2)

Programatically it will be also easier to combine them that way.

tukusejssirs:
Okay, I have asked a friend of mine (a Slovak teacher) what is the grammatically correct option in Slovak language. As for me, _formally_ I think that the _singular option_ is better because we should not take the _companions_ into account (unlike _named_ multiple sainf of the same canonisation level, like _Saints Peter and Paul, apostles_). Then I’ll come back and provide the correct Slovak version.

As for the separation, I agree. I have already suggested this from the l10n point of view in #154.

tukusejssirs:
After consulting with two Slovak language teachers (one of them proofreads books and other kind of written articles), and I have found two things (from grammatical point of view):
[...]

PR #178, opened 2020-03-03
Fix some localisation issues

tukusejssirs:
- Add St Paul IV, Pope celebration to `general.js`;
- Add St Paul IV, Pope key to `en.js` and `sk.js`;
- Remove `saintsCosmasAndDamian` in favour of `saintsCosmasAndDamianMartyrs`;
- Remove `saintHedwigReligiousSaintMargaretMaryAlacoqueVirgin` in favour of `saintHedwigReligiousOrSaintMargaretMaryAlacoqueVirgin`;
- Remove `saintJohnOfGodPriest` in favour of `saintJohnOfGodReligious`;
- Remove `saintJuanDiego` in favour of `saintJuanDiegoCuauhtlatoatzin`;
- Remove `saintMarkApostle` in favour of `saintMarkTheEvangelist`;
- Remove `saintOliverPlunketBishopAndMartyr` in favour `saintOliverPlunkettBishopAndMartyr`;
- Remove `saintPioOfPietralcina` and `saintPioOfPietrelcinaPadrePioPriest` in favour of `saintPioOfPietrelcinaPriest` (with the `en.js` value changed to `Saint Pio of Pietrelcina`);
- Remove `saintEzekielMorenoBishop` in favour of `saintEzequielMorenoBishop`;
- Remove `saintLorenzoRuizAndCompanionsMartyrs` and `saintsLawrenceRuizAndCompanionsMartyrs` in favour of `saintLawrenceRuizAndCompanionsMartyrs`;
- Rename `saintWenceslausMartyrSaintsLawrenceRuizAndCompanionsMartyrs` to `saintWenceslausMartyrSaintLawrenceRuizAndCompanionsMartyrs`;
- Rename `saintClementOfOhridAndGorazdBishopsAndCompanions` to `saintsClementOfOhridAndGorazdBishopsAndCompanions`;
- Rename `saintAgnesCaoGuiyingMartyr` to `saintAgnesTsaoKouYingMartyr` (and change the `en.js` value);
- Fix apostrophes in l10n files;
- Add missing keys from `en.js` to `sk.js`;
- Fix the value of `saintBlaseBishopAndMartyrSaintAnsgarBishop` (change _and_ to _or_) in `en.js` and `sk.js`.


tukusejssirs:
#### Notes

On St John of God: according to [Wiki](https://en.wikipedia.org/wiki/John_of_God), he was not ordained a priest nor does he has _priest_ title in [Slovak liturgical calendar](https://lc.kbs.sk/?den=20210308), nor at [Universalis](https://universalis.com/200/calendar.htm?year=2021).

On St Mark: I could not find any resource (except for [Wiki](https://en.wikipedia.org/wiki/General_Roman_Calendar#New_Zealand)) with liturgical calendar of New Zealand that would use _apostle_ title for St Mark. Source I have found: [Universalis](http://universalis.com/NZ/calendar.htm?year=2021) and [GCatholic](http://www.gcatholic.org/calendar/2018/NZ-en.htm).

On St Oliver Plunkett: I am not sure which surname (either with a single `t` or with doubl `tt`) is correct. On the Internet, there one can find both surname spellings. I’ve chosen one

On St Padre Pio: I have change the the in accordance with [the calendar of the Liturgy Office (England & Wales)](http://www.liturgyoffice.org.uk/Calendar/2020/Sep.shtml) plus the title _Priest_.

On St Agnes Tsao Kou Ying: According to [Wiki](https://en.wikipedia.org/wiki/Agnes_Tsao_Kou_Ying), both should be correct, but _Tsao Kou Ying_ might be the widely used transcription of the Chinese name, but is also lists the _Cao Guiying_ as an alternative transcription.

---

I’d like still fix the comma issue in `sk.js` and probably even #174.

src/locales/en.js
```diff
@@ -290,7 +290,7 @@ export default {
     "saintBernardineOfSienaPriest": "Saint Bernardine of Siena, Priest",
     "saintBernardOfClairvauxAbbotAndDoctorOfTheChurch": "Saint Bernard of Clairvaux, Abbot and Doctor of the Church",
     "saintBeunoAbbot": "Saint Beuno, Abbot",
-    "saintBlaseBishopAndMartyrSaintAnsgarBishop": "Saint Blase, Bishop Martyr and Saint Ansgar, Bishop",
+    "saintBlaseBishopAndMartyrSaintAnsgarBishop": "Saint Blase, Bishop Martyr or Saint Ansgar, Bishop",
```
emagnier: You are changing the `and` to `or` in the localised title, but not in the key?
Also if it should be an `or` instead of an `and`, we should have 2 different keys (and 2 different celebration date) instead.

src/locales/en.js
```diff
@@ -418,7 +417,6 @@ export default {
     "saintHedwigOfPoland": "Saint Hedwig of Poland",
     "saintHedwigOfSilesia": "Saint Hedwig of Silesia",
     "saintHedwigReligiousOrSaintMargaretMaryAlacoqueVirgin": "Saint Hedwig, Religious Or Saint Margaret Mary Alacoque, Virgin",
-    "saintHedwigReligiousSaintMargaretMaryAlacoqueVirgin": "Saint Hedwig, Religious/Saint Margaret Mary Alacoque, Virgin",
```
emagnier: Is it an `or` or an `and` now? I'm a bit confused.

emagnier (changes requested):
Also I don't know what's the ETA on the `_F-TypeScript` branch. In this branch, there is a major change in the codebase, so everything merged outside this branch might be difficult then to merge back in this branch.
So here I wonder if you should open this PR against the `_F-TypeScript` branch instead, or better: wait this TS branch get ready and at least get merged into `dev` (to avoid the same scenario of the `romcal-es6` branch).

src/locales/en.js
tukusejssirs: You are right, I forgot to rename the key.

> Also if it should be an `or` instead of an `and`, we should have 2 different keys (and 2 different celebration date) instead.

I know how to define two keys, but how do I define a celebration (in the calendar files) with two keys ATM? I know we want to implement it, but is it already implemented? If it is not yet implemented, we should have one key (as it is right now, but replace `or` with `and`) until it is implemented.

src/locales/en.js
tukusejssirs: Here I have removed the duplicate key `saintHedwigReligiousSaintMargaretMaryAlacoqueVirgin` and kept the `saintHedwigReligiousOrSaintMargaretMaryAlacoqueVirgin` key only.

And to answer your question explicitly: it is `or`, as there should be two celebrations. If you answer [my previous comment]() and provide me a way to define two celebrations for one day, I’ll do it that way, otherwise this should be okay (according to the old romcal rules).

tukusejssirs:
> Also I don't know what's the ETA on the `_F-TypeScript` branch. In this branch, there is a major change in the codebase, so everything merged outside this branch might be difficult then to merge back in this branch.
> So here I wonder if you should open this PR against the `_F-TypeScript` branch instead, or better: wait this TS branch get ready and at least get merged into `dev` (to avoid the same scenario of the `romcal-es6` branch).

Okay, I’ll port the changes to `_F-TypeScript`.

src/locales/en.js
emagnier: Yes it's now implemented. You now can simply create 2 celebrations with the same date, in the same calendar. And romcal will output all the suitable celebrations for the corresponding date (including all optional memories if it's appropriated).

tukusejssirs: Okay. And is it implemented in the `_F-TypeScript` branch too?

src/locales/en.js
emagnier: Good 😉

src/locales/en.js
emagnier: Yes.

PR #179, opened 2020-03-04
 Fix some localisation issues

tukusejssirs:
- Add St Paul IV, Pope celebration to `general.ts`;
- Add St Paul IV, Pope key to `en.ts` and `sk.ts`;
- Remove `saintsCosmasAndDamian` in favour of `saintsCosmasAndDamianMartyrs`;
- Remove `saintHedwigReligiousSaintMargaretMaryAlacoqueVirgin` in favour of `saintHedwigReligiousOrSaintMargaretMaryAlacoqueVirgin`;
- Remove `saintJohnOfGodPriest` in favour of `saintJohnOfGodReligious`;
- Remove `saintJuanDiego` in favour of `saintJuanDiegoCuauhtlatoatzin`;
- Remove `saintMarkApostle` in favour of `saintMarkTheEvangelist`;
- Remove `saintOliverPlunketBishopAndMartyr` in favour `saintOliverPlunkettBishopAndMartyr`;
- Remove `saintPioOfPietralcina` and `saintPioOfPietrelcinaPadrePioPriest` in favour of `saintPioOfPietrelcinaPriest` (with the `en.ts` value changed to `Saint Pio of Pietrelcina`);
- Remove `saintEzekielMorenoBishop` in favour of `saintEzequielMorenoBishop`;
- Remove `saintLorenzoRuizAndCompanionsMartyrs` and `saintsLawrenceRuizAndCompanionsMartyrs` in favour of `saintLawrenceRuizAndCompanionsMartyrs`;
- Rename `saintWenceslausMartyrSaintsLawrenceRuizAndCompanionsMartyrs` to `saintWenceslausMartyrSaintLawrenceRuizAndCompanionsMartyrs`;
- Rename `saintClementOfOhridAndGorazdBishopsAndCompanions` to `saintsClementOfOhridAndGorazdBishopsAndCompanions`;
- Rename `saintAgnesCaoGuiyingMartyr` to `saintAgnesTsaoKouYingMartyr` (and change the `en.ts` value);
- Fix apostrophes in l10n files;
- Add missing keys from `en.ts` to `sk.ts`;
- Rename `saintBlaseBishopAndMartyrSaintAnsgarBishop` to `saintBlaseBishopOrMartyrSaintAnsgarBishop` (change _And_ to _Or_) and change its value in `en.ts` and `sk.ts`;
- Remove grammatically incorrect commas in `sk.ts`;
- Join the key name and value into one line in the locale files.


tukusejssirs:
My notes are the same as I wrote in [this comment of mine](https://github.com/romcal/romcal/pull/178#issuecomment-594183853) in #178.

In addition to the changes I have made in #178, I have joined the key name and value into one line in the locale files.

emagnier:
> I have joined the key name and value into one line in the locale files.

[Prettier](https://prettier.io/) will probably put them back in 2 lines, because it consider they are too long.

In your terminal, you can test it by doing that in your romcal dir:
```
npx prettier --write src/**/*.ts
```

tukusejssirs:
Now at least I know why they are split into two lines (to my disliking as I prefer soft wrapping).

On the other hand, my next task is to separate the locale keys of celebrations that occur on the same day, therefore at least most of these splitted lines (if not all) would end up on one line. I am already splitting the keys in the `en.ts`; when I split the keys, I’ll redefine the celebration definitions in the calendars. I would take some time, but only after that I’d like to merge this PR.

pejulian:
Hi all,

I have added two scripts:

`npm run prettier:check`
`npm run prettier:fix`

to `package.json`.

Running the first one will inform you if there are files not compliant with our `.prettierrc.js` settings and running the second will fix the errors reported.

On top of this, we have a `npm run lint` script that is run on staged commits to fix any issues,

src/locales/en-CA.ts
```diff
@@ -2,7 +2,6 @@ import { TRomcalLocale } from "../models/romcal-locale";
 
 export default {
     sanctoral: {
-        saintsJeanDeBrebeufAndIsaacJoguesPriestsAndCompanionsMartyrsSaintPaulOfTheCrossPriest:
-            "Saints John de Brébeuf, Isaac Jogues, Priests, and Companions, Martyrs, Secondary Patrons of Canada",
+        saintsJeanDeBrebeufAndIsaacJoguesPriestsAndCompanionsMartyrsSaintPaulOfTheCrossPriest: "Saints John de Brébeuf, Isaac Jogues, Priests, and Companions, Martyrs, Secondary Patrons of Canada",
```
pejulian: We really don't have to manually edit instances such as these because `prettier` will do the work for us on commit (or directly in our IDE's if we have the correct setup). If the outcome of the auto formatter is not desirable, we can agree on new settings that do. In this particular instance, the setting that is in question is [printWidth](https://prettier.io/docs/en/options.html#print-width). The documentation recommends a print width of 80. I opted for 120 because of the nature of our code. However, we can agree on a wider width if this is insufficient (as in the case of this celebration). However, I would advise against this as _overall_ legibility of the code will be impacted (especially in smaller screens) and that cases such as these are rare in the context of our code base.

pejulian:
@tukusejssirs 

Apart from my minor comments about formatting, LGTM!
I've merged your changes into the TS branch!

src/locales/en-CA.ts
tukusejssirs: Okay, I get it.

When I separate the celebrations occurring on the same day, the keys and their values should be shorter. Then we might want to change the `printWidth` to the recommended `80`.

> However, I would advise against this as overall legibility of the code will be impacted (especially in smaller screens) and that cases such as these are rare in the context of our code base.

This depends on the wrapping. I always use soft-wrapping and then this is not an issue—not even on the phones. :smiley:

emagnier: I'm fine with the current `printWidth: 120`, I think this value is a good balance for our codebase. Because reducing the width too much often doesn't help to make code readable.

Issue #181, opened 2020-03-07
Maintenance and bug fixes for new Typescript codebase


pejulian:
Shalom @emagnier and @tukusejssirs!

Here's a list of issues that I suggest we look into (in no particular order) before the `_F-TypeScript` branch is merged to `dev`:

- [x] Revise locale management to make it behave as [here](https://github.com/romcal/romcal/blob/master/src/lib/Utils.js#L9) in `master`. Keep output messages in the console but downgrade error messages to warnings.
- [x] Revise variable names (especially type declarations) to be more consistent and meaningful
- [x] Add more `tsdoc` compliant comments to code to improve quality of API documentation
- [x] Fix known issues with TS code as described [here](https://github.com/romcal/romcal/pull/169#issuecomment-590032562)
- [x] Cleanup folder structure and file naming conventions
- [x] Cleanup `README.md`
- [x] Cleanup of `romcal` consumer facing response (#180).
- [x] Consistency of configuration options (#163)

Suggestions to make this list achievable and this issue "close-able" in a reasonable amount of time:

1. Let's keep the list concise and targeted mostly to bug fixes.
2. If there's a need for new features to be listed here, they should be for better usability of existing features (i.e. #180). New functionality and features should be done at another time.
3. Lets keep this issue list short and sweet and *link* to extended discussion in other issues.

Considerations:

- I started transforming some of the comments in the codebase to be `tsdoc` compliant. However, it is far from completion. Better `tsdoc` compliant comments in our code = better [API documentation](https://romcal.github.io/romcal/) for our consumers.

Let me know your thoughts!

tukusejssirs:
> Revise variable names (especially type declarations) to be more consistent and meaningful

Don’t forget #163 please! :smiley:

>  Cleanup folder structure and file naming conventions

What do you think about my suggestion on the division of the l10n files into multiple files (for more info see [this comment](https://github.com/romcal/romcal/issues/177#issuecomment-595969610))?

>  Update `README.md` with the latest behaviour of the code.

Don’t we wanna move the documentation to GitHub Wiki?

pejulian:
> Don’t forget #163 please! 😃

Added to the list! 😄 

> What do you think about my suggestion on the division of the l10n files into multiple files (for more info see this comment)?

The suggestion definitely has merits, but I'd vote to implement it as a new feature in a future milestone as it does there is some technical impact to the how stuff works now.

> Don’t we wanna move the documentation to GitHub Wiki?

[...]

PR #189, opened 2020-03-21
Add locale for brazilian portuguese

HenriqueDerosa:
## More details
I added a few items to a new file called `pt-BR` where I put some of the main texts to say `it works in Portuguese`. 

The only thing left is the sanctoral, it's not complete yet, but it's a good start to have it in Portuguese.

If I need to make any changes, just tell me. I really loved these repositories of `romcal`.  Thanks

### DIFF
--- a/src/locales/index.js
+++ b/src/locales/index.js
@@ -4,6 +4,7 @@ import fr from './fr';
 import it from './it';
 import pl from './pl';
 import sk from './sk';
+import ptBR from './pt-BR';
 
 export {
   enCA,
@@ -12,4 +13,5 @@ export {
   it,
   pl,
   sk,
+  ptBR,
 };
--- a/src/locales/pt-BR.js
+++ b/src/locales/pt-BR.js
@@ -1,0 +1,95 @@
+export default {
+  "advent": {
+    "season": "Advento",
+    "feria": "{{day}} of the {{week}} week of Advento",
+    "sunday": "{{week}} Domingo do Advento"
+  },
+  "christmastide": {
+    "season": "Natal",
+    "day": "{{day}} do Tempo do Natal",
+    "octave": "{{count}} dia da Oitava de Natal",
+    "sunday": "{{count}} Domingo do Natal"
+  },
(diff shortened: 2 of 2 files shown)


tukusejssirs:
@HenriqueDerosa, thank you for your contribution! However, we are in the middle or refactoring of the code (we move it entirely into TypeScript), therefore it would be awesome if you could create a PR against `_F-TypeScript` branch? :smiley:

emagnier:
🎉 Awesome! Thanks @HenriqueDerosa for your contribution!

We could merge it here. There is no breaking changes, it's only about locales on limited files, and that one is easy to back port on our working branches.

The `_F-TypeScript` is even not ready yet for development usage (with no ETA for the time being).

emagnier (approved):

tukusejssirs:
@emagnier, I thought we want to work on the TS branch primarily, that’s why I suggested to rebase the changes.

emagnier:
Yes, in fact it depends on the type of contributions. 
And also because I don't see the work from the TS branch out soon (there is still a lot of work to do, with a required testing phase for all kinds of usages). 

The rewrite of the codebase is currently a special situation, so for the time being, we might have to discern on a case-by-case basis. Here it's a simple contribution (with a great value!), that don't have too much impact on our codebase 😄

emagnier:
(ping FYI @pejulian)

pejulian:
Happy Easter everyone!

I've been away for awhile due to other commitments.

However, I hope to be able to resume work with the TS branch soon. I think we should consider cutting a working version of the TS branch and have that published for general use before proceeding with the other changes discussed. That way, the new implementation won't be stuck in development phase forever. 

The new version we cut should address any critical bugs in the new code base but otherwise retain most of the functionality as before.

What do you guys think about that?

emagnier:
Hey @pejulian, thanks for your message! Maybe we could continue to discuss about that here https://github.com/romcal/romcal/issues/181#issuecomment-621197401, so it's easier to keep the track?

Issue #256, opened 2021-01-01
Fix some issues in French locale


scriptura:
Hello to the two developers who are contributing to this project.

I did a comparable project (much less advanced in functionality) on a private repository with luxon.js and .json before discovering yours halfway, I still looked at your project in order to see next to what liturgical rules I had been able to pass and that was very helpful to me (and especially the architectural design errors to avoid).

I am French, sorry for my very approximate English, I translate with google translate.

Here are some suggested fixes in French. Your translations in French are not really wrong, but not always good, or else it is not the custom (a bit like me in English, here). I am not talking about Canada, this is only about usage in France.

So :
```
romcal/src/locales/fr.js
"Temps de Pâques" -> "Temps Pascal"
"Saint Sacrement" -> " Le Saint Sacrement"
"Dimanche de Pâques" -> "Résurrection du Seigneur"
"Marie, Mère de Dieu" -> "Sainte Marie, Mère de Dieu"
"Dimanche des Rameaux et de la Passion" -> "Dimanche des Rameaux et de la Passion du Seigneur"
"Très Sainte Trinité" -> "Sainte Trinité"
"Premiers Saints martyrs de la Sainte Église romaine (✝ 64)" -> "Premiers martyrs de l'Église de Rome (✝ 64)"
```
There you go... Thanks for creating this great tool.

tukusejssirs:
Thanks, @Scriptura!

Actually, there are much more issues with the French localisation, however, I don’t speak French (but I understand a bit) and currently I am busy with processing the text of the Martyrology Romanum (see #208). @emagnier (who speaks French) is busy at work / with other stuff, but he wishes to release the new dev version of romcal ASAP.

Anyway, the latest dev code is in the `_F-TypeScript` branch. I see you linked `fr.js` file, while in the dev code, we ported the code into TypeScript, therefore the new filename is [`fr.ts`](https://github.com/romcal/romcal/blob/_F-TypeScript/src/locales/fr.ts).

Just a note: I have checked most of 6our suggested corrections in Lectionnaire III (2016) and I found that they use `Saint-Sacrement` without the initial definite article and with a hyphen. From my point of view, in French (and in English) one might omit the initial definite article in headings/titles. We strive to localise the strings using the official publications, e.g. Roman Missal, Liturgy of the Hours, Lectionaries, Notitiæs (issued by the Holy See).

---

[...]

PR #257, opened 2021-01-01
Improvement of some French translations

emagnier:
Related to https://github.com/romcal/romcal/issues/256. Thank you @Scriptura for your contribution!


src/locales/fr.ts
```diff
@@ -202,7 +202,7 @@ export default {
     eusebius_of_vercelli_bishop: 'Saint Eusèbe de Verceil, Evêque (✝ 371)',
     fabian_i_pope: 'Saint Fabien, Pape et Martyr (✝ 250)',
     fidelis_of_sigmaringen_priest: 'Saint Fidèle de Sigmaringen, Prêtre et Martyr (✝ 1622)',
-    first_martyrs_of_the_holy_roman_church: 'Premiers Saints martyrs de la Sainte Église romaine (✝ 64)',
+    first_martyrs_of_the_holy_roman_church: 'Premiers martyrs de l''Église de Rome (✝ 64)',
```
emagnier: There is twice the apostrophe character here (this is why the automated test is failing).
Also in this case, it will be better to use this character instead: `’`

I guess you can do the correction directly on your branch, here: https://github.com/Scriptura/romcal/blob/patch-1/src/locales/fr.ts (from there, your fix will be included automatically in this PR).

emagnier (commented):
I found only one minor thing that you will have to fix, before we can merge this PR.

Also if you want to update or correct more translations, you can continue to do it here: https://github.com/Scriptura/romcal/blob/patch-1/src/locales/fr.ts
Then, let us know when you think your job is done (so we know if we have to wait, or if your work can be merged).

FYI, GitHub automatically created for you a fork of the romcal project (in your GitHub profile). And then from this fork, a new `patch-1` branch, that is used especially for this PR 👍

src/locales/fr.ts
scriptura: OK. I'm going to try...

scriptura: Here ! And now I know why I see oblique quotes so often in French translations of English texts when we don't have the use of them;)
Sorry, I was sure a double quote would escape the character ...

emagnier: Yes... not in javascript. Thanks for the fix!

emagnier:
Do you plan to do more work for the time being?
If no, I will merge your work (future improvements can always be done later in a future PR).

Also there is still the warning about a failing check, but this is because of a failing test that have nothing to do with this PR, I will fix it from my side.

tukusejssirs:
> FYI, GitHub automatically created for you a fork of the romcal project (in your GitHub profile). And then from this fork, a new `patch-1` branch, that is used especially for this PR 👍

@emagnier, just a sid note: it creates a new branvh when one select to create a new branch when committing the changes to a file. Personally, I prefer creating a branch beforehand, because that way I can name it more reasonable.YMMV :smiley:

src/locales/fr.ts
```diff
@@ -70,7 +70,7 @@ export default {
     baptism_of_the_lord: 'Baptême du Seigneur',
     christ_the_king_sunday: 'Christ, Roi de l’Univers',
     christmas: 'Nativité du Seigneur',
-    corpus_christi: 'Saint-Sacrement',
+    corpus_christi: 'Saint Sacrement',
```
tukusejssirs: I’ll leave it up to @emagnier to decide which form to use, however, Lectionaire III uses the hyphenated form.

src/locales/fr.ts
tukusejssirs: > Here ! And now I know why I see oblique quotes so often in French translations of English texts when we don't have the use of them;)
Sorry, I was sure a double quote would escape the character ...

@Scriptura, well, while that might be a partial reason to the usage of [typographical apostrophe](https://en.m.wikipedia.org/wiki/Apostrophe#Typographic_form) (you call the oblique apostrophe), my personal reason is that I prefer it over the typewriter apostrophe as it looks better. Some might argue that the client application should change the apostrophe form rather than hard-code it within the code.

tukusejssirs (commented):
@Scriptura, @emagnier, I’ve dropped some comments as a review, however, there is only one potential issue.

src/locales/fr.ts
scriptura: For info: normally hyphens in French are only used for a place or a building, and also in the case of a few word contractions.

For example, here is the name of a town in France: "Saint-Médard-en-Jalles".

But of course, we French love the exeptions! Suddenly we will find "Fête-Dieu" or "Fête Dieu", "Saint-Sacrement" or "Saint Sacrement, etc. Both writings are therefore possible a priori. On the AELF they remove the dashes. I'll put it back on my side so it will stick better to your book.

PS: we have lots of things like that in French, which are often a matter of personal preference: this is already the case with the word "scolastique", but "scholastique" is also possible and closer to the original Latin word, d 'where my correction for the latter (by the way in English you also kept the "h").

scriptura:
> Do you plan to do more work for the time being?
> If no, I will merge your work (future improvements can always be done later in a future PR).

Unless you plan to update your code next, leave the PR open for at least 15 days. I am on my code, if I see things that are not said in French I will make suggestions for corrections on this fortnight.

src/locales/fr.ts
tukusejssirs: I understand that the hyphenation is optional. There are quite similar rules in English too. :smiley:

On the spelling of St Scholastica, I agree it looks better with the `h`. Also the same spelling is used in Lectionaire III (the same thing @emagnier would share with you privately if you want to).

tukusejssirs:
@emagnier, I have no idea why this PR fails the Travis CI build. It fails one test only: the colour of _Chair of St Peter_. I looked at `fr.ts` if there is a missing comma or something, if the key name was changed; or if the colour definition in `general.ts` or the test `src/lib/calendar-builder.test.ts` was changed. I found nothing.

```bash
  ● Testing calendar generation functions › Testing liturgical colors › The proper color for the Chair of Peter and the Conversion of St. Paul is white, although both St. Peter and St. Paul were martyrs.

    expect(received).toBe(expected) // Object.is equality

    Expected: 2
    Received: 1

      268 |         ['chair_of_saint_peter_the_apostle', 'conversion_of_saint_paul_the_apostle'].includes(d.key),
      269 |       );
    > 270 |       expect(dates.length).toBe(2);
          |                            ^
      271 |       dates.forEach((d) => expect(d.liturgicalColors[0]).toEqual(LiturgicalColors.WHITE));
      272 |     });
      273 |   });

      at Object.<anonymous> (src/lib/calendar-builder.test.ts:270:28)
```

src/locales/fr.ts
scriptura: > On the spelling of St Scholastica, I agree it looks better with the `h`. Also the same spelling is used in Lectionaire III (the same thing @emagnier would share with you privately if you want to).

OK.

emagnier:
> Unless you plan to update your code next, leave the PR open for at least 15 days. I am on my code, if I see things that are not said in French I will make suggestions for corrections on this fortnight.

@Scriptura: I would like to work on the codebase, but I'm not able to find time in a short term... So take your time 👍

> I have no idea why this PR fails the Travis CI build.

@tukusejssirs: I mentioned a few weeks ago that travis tests wasn't working correctly (in contrary of this situation, it was displaying a green light while there was a buid issue). The problem doesn't come from this PR, but from a previous code update. But now finally Travis work as expected (we just didn't catch this issue before). I'll fix this test soon.

scriptura:
Just a hello to tell you that I'm still here. I had started on something else, but I will come back soon to give you my proposed changes in one go.

Cordially.

scriptura:
I have finished for now. I'll let you watch the changes, feel free to postpone them depending on what you like or not.

For the record: the hyphen on the name of John Paul II in French (Jean-Paul II) would be an error due to French-speaking journalists who popularized it as such. It has remained since ...

For a future translation update, I suggest replacing the unicode apostrophes ` by `'`, even if you add an escape character or even double quotes for the string. Indeedn, even if the Anglo-Saxons think that we use - and the CMS such as WordPress do not hesitate to impose it on us - this is not the case. Up to you...

Greeting.

tukusejssirs:
I haven’t found any issue, though I haven’t consulted the French lectionary.


> For a future translation update, I suggest replacing the unicode apostrophes ’ by`'`, even if you add an escape character or even double quotes for the string. Indeedn, even if the Anglo-Saxons think that we use - and the CMS such as WordPress do not hesitate to impose it on us - this is not the case. Up to you...

@Scriptura, I don’t know if you suggest to use typewriter apostrophe (`'`) because in your opinion (1) it is used in French publications or (2) because it should be styled by the application which would display it.

Nevertheless, even in French the _typographically_ more correct glyph is the typographic apostrophe (`’`); see the following links:
- [pincetonfrancais.be](https://web.archive.org/web/20070524065337/http://www.pincetonfrancais.be/v1/article.php3?id_article=30) (found via [this comment](https://www.proz.com/forum/sdl_trados_support/115468-curly_vs_straight_apostrophe_in_french.html#945911));
- [typedrawers.com](https://typedrawers.com/discussion/2032/french-apostrophe) (although it talks about apostrophe kerning and spacing, they all use typographic apostrophes);
- [French Wikipedia (of Breton language though)](https://fr.m.wikipedia.org/wiki/Apostrophe_(typographie)#En_breton) (more on the [modifier letter apostrophe](https://en.m.wikipedia.org/wiki/Modifier_letter_apostrophe)):
   > L’utilisation de la lettre modificative apostrophe (ʼ, U+02BC) est recommandée par l’Office public de la langue bretonne, mais l’usage de l’apostrophe dactylographique (U+0027) ou de l’apostrophe typographique (U+2019) reste courant.

---

That said, it might be better to rely on the application that uses romcal output to style the apostropnes; what do think, @emagnier?

emagnier:
Great! These modifications are very good improvements!
I will try to do a closer look and do the merge tonight.

About the curly/curved apostrophe, I know it's a matter of debate, but I prefer to keep it since it's traditionally used in French (and still recommended in the syntactic rules). Depending of fonts it also generally looks better (often used by designers, even in English). The use of the straight apostrophe became popular because of the typewriters (invented by English speaking) which contained only this straight character (to simplify the number of keys). And unfortunately this key was not reintroduced on computer French keyboards. This is why in French we often see both chars. But for example, Microsoft Word or Libre Office automatically use the curly apostrophe (instead of the straight char). I found a very interesting article (in French) about the use of this char: https://orthogaffe.wordpress.com/2010/02/21/point-apostrophe/

tukusejssirs:
@emagnier, I fully agree with the article! Thanks for sharing it!

scriptura:
I am aware of the history of the English apostrophe imposed by the typewriter, some French speakers like me are aware of it. But it is also this kind of small detail that makes history, which influences mentalities much more than you think. Search major French-language dailies, [such as this one,](https://www.lemonde.fr/) or even any of all French-language dailies. You will not find any "French" apostrophe. Personally, I don't even know how to use the shortcut to get this apostrophe.

Afterwards, all this remains a detail, but it's funny to talk about it with English speakers ;)

Bien cordialement.

emagnier:
Thanks again for your feedback! Indeed it's a matter of debate :) And I'm French as well, sharing about this topic in English :)

By "French" apostrophe I suppose we are talking about the curly or curved apostrophe (unicode 8217)? I don't know how you see the website that you gave as an example, but on my side (on mac) I see all the apostrophes as curved, at least on all serif fonts. I did a double check with a unicode search and it matches well the curved apostrophe char.

Screenshot 2021-01-22 at 22 02 59

https://www.lemonde.fr/international/article/2021/01/22/l-accord-d-investissements-ue-chine-devoile_6067291_3210.html

emagnier:
It's now merged 🍻
A big thank you for your contribution on improving the french translation 🇫🇷.
You are more than welcome if you see anything else, or just to share ideas.

And thanks for the interesting discussion about apostrophe char. I know that things can change, and the history is never done, but for now I still prefer to keep the curly one. But it was interesting to share about that 😃.

scriptura:
Ah yes, indeed ... I was content to do a CTRL + F4 but the forward apostrophe is still located even if we look for the right. So I just went to see a lot of sites with this new look, and I just learned something that I thought I had learned in web typography.

Thanks for the exchange. Cordially.

Issue #270, opened 2021-04-04
Output celebrations in a array by date


tukusejssirs:
It might be too late for v2.0.0 to change it, but I think we should think wnd talk about this for v3.0.0.

Currently, we output every celebration as a separate (let’s say) object.

I propose we should create an array of sequential dates limited by start end time (like start/end of a year). Each date would be an array itself and would contain an array of celebrations.

This way, the main array (array of dates) for civil years would contain either 365 or 366 elements. Also we would eliminate the duplicate definition of dates for days where there are two or more celebration options.

emagnier:
By default, romcal output exactly 365 (or 366) date items for a civil year, so indeed, only one item per date.

When we need to output all optional memorials, so potentially multiple date items per dates, we need to set `outputOptionalMemorials` to `true` (this option is `false` by default).

I'm not really convinced to change the default returned array to an object of dates. Because on a lot of use cases, like filtering or querying specific criteria, the developer will have to convert it back to an array. And in terms of performances this isn't ideal (on this specific topic, we already have a lot on our plate some).

By the way, on romcal v2, there is already methods to address your specific wish:
https://github.com/romcal/romcal/blob/_F-TypeScript/docs/general-usage.md#-group-by-a-predefined-criteria-groupbycriteria

But generally, I will instead encourage the use of the [`.reduce`](https://github.com/romcal/romcal/blob/_F-TypeScript/docs/general-usage.md#-group-by-any-criteria) method in JavaScript. This is now natively supported everywhere in this coding language, and will allow addressing all specific needs a developer can have for its development requirements.
This is for the same reasons why I've redesigned the [filtering](https://github.com/romcal/romcal/blob/_F-TypeScript/docs/general-usage.md#filter-by-date) options from romcal < v2, since it's already possible to do it natively (and in a simpler and explicit way). I only kept the `.groupBy` method for amateur developers (because it might be a bit more difficult for them, and to provide syntactic sugar since it take a bit more code).

tukusejssirs:
> By default, romcal output exactly 365 (or 366) date items for a civil year, so indeed, only one item per date.
> 
> When we need to output all optional memorials, so potentially multiple date items per dates, we need to set `outputOptionalMemorials` to `true` (this option is `false` by default).

[...]

PR #273, opened 2021-05-18
Output all available object per days (including optionals) & Introduce `strictMode`

emagnier:
@tukusejssirs we add the discussion (https://github.com/romcal/romcal/pull/250#issuecomment-735299623, and here https://github.com/romcal/romcal/issues/270#issuecomment-814001842) about output by default all available celebrations per day. The more relevant first, followed by all optional memorials or commemorations.

Actually, only one liturgical day object is output per day. To get other optional objects, you need to set `outputOptionalMemorials: true`.

---

This PR proposes to remove `outputOptionalMemorials`, and introduce a new `strictMode` option.

- `false` (default): everything that match a specific day is output. The more relevant first, followed by optional memorials or commemorations.
- `true`: only the relevant liturgical day is output. So only one object per day. Which means exactly 365 objects in non-leap years.

@tukusejssirs: What do you think?


tukusejssirs (approved):
@emagnier, LGTM. Thanks for the PR and that you agreed to implement this! :+1: :wink:

PR #274, opened 2021-05-18
Add orderBy dates syntactic sugar

emagnier:
Fix https://github.com/romcal/romcal/issues/270.

Romcal will always output data as an array.
This `.groupBy('dates')` method is a syntactic sugar, that allows to quickly group outputted results by dates.

```js
const data = await Romcal.calendarFor({ year: 2021 });
const groupedByDates = data.groupBy('dates');
```

This also can be addressed natively by JavaScript (using a `reduce` method). But I think it will be a convenient method for any romcal users that are not too comfortable with JavaScript.

---

One thing I'm wondering, should the `groupBy` criteria should be singular or plural?
Here is the actual list of `groupBy` options: https://github.com/romcal/romcal/blob/_F-TypeScript/docs/general-usage.md#group-results-by-criteria

### DIFF
--- a/docs/general-usage.md
+++ b/docs/general-usage.md
@@ -356,13 +356,13 @@ var easterDate = calendar.getDate('easter_sunday'); // "2020-04-12T00:00:00.000Z
 
 romcal offers a convenient way to group data by various criteria. The supported criteria are:
 
-`dates` | `days` | `months` | `daysByMonth` | `weeksByMonth` | `sundayCycles` | `weekdayCycles` | `ranks` | `liturgicalSeasons` | `liturgicalColors` | `psalterWeeks`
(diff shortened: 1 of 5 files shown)


tukusejssirs (approved):
LGTM. Thanks! :+1:

---

> One thing I'm wondering, should the `groupBy` criteria should be singular or plural?

I’d suggest to use singular. It is at least my personal preference. :wink:

> ```js
> const data = await Romcal.calendarFor({ year: 2021, outputOptionalMemorials: true });
> const groupedByDates = data.groupBy('dates');
> ```

You might know this, but still: you need to remove the `outputOptionalMemorials` option. :wink:

emagnier:
Thank you. I've renamed the `groupBy` criterias from plural to singular. I also think it's better.

And the `outputOptionalMemorials` doesn't exist anymore.

Issue #247, opened 2020-10-20
Rename the `source` key used in celebration definitions


tukusejssirs:
I would suggest to rename the `source` key used in celebration definitions to something else, as IMHO a `source` should be the source from where we got the celebration metadata (see #234).

emagnier:
I agree, it's quite confusing.
It's actually directly related to the l10n files: it corresponds to the namespace where to find the value of a locale key.

Better tooling to manage i18n in general are still on the todo list. But in the meantime we could find a better and less confusing name.

What about `localeNamespace`? @tukusejssirs what do you think?

tukusejssirs:
> I agree, it's quite confusing.
> It's actually directly related to the l10n files: it corresponds to the namespace where to find the value of a locale key.
>
> Better tooling to manage i18n in general are still on the todo list.

Indeed, a request for a l10n tool was one of my first issues I created on romcal (see #43). :wink: The issue description is not updated to my latest suggestions though, therefore one needs to read the whole discussion. Basically: now I suggest to use _Martyrologium Romanum_ (2004) as the base for the l10n tool data + many tables I created on [Google Sheets](https://docs.google.com/spreadsheets/d/1huvnEXr9QiCUT3-TuMIxxxG4xggMqG-Vq1E0QwVmBY8). I am in the process of proofreading the OCR’ed version, then I need to fill those tables with all the data from MR2004, finally need to search for the rest. I want to finish the proofreading by the end of August 2021 and filling the data from MR2004 to the tables by the end of 2021. Hopefully, by the end of 2022 we could have filled all the tables in Latin, English, Slovak, Czech, and with your help French. If we (I) find some contributors who would help me, we could accomplish this sooner. One thing I need to know before I start filling the tables, is how should I do this: should I fill the tables on Google Sheets or use a MongoDB database or how? I need to know this, because I wish we could create a `sanctoral` package (feel free to suggest a better name) with all these tables/data. At build time, romcal could select (somehow) just those `sanctoral` data which is required _by the builder_ (person), like only for specific calendars (e.g. only for `general`; only for French-speaking countries; only for European countries; only for a list of specific countries) or up to a calendar level (see calendar inheritence; e.g. only up to national calendars). This could make the build romcal smaller for those who require it, while I could build it with all the available `sanctoral` data (ideally, I’d like to run romcal from the repository, while to JS/TS script would not be required to run in the repo root).

Sorry for being off topic a bit. :wink:

[...]

PR #282, opened 2021-05-18
On calendar definition, rename the source prop to namespace

emagnier:
Closes https://github.com/romcal/romcal/issues/247.

On calendar definition, rename the locale `source` prop to `namespace`.


src/models/liturgical-day/liturgical-day.types.ts
```diff
@@ -195,7 +195,7 @@ export type LiturgicalDayMetadata = {
 };
 
 /**
- * The source of the liturgical day.
```
tukusejssirs, line 198: L198-206 seems to duplication of L125-133. We might want to be more specific in order to better document both parts of the code.

That said it is not a major issue. wink:

tukusejssirs (approved):

tukusejssirs (approved):

src/models/liturgical-day/liturgical-day.types.ts
emagnier: I kept the documentation on the object property (useful when editing a calendar definition, because smart IDEs adds text hint on the currently edited property).

I've reduced the documentation on the object type itself.

PR #309, opened 2021-10-08
Add `DoctorOfTheChurch` title to `irenaeus_of_lyon_bishop`

tukusejssirs:
According the [Catholic News Agency](http://www.catholicnewsagency.com/news/249215/pope-francis-to-declare-st-irenaeus-a-doctor-of-the-church-with-title-doctor-of-unity), Pope Francis _will_ name St Irenaeus of Lyon a doctor of the Church.

I don’t know if we should already update romcal, but the PR should be ready, unless there is something else needed to be done.


emagnier:
For sure, we can update romcal! 

This is just the NPM release that still need to be done manually, until a GitHub Action does it for us.

emagnier (approved):

PR #330, opened 2021-11-23
Add allowSimilarRankItems option on liturgical day definitions

### LINKED ISSUE #172: Immaculate heart of Mary is Memorial ?
`immaculateHeartOfMary` have type rank set to `FEAST`. But I think it should be `MEMORIAL` (at least it is in France and United-States). But I guess it's an error and it should be a `MEMORIAL` in the `general.js` calendar as well.

Futhermore there is a comment just above mentioning `// Memorials`.

emagnier:
Fixes #172.

This also introduce a new `LiturgicalDay` definition parameter: `allowSimilarRankItems: boolean`.

When this option is enabled, in addition to the current liturgical day, similar items can also be output.
The similar items must have the same rank, and a precedence equal or lower to the current item.

This makes possible to output **Immaculate heart of Mary** while keeping existing memorials on this day.
- 2014 June 28: **Immaculate heart of Mary** + **Saint Irenaeus**
- 2015 June 13: **Immaculate heart of Mary** + **Saint Anthony of Padua**
- 2022 June 25: **Immaculate heart of Mary only**, because there are no other memorials.

### DIFF
--- a/lib/general-calendar/proper-of-saints.ts
+++ b/lib/general-calendar/proper-of-saints.ts
@@ -442,8 +442,8 @@ export class GeneralRoman extends CalendarDef {
       precedence: Precedences.GeneralMemorial_10,
       // The Saturday, after the Solemnity of the Most Sacred Heart of Jesus
       dateDef: { dateFn: 'immaculateHeartOfMary' },
-      // If this liturgical day fall the same day of another memorial, do not
-      // replace this another memorial, but keep them both.
+      // If this liturgical day and a memorial occur on the same day,
+      // do not replace the latter, but output them both.
       allowSimilarRankItems: true,
       colors: Colors.White,
       properCycle: ProperCycles.ProperOfTime,
--- a/lib/general-calendar/proper-of-saints.ts
+++ b/lib/general-calendar/proper-of-saints.ts
@@ -442,6 +442,9 @@ export class GeneralRoman extends CalendarDef {
       precedence: Precedences.GeneralMemorial_10,
       // The Saturday, after the Solemnity of the Most Sacred Heart of Jesus
       dateDef: { dateFn: 'immaculateHeartOfMary' },
+      // If this liturgical day fall the same day of another memorial, do not
+      // replace this another memorial, but keep them both.
+      allowSimilarRankItems: true,
       colors: Colors.White,
       properCycle: ProperCycles.ProperOfTime,
     },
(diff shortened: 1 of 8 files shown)


lib/general-calendar/proper-of-saints.ts
```diff
@@ -442,6 +442,9 @@ export class GeneralRoman extends CalendarDef {
       precedence: Precedences.GeneralMemorial_10,
       // The Saturday, after the Solemnity of the Most Sacred Heart of Jesus
       dateDef: { dateFn: 'immaculateHeartOfMary' },
+      // If this liturgical day fall the same day of another memorial, do not
+      // replace this another memorial, but keep them both.
```
tukusejssirs: Don’t want to nitpick on this, but here’s my suggestion. Otherwise, fix at least `fall` → `falls`. :wink:

```suggestion
      // If this liturgical day and a memorial occur on the same day,
      // do not replace the latter, but output them both.
```

tests/dates.test.ts
```diff
@@ -468,6 +468,18 @@ describe('Testing specific liturgical date functions', () => {
         expect(rangeContainsDate(range, immaculateHeartOfMary)).toBeTrue();
       }
     });
+
+    test('When it falls the same day as another memorial, both memorials are kept', async () => {
```
tukusejssirs:
```suggestion
    test('When it occurs on the same day as another memorial, both memorials are output', async () => {
```

tests/dates.test.ts
```diff
@@ -468,6 +468,18 @@ describe('Testing specific liturgical date functions', () => {
...
+      const romcal = new Romcal();
+      const keys = (await romcal.generateCalendar(2015))['2015-06-13'].map((d) => d.key);
+      expect(JSON.stringify(keys)).toEqual(JSON.stringify(['immaculate_heart_of_mary', 'anthony_of_padua_priest']));
+    });
+
+    test('When it falls on a weekday, only one item is output for this day', async () => {
```
tukusejssirs:
```suggestion
    test('When it occurs on a weekday, only one item is output for this day', async () => {
```

tukusejssirs (dismissed):
Some nitpicks, nothing serious. :wink:

tukusejssirs (approved):

emagnier:
Thank you for the review 😉

Issue #296, opened 2021-09-29
Release NPM versions of romcal


emagnier:
The `/scripts/` files from romcal v2 are still in the codebase, but they seem not executed anymore.

@pejulian, you worked mainly on them. Can you provide us a bit more information, how they were run... ?

FYI I only removed the auto changelog, because I think a changelog must stay for humans, containing bullet points of new things, modified things, breaking changes... (before it was just a kind-of log of the commit history). 

At least, we must be able to publish new NPM package of romcal (and now, including all its calendar plugins). For the moment, we should make it manual, i.e. a maintainer (e.g. me) have to run this script manually to publish new NPM versions.

And I think it's generally better to keep it manual. So we publish new NPM versions when it's worth it (as `alpha`, `beta`, or `final` release).

I'll try to work on this script in the coming days.

tukusejssirs:
I can’t comment of the `scripts/` folder. However, IMHO automatic publishing of NPM packages based on the branch of each commit might help those who might want to quickly test the latest code and/or who don’t know how to build romcal from source (incl the installation of all the deps, like Node.js).

That said, for now it might be satisfactory to pubish them manually, but I think you should use this time to create a script/program in order to latter use it for automatic publishing, and also for tuning the process. Also I think that GitHub Actions have some kind of NPM publishing built-in.

emagnier:
Thank you for your feedback, but as I already wrote in different places, this is already the goal. I just want to go step by step to ensure we set up the right tooling / CI that address effectively our need. Please be patient ;)

pejulian:
Hi @emagnier , sorry for the late response.

The `scripts/` folder was originally created to house scripts that would be executed by Travis CI. It contains code to automatically bump the version number of the module in `package.json` and also to automated daily branch merges from `master` to `dev`. 

The whole idea was to create some sort of CD process around our code at the time. The process lacks scripts to run tests though. 

Having said that, I think it would make more sense for this process (if still required) to be migrated to GItHub Actions instead; it is more integrated and less convoluted than the scripts we currently have.

emagnier:
Hello @pejulian, thank you for your message, no worry for the delay 😄 

Yes so now it makes sense, I've removed the `.travis.yml` config file at the beginning of my work on romcal `v3`, but didn't pay attention that it also contains the instruction to run most of these scripts.

[...]

PR #332, opened 2021-11-25
New GitHub Action to automatically publish dev versions to NPM

emagnier:
This PR introduces a new GitHub action that automatically publish a new romcal version to NPM, every time a PR is merged into the `dev` branch. This will help to test any latest versions of the `dev` branch, from NPM or even CDN packages.

What does this GitHub Action do:
- It only fires when new commits are added to the `dev` branch, on the `romcal/romcal` repository.
- On an Ubuntu VM (the latest version available):
  - Checkouts the `dev` branch
  - Installs the right Node.js and NPM versions (defined in `package.json`)
  - Updates browsers list used by caniuse and babel (and the romcal build script)
  - Automatically bumps the romcal version, and saves it on the `package.json` (as `x.x.x-dev.x`)
  - Builds romcal and all calendar bundles
  - Runs tests
  - Npm publish romcal and every calendar bundles

I'm taking this opportunity to review the branch and tag names. Until here, we used the traditional `dev`/`test`/`master` branch names, and `alpha`/`beta`/`latest` NPM tag names. @tukusejssirs you suggested a few months ago to align the wordings of these branches/tags. To simplify, I propose to use:
- `dev` branch / `dev` tag (instead of the `alpha` tag)
- `beta` branch / `beta` tag (instead of the `test` branch)
- `latest` branch / `latest` tag (instead of the `master` branch)

Note 1: this PR only contain CI workflow related to the `dev` branch. GitHub Actions for the `beta` and `latest` branches will come later in another PR, and will have a slightly different behavior.

Note 2: the `package.json` version has been updated since I already made a test of this script, and published a new alpha/dev version or romcal to NPM.

Related to #296.

### DIFF
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -27,12 +27,6 @@ jobs:
       - name: Install node.js and npm
         uses: volta-cli/action@v1
 
-      - name: Install npm dependencies
-        run: npm ci --no-audit
-
-      - name: Update browsers list (caniuse)
-        run: npx browserslist@latest --update-db
-
       - name: Automated version bump
         uses: phips28/gh-action-bump-version@master
         with:
@@ -44,6 +38,12 @@ jobs:
           preid: dev
           tag-prefix:  'v'
           commit-message: 'ci: bump version to {{version}}'
+
+      - name: Install npm dependencies
+        run: npm ci --no-audit
+
+      - name: Update browsers list (caniuse)
+        run: npx browserslist@latest --update-db
 
       - name: Build romcal project (romcal + all calendar bundles)
         run: npm run build
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -36,8 +36,8 @@ jobs:
           rc-wording:
           default: prerelease
           preid: dev
-          tag-prefix:  'v'
-          commit-message: 'ci: bump version to {{version}}'
+          skip-commit: 'true'
+          skip-tag:  'true'
 
       - name: Install npm dependencies
         run: npm ci --no-audit
@@ -50,6 +50,18 @@ jobs:
 
       - name: Run tests
         run: npm run test:without-coverage
+
+      - name: Read package.json
+        uses: culshaw/read-package-node-version-actions@v1
+        id: package
+
+      - name: Commit and tag the updated version
+        uses: EndBug/add-and-commit@v7
+        with:
+          message: 'ci: bump version to ${{ steps.package.outputs.version }}'
+          tag: 'v${{ steps.package.outputs.version }}'
+          author_name: ${{ env.GITHUB_USER }}
+          author_email: ${{ env.GITHUB_EMAIL }}
 
       - name: Npm publish a new 'dev' version (romcal + all calendar bundles)
         env:
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -1,0 +1,57 @@
+name: Dev branch publishing
+
+on:
+  push:
+    branches:
+      - dev
+
+jobs:
+  dev_branch_publishing:
+    runs-on: ubuntu-latest
+
+    environment: npm-publish
+
+    if: github.repository == 'romcal/romcal'
+
+    env:
+      CI: true
+      GITHUB_USER: romcalproject
+      GITHUB_EMAIL: romcalproject@gmail.com
+
+    steps:
+      - name: Git checkout dev branch
+        uses: actions/checkout@v2
+        with:
+          ref: dev
+
+      - name: Install node.js and npm
+        uses: volta-cli/action@v1
+
+      - name: Install npm depencancies
+        run: npm ci --no-audit
+
+      - name: Update browsers list (caniuse)
+        run: npx browserslist@latest --update-db
+
+      - name: Automated version bump
+        uses: phips28/gh-action-bump-version@master
+        with:
+          major-wording: 'MajorBump'
+          minor-wording: 'MinorBump'
+          patch-wording: 'PatchBump'
+          rc-wording:
+          default: prerelease
+          preid: dev
+          tag-prefix:  'v'
+          commit-message: 'ci: bumps version to {{version}}'
+
+      - name: Build romcal project (romcal + all calendar bundles)
+        run: npm run build
+
+      - name: Run tests
+        run: npm run test:without-coverage
+
+      - name: Npm publish a new 'dev' version (romcal + all calendar bundles)
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: node -r ts-node/register scripts/publish.ts
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install node.js and npm
         uses: volta-cli/action@v1
 
-      - name: Install npm depencancies
+      - name: Install npm dependencies
         run: npm ci --no-audit
 
       - name: Update browsers list (caniuse)
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -43,7 +43,7 @@ jobs:
           default: prerelease
           preid: dev
           tag-prefix:  'v'
-          commit-message: 'ci: bumps version to {{version}}'
+          commit-message: 'ci: bump version to {{version}}'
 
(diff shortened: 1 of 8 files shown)


.github/workflows/dev-publishing.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+      - name: Git checkout dev branch
+        uses: actions/checkout@v2
+        with:
+          ref: dev
+
+      - name: Install node.js and npm
+        uses: volta-cli/action@v1
+
+      - name: Install npm depencancies
```
tukusejssirs:
```suggestion
      - name: Install npm dependencies
```

.github/workflows/dev-publishing.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Automated version bump
+        uses: phips28/gh-action-bump-version@master
+        with:
+          major-wording: 'MajorBump'
+          minor-wording: 'MinorBump'
+          patch-wording: 'PatchBump'
+          rc-wording:
+          default: prerelease
+          preid: dev
+          tag-prefix:  'v'
```
tukusejssirs: Do we want to use a prefix? I think it is redundant.

.github/workflows/dev-publishing.yml
```diff
@@ -0,0 +1,57 @@
...
+          commit-message: 'ci: bumps version to {{version}}'
```
tukusejssirs: What about:

```suggestion
          commit-message: 'ci: bump version to {{version}}'
```

.github/workflows/dev-publishing.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Git checkout dev branch
+        uses: actions/checkout@v2
+        with:
+          ref: dev
+
+      - name: Install node.js and npm
+        uses: volta-cli/action@v1
+
+      - name: Install npm depencancies
+        run: npm ci --no-audit
```
tukusejssirs: Shouldn’t we install the deps _after_ bumping the version? Although I don’t know how `phips28/gh-action-bump-version@master` works, from my experience, updating the version number in `package.json`/`package-lock.json` after installing the deps, commiting them, removing `node_packages/` and running `npm i` modifies `package-lock.json`.

And I think I am right:

https://github.com/romcal/romcal/blob/139290e71226075a05705b17396b368b6adf7f79/package.json#L3

https://github.com/romcal/romcal/blob/139290e71226075a05705b17396b368b6adf7f79/package-lock.json#L3

Also note that before version bumping I remove `package-lock.json`, but it shoudn’t be necessary.

scripts/publish.ts
```diff
@@ -0,0 +1,55 @@
...
+  } else {
+    console.log(data, '\n');
+  }
+};
+
+(async () => {
+  // Start by publishing the main romcal library
+  console.log(` - Publishing romcal`);
+  const data = await npmPublish({
+    package: path.join(__dirname, '..', 'package.json'),
```
tukusejssirs:
```suggestion
    package: path.join(__dirname, '../package.json'),
```

scripts/publish.ts
```diff
@@ -0,0 +1,55 @@
...
+    access: 'public',
+    token,
+    tag,
+    dryRun,
+  });
+  afterPublish(data);
+
+  // Retrieve the list of all calendar bundles
+  const bundlesBasePath = path.join(__dirname, '..', 'dist', 'bundles');
```
tukusejssirs:
```suggestion
  const bundlesBasePath = path.join(__dirname, '../dist/bundles');
```

.github/workflows/dev-publishing.yml
emagnier: It's just to stay consistent with previous git tags, but it's not mandatory.

.github/workflows/dev-publishing.yml
emagnier: It doesn't matter, since `package-lock.json` isn't embedded in the NPM package. However you are right, it would be more logical to bump the version before running an install. I'll change the step orders.

scripts/publish.ts
emagnier: It's more performant to keep them separated, but ok for the readability.

scripts/publish.ts
emagnier: It's more performant to keep them separated, but ok for the readability.

.github/workflows/dev-publishing.yml
emagnier: In fact, I prefer to keep it. Because a git tag could possibly contain other thing than a version.

.github/workflows/dev-publishing.yml
tukusejssirs: IMO it does matter, but not from the NPM package point of view.

1. We use `npm ci` which [by default uses `package-lock.json` if it exists](https://docs.npmjs.com/cli/v8/commands/npm-ci).
2. As I tried to demonstrate above, we need to _recreate_ `package-json.lock` after version bump and before commiting it. The easiest way is to run `npm i` (or `npm ci` but with `package-lock.json` removed); I tried to manually bump the version in `package-lock.json`, but `package-lock.json` simply gets modified whenever I install the deps in a clean (just cloned) repo. It does not happen when I recreate `package-lock.json` using `npm i`.

scripts/publish.ts
tukusejssirs: I didn’t know it is more performant when we merge three (more) strings instead of two (less) strings. Why is it so?

If it is more performant, I think we should them separated.

.github/workflows/dev-publishing.yml
tukusejssirs: And it might be reasonable to push the commit _after_ successful build and NPM release. What it the build/release for some reason fails?

emagnier: You are right, even if the `package-lock.json` isn't part of the NPM package, it must be updated in the codebase.
However, `phips28/gh-action-bump-version@master` is already taking care of updating it. So we are good 👍 .

And you also pointed out a good point in your last comment: the automated bump should be run just after the test step (to not create a new commit if the build or the tests are failing). I didn't see available options to bump the version without committing, but it seems reasonable to make the bump step just before the publishing step.
The publishing step can fail if the exact same npm version is already published, or if the NPM token isn't correctly set (so the risks are very limited from my point of view).

tukusejssirs: > However, `phips28/gh-action-bump-version@master` is already taking care of updating it. So we are good :+1:.

I didn’t say it does not care of it (I have never ever used that package nor did I test it), but you said you tested this PR prior to sharing it with us and I have found out that the versions in `package.json` and `package-lock.json` differ, therefore I suspect it does not work as expected, but I might be wrong. :man_shrugging:

> the automated bump should be run just after the test step

I think that bumping should be done prior to installing dependences, but _only in `package*.json` files (and potentionally other files), but no commit should be created _until romcal is tested and packages released_.

As I said, I don’t use _any_ package/program to bump versions in my repos. I bump the version manually in `package.json` files, run `npm i`, test, release and if everything was successful, create a new commit. But then, the published packages might not point to the right commit, but I might be wrong.

>  The publishing step can fail if the exact same npm version is already published, or if the NPM token isn't correctly set (so the risks are very limited from my point of view).

I didn’t say the risks are high, but I personally like to be on the safer side if possible. But then, I leave the final decision up to you. :wink:

tukusejssirs: Tell me if you think this PR is ready to be merged. I don’t see other issues with this PR.

tukusejssirs:
> I'm taking this opportunity to review the branch and tag names.

As for the branch/tag names, go on. I don’t really care for the exact naming, but still prefer to use same names for branches and tags, which you already suggest, so do as you wish. :wink:

.github/workflows/dev-publishing.yml
emagnier: > therefore I suspect it does not work as expected

Ok, I understand now, thanks for your patience 😄. In fact, in this PR I've manually updated the version in the `package.json` file, but not in `package-lock.json`. It's now fixed.

Also, I've added a [missing feature](https://github.com/phips28/gh-action-bump-version/pull/139) in the `phips28/gh-action-bump-version` to prevent doing a commit just after the bump. Commit and tagging is now done just before publishing. So if the build or the tests are failing, no commit/tag is pushed to the repo, and the CI workflow is aborted.

If you don't see anything else, I think this PR is ready to be merged.

tukusejssirs (approved):

emagnier:
FYI this CI workflow isn't working yet, since the `dev` branch is protected. 

I need to figure out how to configure it correctly.

PR #335, opened 2021-12-07
Update GA workflow to be able to push commits on dev

emagnier:
Related to: https://github.com/romcal/romcal/pull/332.

Because `dev` is a protected branch.... this PR:
- updates the GitHub credentials (to be able to push commits on `dev`), additional settings has been made in this repo settings.
- displays the new updated version in the console (after the version is bumped), just to provide more feedback when looking at the console.
- Git commit/tag do not use a third-party module anymore. `git` commands are run directly.
- Prevent this CI workflow to run in a loop (since this workflow is automatically running on a new commit(s) event, do not run again this workflow if the latest commit message starts with "ci:").

### DIFF
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -15,14 +15,15 @@ jobs:
 
     env:
       CI: true
-      GITHUB_USER: romcalproject
-      GITHUB_EMAIL: romcalproject@gmail.com
+      GITHUB_USER: github-actions
+      GITHUB_EMAIL: github-actions@github.com
 
     steps:
       - name: Git checkout dev branch
         uses: actions/checkout@v2
         with:
           ref: dev
+          token: ${{ secrets.REPO_TOKEN }}
 
       - name: Install node.js and npm
         uses: volta-cli/action@v1
@@ -39,6 +40,13 @@ jobs:
           skip-commit: 'true'
           skip-tag:  'true'
 
+      - name: Read package.json
+        uses: culshaw/read-package-node-version-actions@v1
+        id: package
+
+      - name: Display bumped version
+        run: 'echo "New romcal version: ${{ steps.package.outputs.version }}"'
+
       - name: Install npm dependencies
         run: npm ci --no-audit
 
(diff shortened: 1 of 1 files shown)


tukusejssirs (approved):

emagnier:
🎉  The workflow script is working well, new version of romcal and every calendar bundles are correctly built and tested, with a version bumped (with commit + tag), and published! Managing manually the publishing of every calendar bundles was taking a lot of time, so I'm happy to see it working 😄 

There are still one thing that could be improved. There are actually 2 workflows: "Romcal CI" and "Dev branch publishing".
"Romcal CI" is actually running 3 times because of different `push` events on the `dev` branch. I think it's even no necessary to run that on the `dev` branch, because all the builds / tests are also executed from the "Dev branch publishing" script. I'll try to improve that in the next days.

.github/workflows/dev-publishing.yml
```diff
@@ -51,17 +59,15 @@ jobs:
...
-          message: 'ci: bump version to ${{ steps.package.outputs.version }}'
-          tag: 'v${{ steps.package.outputs.version }}'
-          author_name: ${{ env.GITHUB_USER }}
-          author_email: ${{ env.GITHUB_EMAIL }}
+        run: |
+          git config user.name ${{ env.GITHUB_USER }}
+          git config user.email ${{ env.GITHUB_EMAIL }}
+          git add .
+          git commit -m "ci: bump version to ${{ steps.package.outputs.version }}"
+          git push
```
tukusejssirs, line 68: Now I see this line as redundant, as `git push --tags` on L70 does pushing commits and tags alike.

tukusejssirs:
> There are still one thing that could be improved. There are actually 2 workflows: "Romcal CI" and "Dev branch publishing". "Romcal CI" is actually running 3 times because of different `push` events on the `dev` branch. I think it's even no necessary to run that on the `dev` branch, because all the builds / tests are also executed from the "Dev branch publishing" script. I'll try to improve that in the next days.

I actually don’t know what you’re talking about, but go ahead. :wink:

---

I think me might want to publish the packages on [GitHub Releases](https://github.com/romcal/romcal/releases) too.

.github/workflows/dev-publishing.yml
emagnier: `git push` do not push the tags, this is why we need then to do a `git push --tag` specifically for the tags.

tukusejssirs: Aargh, right. I use `git push --follow-tags`, but I use annotated tags only (it does not push lightweight tags).

But according [this SO answer](https://stackoverflow.com/a/3745250/3408342), there are several options. Lastest update suggests to use the following command, which should push lightweight and annotated tags alike (`git` 2.4.x is needed):

```bash
git push --atomic origin dev v${{ steps.package.outputs.version }}
```

emagnier: Thank you! I'll try to have a closer look on that in the next days!

PR #337, opened 2021-12-07
Refactor TypeScript types & export more romcal types

emagnier:
This PR is just a small refactor of some TS types. 

It also exports more types, that can be used on TS projects using romcal.


tukusejssirs (approved):

tukusejssirs:
> It also exports more types, than can be used on TS projects using romcal.

Nice, @emagnier, but I believe there’s a typo in the PR description: `than` → `that`. You might want to fix it if you include it in the commit message. :wink:

emagnier:
Thank you @tukusejssirs for your sharp eye!

PR #348, opened 2021-12-31
Localize the cycle metadata

### LINKED ISSUE #233: Localise cycles
We should localise (celebration) cycle names, too.

That is, localise these strings:
- `Sunday cycle [ABC]`;
- `ferial cycle (I|II)`.

It might be also a good idea to localise `year`.

emagnier:
**Closes #233.**  Related to #346.

- New cycles properties:
  - `properCycleName`
  - `sundayCycleName`
  - `weekdayCycleName`
  - `psalterWeekName`
- New i18n entries (in `en`, `fr`, `it`) to translate these new properties. Then, PRs are welcome to translate all the other existing locales.
- Refactor of the `CycleMetadata` object.

### DIFF
--- a/lib/locales/fr.ts
+++ b/lib/locales/fr.ts
@@ -63,6 +63,20 @@ export const locale: Locale = {
     weekday: 'Férie',
   },
 
+  cycles: {
+    proper_of_time: 'Propre du Temps',
+    proper_of_saints: 'Propre des Saints',
+    sunday_year_a: 'Année A',
+    sunday_year_b: 'Année B',
+    sunday_year_c: 'Année C',
+    weekday_year_1: 'Année impaire',
+    weekday_year_2: 'Année paire',
+    psalter_week_1: 'Semaine I des Psaumes',
+    psalter_week_2: 'Semaine II des Psaumes',
+    psalter_week_3: 'Semaine III des Psaumes',
+    psalter_week_4: 'Semaine IV des Psaumes',
+  },
+
   weekdays: {
     0: 'dimanche',
     1: 'lundi',
--- a/lib/locales/it.ts
+++ b/lib/locales/it.ts
@@ -59,6 +59,20 @@ export const locale: Locale = {
     weekday: 'Feria',
   },
 
+  cycles: {
+    proper_of_time: 'Proprio del Tempo',
+    proper_of_saints: 'Proprio dei Santi',
+    sunday_year_a: 'Anno A',
+    sunday_year_b: 'Anno B',
+    sunday_year_c: 'Anno C',
+    weekday_year_1: 'Anno dispari',
+    weekday_year_2: 'Anno pari',
+    psalter_week_1: 'I Settimana del Salterio',
+    psalter_week_2: 'II Settimana del Salterio',
+    psalter_week_3: 'III Settimana del Salterio',
+    psalter_week_4: 'IV Settimana del Salterio',
+  },
+
   weekdays: {
     0: 'domenica',
     1: 'lunedì',
--- a/lib/models/cycles-metadata.ts
+++ b/lib/models/cycles-metadata.ts
@@ -40,8 +40,8 @@ export class CyclesMetadata implements BaseCyclesMetadata {
       let weekdayCycle: WeekdayCycle;
 
       // Formula to calculate Sunday cycle (Year A, B, C)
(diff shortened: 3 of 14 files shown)


lib/locales/fr.ts
```diff
@@ -63,6 +63,20 @@ export const locale: Locale = {
...
+    proper_of_saints: 'Propre des Saints',
+    sunday_year_a: 'Année A',
+    sunday_year_b: 'Année B',
+    sunday_year_c: 'Année C',
+    weekday_year_1: 'Année impaire',
+    weekday_year_2: 'Année paire',
+    psalter_week_1: 'Semaine I des Psaumes',
+    psalter_week_2: 'Semaine II des Psaumes',
+    psalter_week_3: 'Semaine III des Psaumes',
+    psalter_week_4: 'Semaine IV des Psaumes',
```
tukusejssirs, line 77: Question: do you always write these terms with capital letters in French? If yes, no change required, but if not, I suggest to use lowercase by default.

lib/locales/it.ts
```diff
@@ -59,6 +59,20 @@ export const locale: Locale = {
...
+    proper_of_saints: 'Proprio dei Santi',
+    sunday_year_a: 'Anno A',
+    sunday_year_b: 'Anno B',
+    sunday_year_c: 'Anno C',
+    weekday_year_1: 'Anno dispari',
+    weekday_year_2: 'Anno pari',
+    psalter_week_1: 'I Settimana del Salterio',
+    psalter_week_2: 'II Settimana del Salterio',
+    psalter_week_3: 'III Settimana del Salterio',
+    psalter_week_4: 'IV Settimana del Salterio',
```
tukusejssirs, line 73: Same as with French: do they _always_ use capital letters in Italian?

lib/models/cycles-metadata.ts
```diff
@@ -0,0 +1,111 @@
...
+    // Compute cycle of the liturgical year,
+    // and cache the data since they are the same for every day of the year
+    if (!this.#config.cyclesCache[year]) {
+      const firstSundayOfAdvent = getUtcDateFromString(calendar.startOfLiturgicalYear);
+
+      let sundayCycle: SundayCycle;
+      let weekdayCycle: WeekdayCycle;
+
+      // Formula to calculate Sunday cycle (Year A, B, C)
+      const thisSundayCycleIndex: number = (year - 1963) % 3;
```
tukusejssirs: This make romcal (sort of) backward incompatible with older romcal versions and makes romcal generate calendars for years after `1963` (inclusively). That is not a big deal, but if one tried to generate calendar, e.g., for `1900`, I’d presume it _should_ fail calendar generation, but actually it does not fail and Easter Sunday is on 15 April 1900 (based on [this website](https://tlarsen2.tripod.com/thomaslarsen/easterdates.html), it is correct).

Still, I suggest to remove `1963` constant, as it is not required to calculate the Sunday cycle letter: simply use `year % 3` instead compare the remainder:
- `1` → `A`;
- `2` → `B`;
- `0` → `C`.

tukusejssirs (dismissed):
I have some comments on the changes, but nothing serious, therefore it is mergable as is.

lib/models/cycles-metadata.ts
emagnier: This line of code is coming from older romcal version (at least 1.x), and could be Indeed improved.

lib/locales/fr.ts
emagnier: Except for the Psalter Week where I can't currently confirm, other titles are written like that in the missal.

lib/locales/it.ts
emagnier: Same in Italian, from the Italian missal.

lib/models/cycles-metadata.ts
tukusejssirs: I think it should be changed (maybe in a separate PR), as I think that the historicality of the calendar should be set separately (e.g. using a romcal option, see #23).

tukusejssirs:
I think you can merge this PR as is. :wink:

lib/models/cycles-metadata.ts
emagnier: Fixed.

tukusejssirs (approved):

PR #350, opened 2022-01-02
fix: some l10n issues (fix #295)

tukusejssirs:
This PR fixes some issues in #295.

Bellow is a list of all changes I made in this PR.

---

#### List of changed key names

The following key names were changed in comparison with `dev` branch.

- `adelaide_of_burgundy` → `adelaide_of_burgundy_empress`;
- `adelphus_bishop` → `adelphus_of_metz_bishop`;
- `all_souls` → `commemoration_of_all_the_faithful_departed`;
- `all_the_saints_of_the_diocese_of_paris` → `all_saints_of_the_archdiocese_of_paris`;
- `all_the_saints_of_the_diocese_of_saint_denis` → `all_saints_of_the_diocese_of_saint_denis`;
- `alphonsus_liguori_bishop` → `alphonsus_mary_liguori_bishop`;
- `amandus_first_bishop_of_strasbourg` → `amandus_of_strasbourg_bishop`;
- `amarin_abbot` → `amarin_of_alsace_abbot`;
- `annunciation` → `annunciation_of_the_lord`;
- `arbogast_of_strassburg_bishop_patron_of_the_diocese_of_strasbourg` → `arbogast_of_strasbourg_bishop_patron_of_the_archdiocese_of_strasbourg`;
- `arbogast_of_strassburg_bishop` → `arbogast_of_strasbourg_bishop`;
- `ascension` → `ascension_of_the_lord`;
- `assumption` → `assumption_of_the_blessed_virgin_mary`;
- `aurelia_virgin` → `aurelia_of_strasbourg_virgin`;
- `catherine_laboure_virgin` → `catherine_zoe_laboure_virgin`;
- `ceran_of_paris_bishop` → `ceraunus_of_paris_bishop`;
- `charles_de_foucauld` → `charles_of_jesus_de_foucauld`;
- `christ_the_king_sunday` → `our_lord_jesus_christ_king_of_the_universe`;
- `christmas` → `nativity_of_the_lord`;
- `corpus_christi` → `most_holy_body_and_blood_of_christ`;
- `cyril_the_philosopher_monk_and_methodius_of_thessaloniki_bishop_copatrons_of_europe` → `cyril_constantine_the_philosopher_monk_and_methodius_michael_of_thessaloniki_bishop_copatrons_of_europe`;
[...]

PR #367, opened 2022-08-10
data: update the canonization level of charles_de_foucauld

emagnier:
_Charles de Foucauld_ is now a saint, since May 15, 2022.

### DIFF
--- a/lib/catalog/martyrology.ts
+++ b/lib/catalog/martyrology.ts
@@ -972,7 +972,7 @@ export class Martyrology {
     },
 
     charles_de_foucauld: {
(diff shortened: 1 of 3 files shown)


tukusejssirs (approved):
Yeah, I actually tried to fix in #350, however, I forgot to change the level in [`lib/catalog/martyrology.ts`](https://github.com/romcal/romcal/pull/367/files#diff-5c928ad6b791e4013303fe86266ee6ce845d4d1e55c6f70c35cd003879f473c7).

Anyway, it seems to be alright, so I approve your changes. Thanks :pray:

PR #374, opened 2022-09-30
Fix some issues with holy days of obligations in the GRC

tukusejssirs:
- `holy_family_of_jesus_mary_and_joseph` is `isHolyDayOfObligation` when it occurs on Sunday;
- `baptism_of_the_lord` is `isHolyDayOfObligation` only if it occurs on Sunday;
- `most_sacred_heart_of_jesus` is not `isHolyDayOfObligation` in the GRC.

### DIFF
--- a/lib/general-calendar/proper-of-time.ts
+++ b/lib/general-calendar/proper-of-time.ts
@@ -173,7 +173,6 @@ export class ProperOfTime extends CalendarDef {
     this.#newLiturgicalDayDef(`holy_family_of_jesus_mary_and_joseph`, {
       precedence: Precedences.GeneralLordFeast_5,
       dateDef: { dateFn: 'holyFamily', yearOffset: -1 + yearOffset },
-      isHolyDayOfObligation: false,
       seasons: [Seasons.ChristmasTime],
       periods: [Periods.ChristmasOctave, Periods.ChristmasToPresentationOfTheLord],
       calendarMetadata: {},
@@ -260,7 +259,6 @@ export class ProperOfTime extends CalendarDef {
     this.#newLiturgicalDayDef(`baptism_of_the_lord`, {
       precedence: Precedences.ProperOfTimeSolemnity_2,
       dateDef: { dateFn: 'baptismOfTheLord', yearOffset: yearOffset },
-      isHolyDayOfObligation: true,
       seasons: [Seasons.ChristmasTime],
       periods: [Periods.DaysFromEpiphany, Periods.ChristmasToPresentationOfTheLord],
       calendarMetadata: {},
@@ -504,7 +502,7 @@ export class ProperOfTime extends CalendarDef {
     this.#newLiturgicalDayDef(`most_sacred_heart_of_jesus`, {
       precedence: Precedences.GeneralSolemnity_3,
       dateDef: { dateFn: 'mostSacredHeartOfJesus', yearOffset: yearOffset },
-      isHolyDayOfObligation: true,
+      isHolyDayOfObligation: false,
       seasons: [Seasons.OrdinaryTime],
       periods: [],
       calendarMetadata: { dayOfWeek: 5 },


tukusejssirs:
@emagnier, review this PR please. I think it should be alright, however I am not sure if `isHolyDayOfObligation: false` should be removed from `most_sacred_heart_of_jesus` (it should be the default value).

@emagnier, we should double-check if HDOs in particular calendars are still valid. If not, we should update them. I think you should check HDOs in `france*.ts` prior to merging this PR.

Note however that this PR should be merged _after_ #350.

tsconfig.json
```diff
@@ -30,5 +30,6 @@
     "scripts/**/*",
     "tmp/bundles/**/*",
     "tmp/constants/**/*"
-  ]
+  ],
+  "files": ["types/global.d.ts"]
```
emagnier: Why are you adding this setup here?

lib/catalog/martyrology.ts
```diff
@@ -3,4793 +3,4088 @@ import { MartyrologyCatalog } from '../types/martyrology';
...
+      titles: [Titles.Martyr]
     },
-
     adalbert_of_prague_bishop: {
       canonizationLevel: CanonizationLevels.Saint,
       name: 'Adalbert',
       titles: [Titles.Bishop, Titles.Martyr],
-      dateOfDeath: 997,
+      dateOfDeath: 997
     },
```
emagnier: Is there a rule that have been changed (or still not set), so all commas on last object property are removed?

I'm personally more in favor to keep them (so if we add a new property in the future, the previous line ins't modified in git blame history). But we can discuss.

tsconfig.json
tukusejssirs: As I said [here](https://github.com/romcal/romcal/pull/374#issuecomment-1263296479), this PR should be merged _after_ #350.

This particular change is described in the [#350 description](https://github.com/romcal/romcal/pull/350#issue-1092095725) (at the very end):

> I had to fix the tests too by using [this workaround](https://github.com/jest-community/jest-extended/issues/367#issuecomment-1127269184) (otherwise `jest-extended` would fail to load in tests). Also, I needed to update `jest` (`^28.1.2`) and `jest-extended` (`^2.0.0`). No other change was required to make the tests pass.

lib/catalog/martyrology.ts
tukusejssirs: While I prefer removing final commas in objects/arrays, I didn’t intend to change the style. I simply forgot to add the comma. Also I have configured LSP TypeScript server used in my text editor (Sublime Text) to remove them.

~~I’ll fix it in the commit (I’ll amend the current latest commit).~~ I’ll fix this in this PR only if you want to make the changes in this PR (for explanation continue reading).

On the other hand, I have noticed that you have removed `prettier` NPM commands (from `package.json`; I actually fighted with `prettier` in my repositories, thus I simply removed `prettier` and use `eslint` only). Also running `npm run lint` (`eslint . --ext .ts`) didn’t fixed this issue.

> But we can discuss.

We might do that, but not in this PR. :wink: I don’t want to impose you my preferences, however, what I might want to improve is the `eslint` rules (after some kind of consensus).

Keeping the conversion on topic, I suggest to add [`comma-dangle`](https://eslint.org/docs/latest/rules/comma-dangle) rule to `eslint` config to `["error", "always-multiline"]`. A bit off topic, we should also add `"lint:fix": "eslint . --ext .ts --fix"` to `package.json` scripts (I usually have `lint` set to fix the issues and `lint:check` to check the issues, however, the wording does not matter).

---

@emagnier, do you want to make these changes in this PR? A follow-up PR might be better.

emagnier: > do you want to make these changes in this PR? A follow-up PR might be better.

Not here, it's indeed off-topic here. 
But we should fix this somewhere else, and discuss on how to improve prettier / eslint to make our life better.

emagnier (approved):

lib/general-calendar/proper-of-time.ts
```diff
@@ -170,15 +170,15 @@ export class ProperOfTime extends CalendarDef {
 
     // The Sunday within the Octave of the Nativity of the Lord,
     // Feast of the Holy Family of Jesus, Mary and Joseph.
-    this.#newLiturgicalDayDef(`holy_family`, {
+    this.#newLiturgicalDayDef(`holy_family_of_jesus_mary_and_joseph`, {
       precedence: Precedences.GeneralLordFeast_5,
       dateDef: { dateFn: 'holyFamily', yearOffset: -1 + yearOffset },
-      isHolyDayOfObligation: false,
+      isHolyDayOfObligation: dow === 0,
```
tukusejssirs: @emagnier, I have just noticed that this does not work as I expected, as `dow` is not defined here at all.

How could I get the `holy_family` date? I think putting `const dateDef = holyFamily()` and `const dow = dateDef.getDate()` before L173? Both `holyFamily()` and `getWeekNumber()` would be imported from `lib/utils/dates.ts`.

Note that same problem is with `baptismOfTheLord`.

**Update:** It seems it won’t work as I supposed, as (1) `holyFamily()` is not `public` method (though fixing that would be easy); (2) `Dates` class needs to be initialised with `config` (`RomcalConfig`) and `year`, which might not be quite a good solution, as it is already initialised elsewhere.

emagnier: Indeed, I missed that when reviewing the PR.

In fact, you don't have to take care here of the `dow === 0`, because `isHolyDayOfObligation` is automatically changed to `true` if this liturgical day falls in a Sunday.

emagnier (changes requested):
FYI I've added an extra and automated check to make PRs mergeable (you will see it if you update/rebase the currently opened PRs). This will ensure that everything merged into `dev` do not break tests and the run.

lib/catalog/martyrology.ts
tukusejssirs: Awesome! I’ll remove those lines in the next commit then. :pray:

tukusejssirs:
> FYI I've added an extra and automated check to make PRs mergeable (you will see it if you update/rebase the currently opened PRs). This will ensure that everything merged into `dev` do not break tests and the run.

Perfect! Thanks! :pray:

tukusejssirs:
@emagnier, the issue should be fixed. romcal is buildable locally. It could be merged. :wink:

tsconfig.json
tukusejssirs: @emagnier, you haven’t reacted on this yet. Are you okay with this or not?

If you okay this this or still require some changes? If you require some changes, those should be made in #350.

emagnier (approved):

tukusejssirs:
@emagnier, should I merge this PR or should I let you merge it? I ask this because of the tagging problem (see [this comment](https://github.com/romcal/romcal/pull/350#issuecomment-1277698928)).

tsconfig.json
emagnier: I'm not sure why this is required, has I don't need this line to make tests running.

emagnier: But ok to keep it, because it has no impact I think.

tukusejssirs: Okay, you’re right, it is not required. I have no idea why it didn’t work for me without it then.

If you want, we can remove it.

emagnier:
Yes you can merge. We will see what's going on the tagging issue. I don't have much time to look at it any further right now.

tukusejssirs:
> Yes you can merge. We will see what's going on the tagging issue. I don't have much time to look at it any further right now.

I see. Should update the `version` in `package.json` before merging the PRs?

tsconfig.json
emagnier: If you don't need it to make tests running: yes, remove it (otherwise keep it for now, meanwhile founding a best raison/fix).

emagnier:
> Should update the version in package.json before merging the PRs?

No. This should stay automated. Let's merge this PR, and we will see if it needs more investigation.

emagnier:
This is working now. Before this merge, I have reintroduced a property in GitHub action config https://github.com/romcal/romcal/commit/0399e3091c2d41ee78a27d923086580bcb665088, I just removed previously. And it seems that was the problem.

Issue #383, opened 2022-10-05
Update `eslint` rules


tukusejssirs:
I think we should update the ESLint rules. Rules rules are quite minimalistic and does not cover even the basic style issues (like trailing commas).

Some prefer using some style guides, I prefer using custom ESLint config and add rules that make sense to me, my code style and my coding/programming style. However, I am open to change any rule in the proposed ESLint config (see below). I don’t want to enforce any rule by myself, however, I want to enforce consistency.

Note that I compiled this config by myself and I am more of a TS newbie, the suggested config is not perfect and could be improved.

I commented on some stuff that you might want to consider, or when my preference differs from style currently used by romcal.

#### Some additional stuff to consider

- [ ] consider specifying `parserOptions.ecmaVersion` (should it be `es2015`?);
- [ ] consider using tabs for indentation (less characters to use, thus smaller source code size):
   - this would require changes in `.editorconfig` and `.prettierrc`;
- [ ] consider removing semicolons on EOLs;
- [ ] I prefer sorting JSON file properties, as well as TS/JS object properties (including TS type properties), thus could we sort `.eslintrc.json` too?
- [ ] let’s keep the trailing commas:
   - I prefer removing them, however, I understand that the diffs are more tidy when trailing commas are used;
- [ ] consider installing the following ESLint plugins (for rules and usage, see the proposed ESLint config below:

   ```bash
   npm -D i \
     # Note: This provides `plugin:import/recommended`.
     eslint-plugin-import \
     # Note: This provides `sort-keys-custom-order`.
     eslint-plugin-sort-keys-custom-order \
     # Note: This provides `eslint-plugin-tsdoc`.
     eslint-plugin-tsdoc \
     # Note: This provides `unicorn`.
     eslint-plugin-unicorn
   ```

#### Suggested configuration

<details>
   <summary><code>.eslintrc.json</code></summary>

```js
{
  "env": {
    "browser": false,
    "es6": true,
    "node": true
  },
  "extends": [
    "eslint:recommended",
    // Note: `eslint-plugin-import` plugin intends to support linting of ES2015+ (ES6+) import/export syntax, and prevent issues with misspelling of file paths and import names. All the goodness that the ES2015+ static module syntax intends to provide, marked up in your editor.
    "plugin:import/recommended",
    "prettier"
  ],
  "overrides": [
    {
      "files": [
        "*.js",
        "*.jsx"
      ],
      "rules": {
[...]

PR #391, opened 2022-10-13
chore(style): add NPM script for styling with `prettier` and run it on the codebase

tukusejssirs:


tukusejssirs:
@emagnier, we could run `prettier` in `npm run build` too, as well as as part of `.github/workflows/dev-publishing.yml`.

emagnier:
Not sure... I think `npm run build` should only focus on building, no code formatting.

However, GitHub Action should throw an error if the prettier checks return something that need to be prettierized..

package.json
```diff
@@ -12,6 +12,8 @@
     "clean": "rimraf coverage dist tmp",
     "lint": "eslint . --ext .ts",
     "prebuild": "npm run lint",
+    "style": "prettier -w --config .prettierrc lib",
+    "style:check": "prettier -c --config .prettierrc lib",
```
emagnier: I'm not very convinced by the `style` word here. "Style" is too vague and can express anything in my point of view. Maybe just use the word `prettier`?

tukusejssirs:
> Not sure... I think `npm run build` should only focus on building, no code formatting.

Well, I agree with you, however `prebuild` script in `package.json` is set to `npm run lint`, which is why I thought we _could_ add it in there.

However, I too think that the `build` script should only try to build, no style checking, thus no `eslint` or `prettier` should be run in there. I suggest to remove `prebuild` script altogether.

I think we should consider creating a precommit hook to run both `eslint` and `prettier`.

> However, GitHub Action should throw an error if the prettier checks return something that need to be prettierized..

We definitely should check style using both `eslint` and `prettier` before merging anything into romcal.

package.json
tukusejssirs: Yeah, we can change it to `prettier`, as we use to have in between versions `v1.3.1-alpha.6` and `v2.0.0-alpha.0`.

emagnier:
Prettier should not be run when running a build (or from the prebuild script), because it will potentially modify files. When running a build, we are not supposed to modify the codebase. In addition, if the Prettier checks are failing, it should not break the build (as the contributor is probably working on something, but didn't run Prettier yet).

However, it's a good idea to:
- run Prettier in pre-commit hooks
- run Prettier:check in CI before merging anything

---

Eslint is a different topic (it's not only code styling/formating). This linter will mainly prevent problems in the codebase, analyze and detect code smell, and ensure keeping quality and codebase consistency. So having it before running the build is valuable IMO.

However, I'm not for adding eslint in pre-commit hooks. Because sometimes it's useful to create intermediate commits (while eslint issues are not all fixed yet) before submitting a PR (for example, when developing new features or doing code refactor).

package.json
```diff
@@ -12,6 +12,8 @@
     "clean": "rimraf coverage dist tmp",
     "lint": "eslint . --ext .ts",
     "prebuild": "npm run lint",
+    "prettier": "prettier -w --config .prettierrc lib",
+    "prettier:check": "prettier -c --config .prettierrc lib",
```
emagnier: What about also prettify the `docs`, `scripts` and `tests` directories?

tukusejssirs:
> Prettier should not be run when running a build (or from the prebuild script), because it will potentially modify files. When running a build, we are not supposed to modify the codebase.

Well, I actually meant running `prettier -c` and `eslint` (without `--fix`), however, that should be placed in a pre-commit (or pre-merge into `dev` branch) hook rather than as a pre-build step.

> In addition, if the Prettier checks are failing, it should not break the build (as the contributor is probably working on something, but didn't run Prettier yet).

Agreed. I think if `prettier` checks fail, it should not get merged into `dev` in the first place.

> However, it's a good idea to:
> 
>     * run Prettier in pre-commit hooks
> 
>     * run Prettier:check in CI before merging anything

I actually don’t think that fixing `prettier`/`eslint` issues should be in pre-commit/pre-merge hooks. I think it should be the committers duty. However, I am not totally against it, thus if you think we should do this automatically, I am okay with it.

> Eslint is a different topic (it's not only coding style/format).
>
> This linter will mainly prevent problems in the codebase, analyze and detect code smell, and ensure keeping quality and codebase consistency. So having it before running the build is valuable IMO.

In our current `eslint` config, it may be, however, it might check some _possible_ mistakes and errors (see my proposed `eslint` config in #383).

Still, even if there are styling issues, we should consider them blockers anyway. IMHO code style should be consistant. Contributors can style their code as they please and `eslint` should fix them all (or at least warn about them).

You seem to look at `prettier` and `eslint` as two different things, however, they have much in common. Prettier describes itself as an _opinionated code **formatter**_, while ESLint _statically analyzes code to quickly find problems_. Both fix/check styling issues, while (as I understand it) Prettier only deals with the format (style).

I personally ditched Prettier in my projects, as I kept fighting with it. I personally prefer using ESLint only. There is not one thing that Prettier does that we can’t do in ESLint (although some of the rules work differently).

> However, I'm not for adding eslint in pre-commit hooks. Because sometimes it's useful to create intermediate commits (while eslint issues are not all fixed yet) before submitting a PR (for example, when developing new features or doing code refactor).

Okay, then I suggest to create a pre-merge check (via GitHub Actions) for check the code both by `prettier` and `eslint`, no need to create a pre-commit hook at all.

Note that this is not the way I code. I try to fix all styling/formatting issues prior to creating a commit (yes, I forget to do so sometimes). However, the style/format of individual commits is not that important, only what actually merges into `dev`.

package.json
tukusejssirs: I have no problem with that. Do you suggest to simply update the command or update `.prettierrc`?

emagnier: Just adding that is enough, and your PR will be ready 👍 

```suggestion
    "prettier": "prettier -w --config .prettierrc {lib,docs,tests}",
    "prettier:check": "prettier -c --config .prettierrc {lib,docs,tests}",
```

tukusejssirs: Final question: should plain `prettier` script auto-fix issues or only check for issues? Accordingly, we should use `prettier:fix` or `prettier:check` for the other.

I have noticed that `lint` is just a check. We might want to be consistant in here.

emagnier: +1 for consistency with `lint`.

Let's go for `prettier` + `prettier:fix`. 
Thanks 🙏

tukusejssirs:
```suggestion
    "prettier": "prettier -c --config .prettierrc {lib,docs,tests}",
    "prettier:fix": "prettier -w --config .prettierrc {lib,docs,tests}",
```

tukusejssirs:
@emagnier, I have resolved the issue you raised and formatted all the files using `prettier`. If it looks good to you, you could re-approve the PR. :wink:

emagnier (approved):

PR #393, opened 2022-10-13
feat(calendar): add Slovak diocesan calendars

tukusejssirs:
TODO

### DIFF
--- a/lib/catalog/martyrology.ts
+++ b/lib/catalog/martyrology.ts
@@ -1001,6 +1001,15 @@ export class Martyrology {
       titles: [Titles.Apostle],
       count: 2
     },
+    dedication_of_the_cathedral_of_saint_elizabeth_of_hungary_kosice_slovakia: {
+      name: 'Dedication of the Cathedral of Saint Elizabeth of Hungary, Košice, Slovakia',
+    },
+    dedication_of_the_cathedral_of_saint_francis_xavier_banska_bystrica_slovakia: {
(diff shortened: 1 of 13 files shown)


tukusejssirs:
@emagnier, we need to define the structure and/or calendar filename format:
- should we add all calendars into the filename, e.g. `${continent}.${subcontinent}.${region}.${country}.${diocese}.${parish}.${city}.ts` (omit what is not needed)? This might be hard to parse, as sometimes we omit some parts of the name and sometimes the diocese/parish/city names are the same, thus hard to distinguish (unless we add a prefix like `c_americas` where `c` stands for `continent`); 
- you have already added some diocesan calendars for France, but now the question is: should we prefix all national calendars with the continent name (e.g. `france.ts` → `europe.france.ts`)?
- we might go even further: we should also use subcontinents, as some countries span across multiple continents (e.g. Russia) and thus their continental patronages varies:
  - therefore I suggest to treat the following as continents and some of them have subcontinents:
    - `afrika`;
    - `americas`;
       - `north-america`;
       - `south-america`;
    - `antarctica`;
    - `australia-and-oceania`;
       - `australia`;
       - `oceania`;
    - `eurasia`;
       - `asia`;
       - `europe`;
- another thing that occurred to me is that we might want to explicitly define in each calendar from which calendar to inherit the celebrations:
  - this might be useful for calendars for religious communities, as well as for countries spanning multiple continents.

PR #394, opened 2022-10-17
fix(general): set `epiphanyOnSunday` and `corpusChristiOnSunday` to ` false` (fix #217)

### LINKED ISSUE #217: Corpus Christi on Sunday and `general.ts`
I don’t think that in the GRC Corpus Christi is celebrated on Sunday, however, in `general.ts` we have `corpusChristiOnSunday` set to `true`.

tukusejssirs:


tukusejssirs:
@emagnier, I had to fix some tests that the PR caused to fail. They all pass now, please review them too. :pray:

emagnier (approved):

Issue #386, opened 2022-10-12
Calendar metadata: `weekOfSeason` not computed correctly for timezones that have greater offset from UTC


emagnier:
Related: #384

---

romcal fails to calculate correct `weekOfSeason` when the locale timezone has greater ofset from UTC. In romcal, we should always use UTC timezone (via `Date.prototype.getUTC*()` methods), regardless of the local timezone.

Note: Originally, adding a complete test coverage for all calendar metadata was part of this issue, however, that was moved to a separate issue (#398).

tukusejssirs:
@emagnier, what’s the bug? Is [this](https://github.com/romcal/romcal/discussions/384#discussioncomment-3863574) wrong?

As for the tests, they’re indeed needed.

tukusejssirs:
@emagnier, to further prove that this is _not_ an issue (apart from missing tests), I have created following (quick) test. I manually created:
- an array of all dates of Ordinary Time in 2022;
- an array of all expected `weekOfSeason` values that correspond to the array of dates.

Then I simply compare the expected values to the values computed by romcal. They all match.

```js
const Romcal = require('romcal')
const {Slovakia_En} = require('@romcal/calendar.slovakia')
const r = new Romcal({localizedCalendar: Slovakia_En})

const testAllOrdinaryDays = [
	'2022-01-10', '2022-01-11', '2022-01-12', '2022-01-13', '2022-01-14', '2022-01-15',
	'2022-01-16', '2022-01-17', '2022-01-18', '2022-01-19', '2022-01-20', '2022-01-21', '2022-01-22',
	'2022-01-23', '2022-01-24', '2022-01-25', '2022-01-26', '2022-01-27', '2022-01-28', '2022-01-29',
	'2022-01-30', '2022-01-31', '2022-02-01', '2022-02-02', '2022-02-03', '2022-02-04', '2022-02-05',
	'2022-02-06', '2022-02-07', '2022-02-08', '2022-02-09', '2022-02-10', '2022-02-11', '2022-02-12',
	'2022-02-13', '2022-02-14', '2022-02-15', '2022-02-16', '2022-02-17', '2022-02-18', '2022-02-19',
	'2022-02-20', '2022-02-21', '2022-02-22', '2022-02-23', '2022-02-24', '2022-02-25', '2022-02-26',
	'2022-02-27', '2022-02-28', '2022-03-01',
[...]

PR #397, opened 2022-10-18
fix(general): fix `weekOfSeason` calculation and related stuff (fix #384)

tukusejssirs:
Changes:

- always use `Date.prototype.getUTC*()` methods instead of `Date.prototype.get*()`:
   - in `lib/`, in `tests/`, in `docs/` and in `examples/`;
- fix and improve `getWeekNumber()` function in `lib/utils/dates.ts`;
- fix `If first day of the year 2012 is a Sunday, Mary Mother of God is on that Sunday and the Sunday proceeding will be Epiphany of the Lord` test in `tests/dates.test.ts` (L563):
   - week number is not the same as number of Sundays in a range, thus:
      - Epiphany Sunday should be in the first ISO week of the year;
      - Epiphany should be the 2nd Sunday day of the year:
         - because of this, I had to add `getNumberOfWeekDaysInRange()` function to `lib/utils/dates.ts`, however, I don’t really think we should test how many week days (like Sundays in here) can be found in a range;
         - I think we should remove this test and `getNumberOfWeekDaysInRange()` function.


tukusejssirs:
@emagnier, I have no idea why [`volta-cli/action@v1`](https://github.com/romcal/romcal/actions/runs/3272042276/jobs/5382590181#step:3:11) would fail. Is zloirock/core-js#648 (closed as `can't reproduce`) related?

Now this PR should fix #386 (without adding tests), thus it is somewhat important to merge ASAP. When you find some time for romcal, you could do this as the first thing.

In regards to #386, I am inclined to update the issue description to provide better description of the bug. Also I want to move the addition/creation of tests into separate issue, where we could first discuss the general structure of the tests (in order to let them be more consistent and more thorough/robust; see #398). WDYT?

Also we should consider adding some tests with timezone(s) different from local one (like `TZ='America/Bogota' npm run test`). Now it all passes (with and without defining a custom timezone), however, we should do it programmatically via tests.

lib/utils/dates.ts
```diff
@@ -46,40 +46,48 @@ export const isValidDate = (maybeDate: unknown): boolean => {
...
 
+/**
+ * Gets the number of week days in range, including the end date
+ *
+ * @param   {Date}    start      Start date
+ * @param   {Date}    end        End date
+ * @param   {number}  dayOfWeek  Day of week to count
+ * @return  Number of week days in the range, including the end date
+ */
+export const getNumberOfWeekDaysInRange = ({start, end, dayOfWeek = 0}: {start: Date, end: Date, dayOfWeek?: 0 | 1 | 2 | 3 | 4 | 5 | 6}): number => {
```
emagnier: Since this function is only used in tests, not in the main romcal library, could you move it at the top of `dates.test.ts`?

lib/utils/dates.ts
```diff
@@ -46,40 +46,48 @@ export const isValidDate = (maybeDate: unknown): boolean => {
...
-      // If the next year starts before the middle of the week, it is week #1 of that year
-      weekNum = nDay < 4 ? 1 : 53;
-    }
-  } else {
-    weekNum = Math.floor((dayNum + day - 1) / 7);
-  }
-  return weekNum;
+  const d = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+  d.setUTCDate(d.getUTCDate() + 4 - (d.getUTCDay() || 7));
+  return Math.ceil(((d.getTime() - new Date(Date.UTC(d.getUTCFullYear(), 0, 1)).getTime()) / 864e5 + 1) / 7)
```
emagnier, line 70: 👍

emagnier (commented):
Thank you for taking the time to fix this!

There is only one minor comment to address (move `getNumberOfWeekDaysInRange` outside the main romcal library). Other than that, everything LGTM 👍

emagnier:
FYI, I just re-run the failing job and the CI seems happy now (we got the green light).

> In regards to https://github.com/romcal/romcal/issues/386, I am inclined to update the issue description to provide better description of the bug. Also I want to move the addition/creation of tests into separate issue, where we could first discuss the general structure of the tests (in order to let them be more consistent and more thorough/robust; see https://github.com/romcal/romcal/issues/398). WDYT?

I agree 👍  (yes, update the issue according to your comment here)

In addition, we should use Jest snapshots (or something else), to compare data of full generated years (a set of different years with a set of different calendars).

> Also we should consider adding some tests with timezone(s) different from local one (like TZ='America/Bogota' npm run test). Now it all passes (with and without defining a custom timezone), however, we should do it programmatically via tests.

I agree 👍

lib/utils/dates.ts
tukusejssirs: You’re right, I move it into `dates.test.ts`.

Note, however, that this function is sourced from [SO](https://stackoverflow.com/a/37408412/3408342). I think we could add a source to the JSDoc (which should be changed to TSDoc) and probably merge all those variables, as they are used only once. Something like this:

```ts
/**
 * Gets the number of week days in range, including the end date
 * 
 * @remarks This function was sourced from [StakOverflow](https://stackoverflow.com/a/37408412/3408342).
 *
 * @param    start     - Start date
 * @param    end       - End date
 * @param    dayOfWeek - Day of week to count
 * @returns  Number of week days in the range, including the end date
 */
const getNumberOfWeekDaysInRange = ({start, end, dayOfWeek = 0}: {start: Date, end: Date, dayOfWeek?: 0 | 1 | 2 | 3 | 4 | 5 | 6}): number => {
  // Calculate the number of weeks (even partial) from the first target day to the end
  return Math.ceil(
    // Calculate the number of days from the first target day to the end
    Math.max(
      // Calculate the number of days between start and end, including the end date (thus `+ 1` at the end)
      Math.ceil((end.getTime() - start.getTime()) / 864e5) + 1
      // Calculate the nb of days before the next target day (e.g. next Sunday after start)
      - (7 + dayOfWeek - start.getUTCDay()) % 7,
      0
    ) / 7
  );
}
```

lib/utils/dates.ts
tukusejssirs: Well, I am not that good that algorithm. I sourced it from [SO](https://stackoverflow.com/a/6117889/3408342). :wink:

lib/utils/dates.ts
tukusejssirs: This should be fixed now.

emagnier (dismissed):
Thank you ❤️ 

Just a small comment, if you modify the PR from a review, it's not worth to squash all the commits together, because it makes the code review more difficult to follow, especially when a lot of code is modified (to see what was adjusted following a comment).
But at the end indeed, it's important to squash everything's with a clean message before merging, for a clean git history (if it's not useful to keep all the commits separated).

tukusejssirs:
> Just a small comment, if you modify the PR from a review, it's not worth to squash all the commits together, because it makes the code review more difficult to follow, especially when a lot of code is modified (to see what was adjusted following a comment). But at the end indeed, it's important to squash everything's with a clean message before merging, for a clean git history (if it's not useful to keep all the commits separated).

Yeah, you’re right. I squashed the changes automatically. :wink:

On the other hand, GitHub provides us a simple [Compare](https://github.com/romcal/romcal/compare/4744345039972cea49a56bc9859570f37a76e139..ab70263668d0d371b9d63b79d49e87961aa1a178) button right below the log about force-pushed commit (right above your latest comment). :wink:

emagnier:
I hadn't noticed this _Compare_ button. A feature that appeared in GitHub without me realizing it 😅. Thank you, it's also a handy tool. So about commits, keep doing as you want 👍

tukusejssirs:
@emagnier, I had to resolve merge conflicts; could you re-approve the PR please? :pray:

emagnier (approved):

tukusejssirs:
I’ve just noticed that I fixed #384 (which is a discussion) instead of #386, thus is is not automatically close. :man_facepalming:

PR #414, opened 2023-01-16
fix typos

ghost:
just fixing a few typos

May God bless.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@ The new architecture helped to introduce new features without performance and co
 
 ### Breaking Changes
 
-- Data schema outputted by romcal has been drastically remodeled. Please have look to the documentation for more informations.
+- Data schema outputted by romcal has been drastically remodeled. Please have look to the documentation for more information.
 - ascensionOnSunday, corpusChristiOnSunday, epiphanyOnSunday option names have been renamed.
 - General romcal option names have been unified. `calendar` is renamed to `scope` and take the values: `gregorian` or `liturgical`.
 - The provided liturgical year now correspond to the second year (the almost whole part of the civil year).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@ The new architecture helped to introduce new features without performance and co
 
 ### Breaking Changes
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -66,7 +66,7 @@ The new architecture helped to introduce new features without performance and co
 
 ### Breaking Changes
 
-- Data schema outputted by romcal has been drastically remodeled. Please have look to the documentation for more informations.
+- Data schema outputted by romcal has been drastically remodeled. Please have look to the documentation for more information.
```
tukusejssirs: I prefer using `output` in the past tense, although I know (but don’t use it) that `outputted` is grammatically correct word. We might consider changing this too.

```suggestion
- Data schema output by romcal has been drastically remodeled. Please have look to the documentation for more information.
```

tukusejssirs (approved):
Thank you, @sinapis-amicus, for taking your time to fix these typos.

I have just one comment that I prefer using `output` instead of `outputted` as the past tense of `output` (just like we use `put` for the past tense of `put`), although that is just a nitpick, as both `output` and `outputted` are correct in English.

/cc @emagnier

tukusejssirs (approved):

tests/config.test.ts
```diff
@@ -3,7 +3,7 @@ import Romcal from '../lib';
 const { RomcalConfig, LiturgicalDayConfig } = Romcal;
 
 describe('getConfig()', () => {
-  test('should get general config if country doesnt have default configurations', async () => {
+  test('should get general config if country doesn't have default configurations', async () => {
```
emagnier: The `'` need to be escaped here, otherwise it creates a syntax error in the codebase. 

Or better, to keep it friendly to read (and a consistent test codebase): 
```suggestion
  test('should get general config if country does not have default configurations', async () => {
```

emagnier (changes requested):

tests/config.test.ts
tukusejssirs: Yeah, I missed that, thanks, @emagnier! 🙏

emagnier (approved):

emagnier:
Thank you @sinapis-amicus for your help!! Very appreciated 🙏

PR #424, opened 2023-03-07
fix: xx-XX locale code ID not correctly formatted

### LINKED ISSUE #423: Not able to get  localized translations for pt-br, en-gb, en-ie
Hi,

I'm new to romcal and trying to implement it in my app, it works fine overall but I can't get it to output the localized translations for the following locales : pt-br, en-gb, en-ie.

I tried to run the react-app from the examples folder locally and I get the same problem.

Here are the steps to reproduce:

`cd .\react-app\`

change package.json to avoid an error with node 18 (Error: error:0308010C:digital envelope routines::unsupported).
 
```
"scripts": {
    "start": "react-scripts --openssl-legacy-provider start"
...
}
```

then 

`yarn`

then 

`yarn start`

I see the romcal app in the browser, if I select any locale from the dropdown like en, fr, es, cs, it... it works fine:

romcal_es

But if I select  pt-br, en-gb or en-ie, I don't have the translations no matter the calendar selected:

romcal_pt

Do you have any idea where it comes from?

Any help would be much appreciated.

emagnier:
Locale code ID was not passed to the i18next library with the right format. As a result, the locale code formats with 2 elements (`xx-XX`) were not correctly taken into account in romcal (e.g. `pt-BR`, `en-GB`, `en-IE`).

This fix ensures the right locale code format is always passed to i18next.

Fixes #423.


tukusejssirs (approved):

tukusejssirs:
LGTM. Thanks, @emagnier! :pray:

Issue #43, opened 2018-05-09
enhancement: Localisation tool


tukusejssirs:
I’ve been thinking for a quite long time, what would be the best way to deal with the localisation issues (like duplicates, typos, missing keys, etc).

I think we need a localisation/translation/internatisation tool. But which? :)

I’ve been contributing some translations for [MuseScore](musescore.org) at [Transifex](transifex.com), which is quite good, [free for open-source projects](https://www.transifex.com/pricing/) (scroll to the bottom to check that it is so). But it is probably too complex for our needs.

Then, just recently, I found something like [this](http://www.contactspls.com/translate/) (click on either `Contacts+ app` or `Google Play` in the row of your chosen language). It is based on [PASTT: PHP Android String Translation Tool](https://github.com/erickok/pastt/). It looks quite simple. :) And I like it. ;) But unfortunately, I can’t make it real. :( I might learn how to create a website with `pastt`, but it would require me to learn quite much and also time is a important element, too.

If you would agree with me and would like to implement `pastt` for keys (and other strings) translation, our table should contain:
- key name,
- `enUs` key value (as source language),
- key value text box for the target language,
- some notes for translation for the particular key (like the notes in empty text boxes of target language in [here](http://www.contactspls.com/translate/)).

The other thing is that after every translation submission, we might automatically generate a new `*.mjs` file, which would be sorted (that would fix issue #18), consistent, and afterward we might commit it to github (or just create a pull request). With this part I could help if the server, on which the localisation tool would be hosted, would be Linux.

tukusejssirs:
What about using [`jed`](https://messageformat.github.io/Jed/) and good old `gettext` (`.po` files)?

Dev1an:
I think we need to move away from storing all the text and dates in javascript files but instead put them in a database. This will make things a lot easier for translations, and all other editing tools in general.

tukusejssirs:
Do you suggest using `sqlite` or something else?

tukusejssirs:
Still, I think we need to have a translation platform using which we could (somehow) generate new versions of the translations.

tukusejssirs:
This issue is multifold:

- translation plaform:
  - l ike [Transifex](transifex.com), which does support directly loading translation from databases;
  - Transifex allows translators to translate online from any kind of device/OS or one can download the translations and translate offline using user-chosen tool and then upload it back;
- translation generation (plud checking and proofing);
  - check the files if they abide rules;
- translation storing (files/db):
[...]

PR #426, opened 2023-03-12
chore: fix building romcal on MS Windows (fix #425)

### LINKED ISSUE #425: unable to build source code
Hi,

I'm trying to build romcal locally on a windows 10 PC to see if I can add some translations to the locales files.
However it starts to build then fails.

I'm on node v18.12.1.

I don't have much experience in dev, here's the steps I follow:

`git clone git@github.com:romcal/romcal.git
`
`cd romcal/
`
`npm install
`
`npm run build
`

Then it starts building but fails with the following error:

romcal_build_error

It looks like [calendar] is undefined, I don't know if I'm doing something wrong, any help would be appreciated.

Thanks.

tukusejssirs:


scripts/build.ts
```diff
@@ -215,11 +215,11 @@ log(chalk.bold(`\n  –– ${chalk.red('Romcal')} builder ––`));
       // Do not output index.ts on iife format
       // and only output locale.iife.js on iife format
       if (
-        (format !== 'iife' || !/\/index\.ts$/.exec(p)) &&
+        (format !== 'iife' || !/[\\/]index\.ts$/.exec(p)) &&
```
tukusejssirs, line 218: @emagnier, we could use `node:path.sep` instead of `[\\/]`, not only in regexes, but only in logs. WDYT?

emagnier: Good idea. But do not spend too much time on a refactor here (except if something isn't working on Windows), as I started a refactor on that part too (among others, converting the codebase to a monorepo and refactoring the build scripts).

emagnier (approved):
Thank you for the fix!

scripts/build.ts
tukusejssirs: Okay, then I won’t change this to `node:path.sep`.

> I started a refactor on that part too (among others, converting the codebase to a monorepo and refactoring the build scripts

Where can I find this work of yours? I couldn’t find anything related in your fork.

Regarding monorepo: which monorepo tool you want to use? I presume Yarn-based, although I prefer using [Nx](https://nx.dev/) monorepo. What will be are apps/libs, how do you want to split romcal?

emagnier: It's not pushed to GitHub yet, and it's progressing a bit slowly, as I don't have much free time to work on it currently...

FYI I use `pnpm` as it is faster than `npm` and `yarn`, and also use less disk space (especially when caching npm packages). About other tooling, I'm using `esbuild` & `tsup`, and I'm considering using `turborepo` to orchestrate and streamline all of that.

It also requires quite some code refactoring, and not everything is ready yet. At least one of the goal is to split the core romcal library and the data (calendar definitions, locales...). I hope to be able to share something very soon.

tukusejssirs: Yeah, I have recently started using `pnpm` too, but not in all repos (for some stuff needs to be modified).

I have read about Turborepo, but I haven’t actually used it. I use Nx and I am quite happy with it. Have you ever used Nx?

Regarding splitting romcal into libs: how about this:
- core romcal library;
- GRC calendar definitions (`proper-of-times.ts`, `proper-of-saints.ts`);
- particular calendars;
- localisations.

I’ve been even thinking about changing the data format from TS/JS to JSON with JSON schema defined for all calendars and locales. That way, the data could be used by any other app in any language. (Something similar was suggested by [@Dev1an](https://github.com/Dev1an) in #43.) WDYT?

---

Anyway, we should move this conversation to discussions. :wink:

emagnier: I never used Nx yet, just read a bit on it. 
It's also currently on the radar to change the data format: make them easier to edit/improve (for contributors), and easier to consume (by romcal or any other tool).
More details soon, and yes in another place from here 😄

tukusejssirs: Impatiently waiting to review your work. :wink:

PR #428, opened 2023-03-12
refactor: import named exports from `node:path` instead of using the default import

tukusejssirs:
At the same time, sort the imports in `scripts/build.ts`.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,9 +1,10 @@
+import { dirname, resolve } from 'node:path';
+
 import chalk from 'chalk';
 import { generateDtsBundle } from 'dts-bundle-generator';
 import { build, Format, Platform } from 'esbuild';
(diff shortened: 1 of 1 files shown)


emagnier (approved):

tukusejssirs:
@emagnier, [some jobs failed](https://github.com/romcal/romcal/actions/runs/4452181376) as I forgot to rebase my branch, but it is actually merged into `dev`. Moreover, I believe that re-running the jobs will create another romcal version, so I think we can consider this being okay.

WDYT? :thinking:

emagnier:
Probably a conflict while updating the romcal version? Because of 2 PRs merged on the same time?

By the way, I've manually updated the package-lock.json on the dev branch, and everything is back to normal on the CI & npm/tag versions.

Issue #433, opened 2023-03-23
Add a `LiturgicalDay` property if a celebration was transfered due to occurrence


tukusejssirs:
I think it might be a good idea to add a `LiturgicalDay` property if a celebration was transfered due to occurrence, for example: `joseph_spouse_of_mary` was transfered this year from 19 March 2023 to 20 March 2023 due to occurence with a Lenten Sunday.

We might also add a reason why it was transfered. It could be sufficient to add a key of the other celebration (e.g. this year it would be `lent_4_sunday`).

emagnier:
The `dateExceptions` property from the calendar definition already contains the explanation for what reason it might be transferred to another date. Currently, this property isn't output in the final `LiturgicalDay` object, which is an oversight.

PR #434, opened 2023-03-28
fix: dateExceptions not correctly merged from previous item definition

### LINKED ISSUE #432: Fix transfer of `joseph_spouse_of_mary` in `slovakia.ts`
I have just noticed that `joseph_spouse_of_mary` is not transfered this year to Monday 20 Mar 2023 in `slovakia.ts`. It works as expected in GRC.

https://github.com/romcal/romcal/blob/5d3fc9c7be367bbb679bbcf63479a4db5e461241/lib/general-calendar/proper-of-saints.ts#L218-L234

https://github.com/romcal/romcal/blob/5d3fc9c7be367bbb679bbcf63479a4db5e461241/lib/particular-calendars/slovakia.ts#L21-L23

I think adding `dateExceptions` to `slovakia.ts` would work this issue around, but I think this needs to be fixed more deeply.

I believe this issues comes from the fact that `joseph_spouse_of_mary` is not a solemnity in Slovakia. I think that each holy day of obligation defined in the GRC should be transfered to the same date when necessary, regardless if that celebration is a HDO in a particular calendar or not.

emagnier:
- `dateExceptions` property not correctly merged from previous liturgical day definition.
- `dateExceptions` property not output on the final `LiturgicalDay` object.

Fixes #432 & #433.

### DIFF
--- a/lib/models/liturgical-day-def.ts
+++ b/lib/models/liturgical-day-def.ts
@@ -110,6 +110,8 @@ export default class LiturgicalDayDef implements BaseLiturgicalDayDef {
       ? input.dateExceptions
       : input.dateExceptions
       ? [input.dateExceptions]
+      : previousDef
+      ? previousDef.dateExceptions
       : [];
 
     if (!input.precedence && !previousDef) {
--- a/lib/models/liturgical-day.ts
+++ b/lib/models/liturgical-day.ts
@@ -8,6 +8,7 @@ import { Id } from '../types/common';
 import {
   BaseLiturgicalDay,
   DateDef,
+  DateDefException,
   FromCalendarId,
   i18nDef,
   LiturgyDayDiff,
@@ -26,6 +27,7 @@ class LiturgicalDay implements BaseLiturgicalDay {
   readonly id: Id;
   readonly date: string;
   readonly dateDef: DateDef;
+  readonly dateExceptions: DateDefException[];
   readonly precedence: Precedence;
   readonly rank: Rank;
   readonly allowSimilarRankItems: boolean;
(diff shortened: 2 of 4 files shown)


lib/models/liturgical-day-def.ts
```diff
@@ -110,6 +110,8 @@ export default class LiturgicalDayDef implements BaseLiturgicalDayDef {
       ? input.dateExceptions
       : input.dateExceptions
       ? [input.dateExceptions]
+      : previousDef
+      ? previousDef.dateExceptions
       : [];
```
tukusejssirs, line 115: Could this be written in a more readable way please? :wink:

```suggestion
      ? input.dateExceptions
      : input.dateExceptions
        ? [input.dateExceptions]
        : previousDef
          ? previousDef.dateExceptions
          : [];
```

tukusejssirs (approved):
Thanks, @emagnier, for such quick fix. :pray:

I just have a small nitpick, but you can ignore it if you want. :wink:

lib/models/liturgical-day-def.ts
emagnier: These are flattened by Prettier (we don't have hands on it). But it highlights that, generally, it's not a very good practice to chain many conditional ternary operators. I did it anyway here (as a quick fix), knowing that I already rewrote this part differently in my current refactor (for the monorepo).

tukusejssirs: Yeah, `prettier`, I keep fighting with it. :wink:
> But it highlights that, generally, it's not a very good practice to chain many conditional ternary operators.

Yeah, but are two ternary operators considered many? 🤔

> I already rewrote this part differently in my current refactor (for the monorepo).

Okay, then disregard this conversation.

Anyway, I long to see your WIP on monorepo. Could you create a WIP/draft PR please? Thanks! :pray:

PR #447, opened 2023-08-16
fix: calculation of days in a month to be done in UTC

emagnier:
- Calculation of days in a month to be done in UTC.

### DIFF
--- a/lib/utils/dates.spec.ts
+++ b/lib/utils/dates.spec.ts
@@ -1,0 +1,13 @@
+import { daysInMonth } from './dates';
+
+test('daysInMonth returns the correct number of days in the month of a given date', () => {
+  const date1 = new Date('2023-02-15T00:00:00Z'); // February 2023 (28 days)
+  const date2 = new Date('2024-02-15T00:00:00Z'); // February 2024 (29 days)
+  const date3 = new Date('2023-05-15T00:00:00Z'); // May 2023 (31 days)
+  const date4 = new Date('2023-06-15T00:00:00Z'); // June 2023 (30 days)
+
+  expect(daysInMonth(date1)).toBe(28);
+  expect(daysInMonth(date2)).toBe(29);
+  expect(daysInMonth(date3)).toBe(31);
+  expect(daysInMonth(date4)).toBe(30);
+});
--- a/lib/utils/dates.ts
+++ b/lib/utils/dates.ts
@@ -53,7 +53,7 @@ export const isValidDate = (maybeDate: unknown): boolean => {
  * @return     {number}  Total number of dates in the particular month
  */
 export const daysInMonth = (date: Date): number => {
-  return new Date(date.getUTCFullYear(), date.getUTCMonth() + 1, 0).getUTCDate();
+  return new Date(Date.UTC(date.getUTCFullYear(), date.getUTCMonth() + 1, 0)).getUTCDate();
 };
 
 /**


lib/utils/dates.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { daysInMonth } from './dates';
+
+test('daysInMonth returns the correct number of days in the month of a given date', () => {
+  const date1 = new Date('2023-02-15T00:00:00Z'); // February 2023 (28 days)
+  const date2 = new Date('2024-02-15T00:00:00Z'); // February 2024 (29 days)
+  const date3 = new Date('2023-05-15T00:00:00Z'); // May 2023 (31 days)
+  const date4 = new Date('2023-06-15T00:00:00Z'); // June 2023 (30 days)
```
tukusejssirs, line 7: @emagnier, do you think those comments are needed? IMO there is no value in keeping them, as the month is derived from the `new Date()` argument and the number of days are in `toBe()` method.

What we could do is to rename `date{1,4}` to `${month}${year}` if we want it to be more readable.

WDYT? :thinking:

emagnier: Good idea, I'll keep that in mind for a future update.

tukusejssirs: I usually omit all useless comments, esp if the variable/class/function/method names) are named properly so that the code would make sense just by reading it. :man_shrugging:

emagnier: At the same time, we're not dealing with the code of the Romcal library here, but with tests. So, this remains somewhat anecdotal in this context. 

What's important in the tests here is that whatever is being tested is clearly identified (whether through comments or variable names), and I appreciate it when everything is readable and aligned.

tukusejssirs: > At the same time, we're not dealing with the code of the Romcal library here, but with tests. So, this remains somewhat anecdotal in this context.

Well, yeah, but the tests were created in this commit, therefore IMO this is the right place to discuss (although I am late to the party, as the PR was already merged).

> What's important in the tests here is that whatever is being tested is clearly identified (whether through comments or variable names), and I appreciate it when everything is readable and aligned.

So do I, however I strive to eliminate unnecessary comments when the code can ‘speak’ for itself so that it is clear enough what it does.

---

Therefore, I’d prefer something like this:

```ts
const feb2023 = new Date('2023-02-15T00:00:00Z');
const feb2024 = new Date('2024-02-15T00:00:00Z');
const may2023 = new Date('2023-05-15T00:00:00Z');
const jun2023 = new Date('2023-06-15T00:00:00Z');

expect(daysInMonth(feb2023)).toBe(28);
expect(daysInMonth(feb2024)).toBe(29);
expect(daysInMonth(may2023)).toBe(31);
expect(daysInMonth(jun2023)).toBe(30);
```

Along with the test description, one knows exactly what the test does.

YMMV. 🤷‍♂️

emagnier: As a senior developer, I completely understand what you meant and I agree with it. Let's not spend too much time on these discussions and details, especially since this PR has already been merged. As I mentioned before, I will work on improving this at the next opportunity (feel free to do it if you'd like, or to contribute to such improvements in general).

And otherwise, there's no need to be afraid of approaching it iteratively (through continuous improvement), as long as there's already a qualitative contribution in each iteration. Of course, every code review and suggestion are important to consider. However, it's also important to find the right balance, so as not to spend too much time on these kinds of details and to stay focused on what enables progress and bring the most added value to this project overall 😄.

PR #456, opened 2023-09-19
Update pl.ts

ciesiolka:
Align day names as they are printed in official liturgical books

PR #403, opened 2022-11-06
feat(hu): add Hungarian locale

MMate2007:
This PR adds initial Hungarian localisation based especially on the Hungarian Roman Missal (1991, editio typica altera), secondarily on the Hungarian [_Ordo >

Only those celebrations names were localised which are celebration in the national Hungarian calendar, however, no missing celebrations were added (those nee>

fix #248

PR #458, opened 2023-09-19
Fixed typos in pl.ts

ciesiolka:


tukusejssirs:
Hmm, @ciesiolka, why don’t you simply merge the changes in this PR with those in #456? :thinking:

lib/locales/pl.ts
```diff
@@ -259,7 +259,7 @@ export const locale: Locale = {
     francis_xavier_priest: 'Św. Franciszka Ksawerego, prezbitera',
     frederic_ozanam_founder: 'Bł. Fryderyka Ozanama, założyciela',
     friday_of_the_passion_of_the_lord: 'Wielki Piątek',
-    gabriel_taurin_dufresse_bishop: 'Św. Gabriel-Taurina Dufresse, biskupa i męczennika',
+    gabriel_taurin_dufresse_bishop: 'Św. Jana Dufresse, biskupa i męczennika',
```
tukusejssirs, line 262: This one is strange. Why the change of the name? :thinking:

lib/locales/pl.ts
```diff
@@ -359,13 +359,13 @@ export const locale: Locale = {
     margaret_of_scotland: 'Św. Małgorzaty Szkockiej',
     maria_goretti_virgin: 'Św. Marii Goretti, dziewicy i męczennicy',
     marie_eugenie_of_jesus_milleret_de_brou_virgin:
-      'Św. Marii Eugenii od Jezusa Milleret de Brou, dziewicy i założycielky',
+      'Św. Marii Eugenii od Jezusa Milleret de Brou, dziewicy i założycielki',
     mark_evangelist: 'Św. Marka, Ewangelisty',
-    martha_of_bethany_mary_of_bethany_and_lazarus_of_bethany: 'Św. Marty, Marii i Łazarza',
+    martha_of_bethany_mary_of_bethany_and_lazarus_of_bethany: 'Świętych Marty, Marii i Łazarza',
```
tukusejssirs, line 364: All celebrations have the canonisation title shortened; why this one not? :thinking:

lib/locales/pl.ts
```diff
@@ -390,7 +390,7 @@ export const locale: Locale = {
     matthias_apostle: 'Św. Macieja, Apostoła',
     maurice_tornay_priest: 'Bł. Maurycego Tornaya, prezbitera i męczennika',
     maximilian_mary_raymund_kolbe_priest: 'Św. Maksymiliana Marii Kolbego, prezbitera i męczennika',
-    mederic_of_autun_and_droctoveus_of_autun_abbots: 'Św. Mederyka i Droctovæi, opatów',
+    mederic_of_autun_and_droctoveus_of_autun_abbots: 'Świętych Mederyka i Droktoweusza, opatów',
```
tukusejssirs, line 393: Again: canonisation name typed in full: why?

tukusejssirs (changes requested):

lib/locales/pl.ts
ciesiolka: His full name was Jean (John) Gabriel Taurin Dufresse. This is the name Vatican is using in documents in English - see: https://www.vatican.va/news_services/liturgy/saints/ns_lit_doc_20001001_zhao-rong-compagni_en.html

> Blessed John Gabriel Taurin Dufresse, M.E.P., Bishop. He was arrested on 18 May 1815, taken to Chengdu, condemned and executed on 14 September 1815.

All sources in polish I could find call him using his first name.

lib/locales/pl.ts
ciesiolka: `Św.` means `Świętego` (singular). When there are more than one saint `Świętych` (plural) is used. The convention is already followed in the file - check out entry for `Basil the Great and Gregory Nazianzen` or `Joachim and Anne`.

The proper way to abbreviate plural `Świętych` in polish language would be `Św.Św. Marty, Marii i Łazarza` which looks weird and nobody is using that (maybe except for linguists).

lib/locales/pl.ts
ciesiolka: Same as above.

lib/locales/pl.ts
tukusejssirs: Then it is my mystake amd I need to fix it everywhere. 🤦‍♂️

Thanks for sharing the source with me! :pray:

I am a bit confused though, as [Wikipedia](https://en.m.wikipedia.org/wiki/Louis_Gabriel_Taurin_Dufresse) (yes, a third-party source ‘against’ your second-party one, according to [my source categorisation](https://github.com/romcal/romcal/pull/403#issuecomment-1704299089)) replaces _John_ with _Louis_ for some reason. The [French Wikipedia](https://fr.m.wikipedia.org/wiki/Gabriel-Taurin_Dufresse) uses simply _Gabriel-Taurin_. 🤔

lib/locales/pl.ts
tukusejssirs: > `Św.` means `Świętego` (singular).

I understand this. :wink:

> When there are more than one saint `Świętych` (plural) is used. The convention is already followed in the file - check out entry for `Basil the Great and Gregory Nazianzen` or `Joachim and Anne`.

For some zome reason, I failed to notice this. I need to note this in the Polish l10n rules (not yes published).

> The proper way to abbreviate plural `Świętych` in polish language would be `Św.Św. Marty, Marii i Łazarza` which looks weird and nobody is using that (maybe except for linguists).

What?! 👀 That indeed looks weird. I have never ever seen a language where they would double the shortening (in Slovak, we use `sv.`/`bl.` for both singular and plural; in English, an `s` is appended; in Latin, the first letter is doubled, i.e. `S.` → `Ss.`, `B.` → `Bb.`).

emagnier:
Since I don't speak Polish at all, I'm not of much help here. Is this PR ready to be merged? Or are there still some issues to address? @tukusejssirs, can you approve this PR if it's ok with you?

ciesiolka:
> Is this PR ready to be merged? Or are there still some issues to address? @tukusejssirs, can you approve this PR if it's ok with you?

From my POV it's ready to be merged. As of now I am not going to modify it any further.

lib/locales/pl.ts
```diff
@@ -487,7 +487,7 @@ export const locale: Locale = {
     stanislaus_of_szczepanow_bishop_patron_of_poland: 'Św. Stanisława, biskupa i męczennika, głównego patrona Polski',
     stephen_i_of_hungary: 'Św. Stefana Węgierskiego',
     stephen_the_first_martyr: 'Św. Szczepana, pierwszego męczennika',
-    sunday_of_the_word_of_god: '{{week}}. Niedziela zwykła, lub Niedziela Słowa Bożego',
+    sunday_of_the_word_of_god: '{{week}} Niedziela zwykła, czyli Niedziela Słowa Bożego',
```
tukusejssirs, line 490: @ciesiolka, just to confirm: `3 Niedziela zwykła, czyli Niedziela Słowa Bożego` is the form used in Polish, right? There is no `.` after the week number? Or do you use Roman Numerals (`III`) instead? :thinking:

tukusejssirs (approved):
Unless my nitpick regarding the usage of full stop (`.`) is incorrect, LGTM.

lib/locales/pl.ts
ciesiolka: There's a rule which says that ordinals have appended period, but it can be omitted when it's clear from the context that the number is an ordinal. This is the case. Also, official liturgical books don't use periods in titles.

ciesiolka:
Yep, full stop usage is correct in the current form. Can be merged.

lib/locales/pl.ts
tukusejssirs: Understood. If possible, use the format used by the official liturgical books (prefer books more recently published to the older ones), i.e. if there is no `.` after ordinal numbers, don’t use them in any l10n key (`.seasons` nor `.names`).

emagnier:
👍

PR #459, opened 2023-09-23
feat(el): greek calendar & locale

ciesiolka:
This commit contains:
- translation of saint names as appear in Roman Missal in Greek language (2006)
- celebrations from proper calendars of Greek Dioceses of Athens, Syros and Tinos (all from 2013)
- saints mentioned in those calendars were added to `martyrology.ts`, except for:
  - dedications of own churches
  - `timothy_of_ephesus_and_titus_the_first_bishop_of_crete_bishops` - variant of existing `timothy_of_ephesus_and_titus_of_crete_bishops`
  - `lydia_of_philippi_first_european_christian` - variant of `lydia_of_philippi`
  - `zechariah_and_elizabeth_parents_of_john_the_baptist`
  - `twelve_apostles`

Only some of the inserted saints were translated to Latin. To be fixed in a separate PR.

Names of some of saints inserted to the General Calendar after 2006 were added. If found, please report missing ones.

Important notice, not directly related to the commit: The Easter in Greece is celebrated according to Julian calendar, in the same date as in Orthodox Church. It means the Ash wednesday, Palm Sunday, Pentecost, Corpus Christi, etc. are shifted. I don't know if romcal supports that.

PR #460, opened 2023-09-25
Allow calendars to be computed from a Julian Easter date

emagnier:
This provides the ability to compute a calendar, based on a _Julian Easter date_ (in addition to the _Gregorian Easter date_, which remain the default calculation type).

Within a calendar definition, you now can specify an `easterCalculationType` property, which accepts a `gregorian` or `julian` value (defaults to `gregorian`).

```ts
// Example for greece.ts
export class Greece extends CalendarDef {
  parentCalendar = Europe;

  particularConfig: ParticularConfig = {
    easterCalculationType: 'julian',
  };

  inputs: Inputs = {}
}
```

It's also possible to specify the Easter calculation type, directly from the Romcal settings : 
```ts
const romcal = new Romcal({ easterCalculationType: 'julian' });
const calendar = await romcal.generateCalendar(2019);
```

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -94,7 +94,6 @@ import {
 import { MartyrologyCatalog, MartyrologyItem, SaintCount, SaintDate, SaintDateDef } from './types/martyrology';
 import {
   addDays,
-  computeGregorianEasterDate,
   dateDifference,
   Dates,
   daysInMonth,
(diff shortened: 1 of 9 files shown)


emagnier:
@ciesiolka this might interest you, and will address a potential need exposed here https://github.com/romcal/romcal/pull/459.

Issue #126, opened 2019-12-15
Add Hungarian locale


tukusejssirs:
Just like with the `cs.js` (#125), I’d like to add Hungarian locale, too, when I’ll find some time.

And again, this might be done for `v2.0.0`, but I think the separation and localisation improvements will be much time-consuming, so I will do this when I’ll have some time. Time issue is just kind of a reminder to myself.

Issue #136, opened 2019-12-18
Add `dedicationOfAParticularChurch` to `general.js`


tukusejssirs:
`dedicationOfAParticularChurch` celebration definition should be moved from particular calendars to `general.js`. It is celebrated on 26 Oct as solemnity in churches of which the date of consecration is unknown.

emagnier:
I'm not sure about moving to general, because this solemnity will overwrite the feria or any other celebration on this day. And this might not be relevant for all church.
And actually, "optional solemnity" doesn't exist in romcal.

tukusejssirs:
After searching a bit on the Internet, you might be right. Even with your implicit suggestion to remove this celebration from `romcal` (non-parish calendars) altogether. Then, in a separate repo, we could create the modification of `romcal` for _local church calendars_ (i.e. the diocesan and parish calendars).

[Wiki](https://en.wikipedia.org/wiki/General_Roman_Calendar) has several celebration definitions in different particular or religious calendars (and on different dates). It even adds the [following](https://en.wikipedia.org/wiki/General_Roman_Calendar#Local_calendars):

> The calendar for a diocese is typically based on a national calendar, such as those listed above, with a few additions. For instance, the anniversary of the dedication of the cathedral is celebrated as a Solemnity in the cathedral church and as a Feast in all the other churches of the diocese. The feast day of the principal Patron saint of the diocese is celebrated as a Feast throughout the diocese. ( [General Norms for the Liturgical Year and the Calendar](http://www.catholicculture.org/library/view.cfm?recnum=5932), Table of Liturgical Days according to their order of precedence, 4 and 8.)
>
> The calendar of a parish is based on the calendar of its diocese, but—in addition to the celebrations in the diocesan calendar—there are other celebrations, including the anniversary of the dedication of the parish church and the feast day of the principal Patron saint of the church, both of which are celebrated as Solemnities.

tukusejssirs:
It has just occured to me that we could add a romcal option (like `addDedicationOfAParticularChurch`) that would add the celebration to the output if set to `true` (else it would not output it; `false` by default), because many Ordos (Slovakia, Poland, Canada, etc) are including it. IMO if we want to implement #20, we should want to have all options available in order to be as faithfull as possible.

Note to myself: when it’s gonna be implemented, `dedication_of_consecrated_churches` is to be celebrated on the following dates in the following calendars:
[...]

PR #461, opened 2023-10-06
New CI script to check locale data consistency

emagnier:
New script: `npm run locale-checks`, to check the consistency of the locale files.

🔴  **It reports errors when:**
  - [1] There are useless martyrology items.
  - [2] There are missing martyrology items.
  - [3] There are missing localized 'en' items (in the Romcal project, English takes precedence as the primary language, so all content should be localized in English).
  - [4] There are useless localized items.
  - [5] The martyrology items are not sorted alphabetically.
  - [6] The localized name items are not sorted alphabetically.

⚠️  **It issues warnings when:**
  - [7] There are missing localized items for the Proper of Time and the General Roman calendar.

This script is also integrated into the CI (executed on each new and modified PR). If an error occurs, the PR will not be validated. And I took advantage of it to correct some inconsistencies and missing data in the current locale & martyrology files.

In a future PR, I would like to add the ability to run checks for any locale and calendar, and automatically generate a documentation for what still needs to be translated, along with translation rates per calendars.


lib/catalog/martyrology.ts
```diff
@@ -978,9 +978,6 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       name: 'Declan',
       titles: [Titles.Bishop],
     },
-    dedication_of_consecrated_churches: {
-      name: 'Dedication of Consecrated Churches whose date of Consecration is unknown',
-    },
```
tukusejssirs, line 983: Why is this removed? If we should list event celebrations in `martyrology.ts`, we should keep `dedication_of_consecrated_churches`.

```suggestion
    dedication_of_consecrated_churches: {
      name: 'Dedication of Consecrated Churches Whose Date of Consecration is Unknown',
    },
```

lib/locales/fr.ts
```diff
@@ -272,6 +272,8 @@ export const locale: Locale = {
     damasus_i_pope: 'Saint Damase Ier, pape († 384)',
     damien_de_veuster_priest: 'Saint Père Damien, prêtre et Missionnaire Picpus († 1889)',
     daniel_brottier_priest: 'Binheureux Daniel Brottier, prêtre, apôtre des Orphelins d’Auteuil († 1936 à Paris)',
+    dedication_of_consecrated_churches:
+      'Dédicace des églises consacrées dont on ne connaît pas la date de consécration', // src: missal_fr_2021_ed3
```
tukusejssirs: Just a nitpick: we might change this, however, everywhere I use `mr_${lang}_${year}_${edition}` format.

```suggestion
      'Dédicace des églises consacrées dont on ne connaît pas la date de consécration', // src: mr_fr_2021_ed3
```

lib/locales/it.ts
```diff
@@ -264,6 +264,7 @@ export const locale: Locale = {
     clement_i_pope: 'San Clemente I, papa e martire',
     columba_of_iona_abbot: 'San Colombano, abate',
     columba_of_iona_abbot_copatron_of_ireland: 'San Colombano, abate e patrono secondario dell’Irlanda',
+    columban_of_luxeuil_abbot: 'San Colombano, abate', // src: missal_it_2020_ed3
```
tukusejssirs: Just a nitpick: we might change this, however, everywhere I use `mr_${lang}_${year}_${edition}` format.

```suggestion
    columban_of_luxeuil_abbot: 'San Colombano, abate', // src: mr_it_2020_ed3
```

lib/locales/it.ts
```diff
@@ -432,6 +438,7 @@ export const locale: Locale = {
     our_lady_of_fatima: 'Beata Vergine Maria di Fatima',
     our_lady_of_guadalupe: 'Beata Vergine Maria di Guadalupe',
     our_lady_of_guadalupe_patroness_of_the_americas: 'Beata Vergine Maria di Guadalupe, Patrona delle Americhe',
+    our_lady_of_loreto: 'Beata Vergine Maria di Loreto', // src: missal_it_2020_ed3
```
tukusejssirs: Just a nitpick: we might change this, however, everywhere I use `mr_${lang}_${year}_${edition}` format.

```suggestion
    our_lady_of_loreto: 'Beata Vergine Maria di Loreto', // src: mr_it_2020_ed3
```

lib/locales/it.ts
```diff
@@ -447,6 +454,7 @@ export const locale: Locale = {
     patrick_of_ireland_bishop_patron_of_ireland: 'San Patrizio, vescovo e patrono dell’Irlanda',
     paul_miki_and_companions_martyrs: 'San Paolo Miki sacerdote e compagni, martiri',
     paul_of_the_cross_priest: 'San Paolo della Croce, sacerdote',
+    paul_vi_pope: 'San Paolo VI, papa', // missal_it_2020_ed3
```
tukusejssirs: Just a nitpick: we might change this, however, everywhere I use `mr_${lang}_${year}_${edition}` format.

```suggestion
    paul_vi_pope: 'San Paolo VI, papa', // src: mr_it_2020_ed3
```

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+import { SEASONS } from '../lib/constants/seasons';
+import { WEEKDAYS } from '../lib/constants/weekdays';
+import { GeneralRoman } from '../lib/general-calendar/proper-of-saints';
+import { locales } from '../lib/locales';
+import { CalendarDef } from '../lib/models/calendar-def';
+import { particularCalendars } from '../lib/particular-calendars';
+import { toPackageName } from '../lib/utils/string';
+import { RomcalBuilder } from './bundle';
+
+type LabelType = 'info' | 'warn' | 'error';
```
tukusejssirs: I find the name of this `type` a bit confusing. It seems to me it is actually used `LogLevel`.

And why don’t you use `enum` instead? :thinking:

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+      return bgColor[type].bold('\n ERR  ') + ' ';
+  }
+};
+
+const list = (type: LabelType, items: string[]): string => color[type](items.map(dasherize).join('\n'));
+
+/**
+ * Output log if the values are not falsy.
+ */
+const logIf = (type: 'warn' | 'error', description: string, values: string[] | StringArrayRecord | boolean): void => {
```
tukusejssirs: If `LabelType`/`LogLevel` is an `enum`, would’t it be better to use `LabelType.warn | LabelType.error` as the type of `type` arg? :thinking:

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+  if (values === false) return;
+  console[type](label(type) + chalk.bold(description));
+  if (values === true) return;
+  if (Array.isArray(values)) {
+    console[type](list(type, values));
+  } else {
+    Object.entries(values).forEach(([namespace, keys]) => {
+      console[type]('       ' + color[type].bold(`${namespace}:`));
```
tukusejssirs:
```suggestion
      console[type](`       ${color[type].bold(`${namespace}:`)}`);
```

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+  // For a region locale (XxXx), merge it with the base locale (Xx) if exists
+  const locale =
+    localeKey.length === 4 && Object.prototype.hasOwnProperty.call(locales, localeKey.slice(0, 2))
+      ? merge(locales[localeKey.slice(0, 2)], locales[localeKey])
+      : locales[localeKey];
+
+  return findMissingInArrayWithNamespacedLocaleKeys(Array.from(localeComputedKeys), locale, option);
+};
+
+const devLocale = { id: 'dev' };
```
tukusejssirs: I didn’t know we can use `dev` as an ID. Where is that defined? I thought it should of the `Locale` type which is defined as `string`, however, it needs to be found in one of the l10n files.

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+logIf('error', `Useless martyrology items:`, uselessMartyrologyKeys);
+
+/**
+ * [2] If there are missing martyrology items.
+ */
+const missingMartyrologyKeys = findMissingInArray(allMartyrologyKeys, Object.keys(Martyrology.catalog));
+logIf('error', `Missing martyrology items:`, missingMartyrologyKeys);
+
+/**
+ * [3] If there are missing localized 'en' items
```
tukusejssirs: Nitpick: I would remove the trailing dots (`.`) from short code comments, however, I definitely would try to be consistent → either add a dot here or remove them from other short code comments.

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Compute data to check
+ */
+for (let i = 0; i < allCalendars.length; i++) {
+  const calendar = allCalendars[i];
+
+  // Init config
+  const isGRC = calendar.name === GeneralRoman.name;
+  const builder = isGRC ? new RomcalBuilder(devLocale) : new RomcalBuilder(devLocale, calendar);
```
tukusejssirs: Can’t `calendar` be set to to `general`? :thinking:

scripts/locale-checks.ts
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * [6] If the localized name items are not sorted alphabetically.
+ */
+Object.values(locales).forEach((locale) => {
+  const areNotSortedNames =
+    Object.keys(locale.names ?? {})
+      .sort()
+      .join('') !== Object.keys(locale.names ?? {}).join('');
+  logIf('error', `Localized '${u(locale.id)}' names are not sorted alphabetically.`, areNotSortedNames);
+});
```
tukusejssirs: Didn’t you want to create a function to check whether the keys are sorted? You use nearly the same code, apart from the input object (`Martyrology.catalog` vs `locale.names ?? {}`) and the error string (`Martyrology keys are not sorted alphabetically.` vs `Localized '${u(locale.id)}' names are not sorted alphabetically.`).

lib/locales/es.ts
```diff
@@ -272,9 +272,7 @@ export const locale: Locale = {
     daniel_brottier_priest: 'Beato Daniel Brottier, presbítero',
     david_of_wales_bishop: 'San David, obispo',
     david_of_wales_bishop_patron_of_wales: 'San David, obispo y Patrono de Gales',
-    dedication_of_consecrated_churches_on_last_sunday_of_october:
-      'Dedicación de iglesias consagradas el último domingo de octubre',
-    dedication_of_consecrated_churches_on_october_25: 'Dedicación de iglesias consagradas el 25 de octubre',
+    dedication_of_consecrated_churches: 'Dedicación de iglesias consagradas cuya fecha de Consagración se desconoce', // TODO: l10n to be review: Dedication of Consecrated Churches whose date of Consecration is unknown
```
tukusejssirs:
```suggestion
    dedication_of_consecrated_churches: 'Dedicación de iglesias consagradas cuya fecha de Consagración se desconoce', // TODO: l10n to review: Dedication of Consecrated Churches Whose Date of Consecration is Unknown
```

lib/locales/pt-br.ts
```diff
@@ -251,8 +251,7 @@ export const locale: Locale = {
     cyril_of_jerusalem_bishop: 'São Cirilo de Jerusalém, bispo e doutor da Igreja',
     damasus_i_pope: 'São Dâmaso I, papa',
     daniel_brottier_priest: 'Beato Daniel Brottier, presbítero',
-    dedication_of_consecrated_churches_on_last_sunday_of_october: 'Dedicação de Igrejas Consagradas no Último Domingo de Outubro',
-    dedication_of_consecrated_churches_on_october_25: 'Dedicação de Igrejas Consagradas em 25 de Outubro',
+    dedication_of_consecrated_churches: 'Dedicação de Igrejas Consagradas', // TODO: l10n to be review: Dedication of Consecrated Churches whose date of Consecration is unknown
```
tukusejssirs:
```suggestion
    dedication_of_consecrated_churches: 'Dedicação de Igrejas Consagradas', // TODO: l10n to review: Dedication of Consecrated Churches Whose Date of Consecration is Unknown
```

tukusejssirs (changes requested):
Hey, @emagnier! Thanks for this PR! :pray: I wanted to work som something similar (albeit as Jest unit tests), but unfortunately, it takes me much longer than I anticipated to fix #126 and review/fix #459, after which I wanted to work on this and on the monorepo stuff (cf #445). Anyway, FYI, these are the tests I wanted to check (please compare these with the checks implementation in this PR please):

- check if only keys defined in `en` are defined in all other locales:
  - this should be done;
- check if only keys defined in `en` are used in all calendar files (incl the Proper of Time and Proper of Saints):
  - this seems to be out of scope of this PR;
- check if all `en` l10n keys (or parts of them, based on the celebrations in all calendars) are defined in `martyrology.ts`:
  - this should be done.

Why did you chose to create a separate script for these checks? Jest unit tests could be better for this purpose, without the need of an extra NPM script.

> New script: `npm run locale-checks`, to check the consistency of the locale files.

Maybe we should rename the `locale-checks` script to something else, as we might have some other check types in the future. Consider using `checks:locale` or even `checks:l10n`.

>   - [1] There are useless martyrology items.

I am a bit against this check. Yeah, I understand the reason why what only those saints/blesseds/celebrations in `martyrology.ts` which are used by romcal, however, I still think that it should contain all / as much as possible / most saints (at least all listed in _Martyrologium Romanum_ from 2004 and all persons beatified/canonised since then). Moreover, I think ever servants of God and venerables could be listed in there. That said, it’ll be a WIP.

I also wanted to suggest in #445 that we should create a separate library from `martyrology.ts` in the Nx monorepo. I want talk about my monorepo-related suggestions in #445 more in depth. 

>   - [4] There are useless localized items.

This one hard a bit.

While I understand that items not used by romcal could be removed, from the above-mentioned point of view (i.e. leaving items not used by romcal in `marytorology.ts`), I think we should flag to remove only those items that are not inscribed in `martyrology.ts`.

> In a future PR, I would like to add the ability to run checks for any locale and calendar, and automatically generate a documentation for what still needs to be translated, along with translation rates per calendars.


---

Besides adding these _checks_, it might a good idea to create a _fix_ script too for some of these issues (e.g. sorting items alphabetically).

lib/catalog/martyrology.ts
emagnier: This one has been removed because there are neither celebrations related to `dedication_of_consecrated_churches` nor any martyrology reference to a `dedication_of_consecrated_churches` entity.

Furthermore, the martyrology is not intended for such types of celebrations. It should only contain elements related to individuals (saints or blessed individuals).

lib/locales/fr.ts
emagnier: I guess "mr" stands for "Missale Romanum" ? I wasn't sure, but ok to use this abbreviation 👍👍

scripts/locale-checks.ts
emagnier: At this stage, you can pass whatever you want. Also this script is accessing internal Romcal components, which should not be accessed directly by end users utilizing Romcal. Here for some checks, I don't really need a locale data, so I'm creating a mock object when a locale is required.

scripts/locale-checks.ts
emagnier: 👍 I prefer to keep the trailing dot everywhere.

scripts/locale-checks.ts
emagnier: FYI, we're using an internal Romcal component here. The GRC is already integrated into the calendar computation, so there's no need to reintroduce it at this step.

emagnier:
Hello @tukusejssirs!
Thank you very much for your feedbacks and code review 🙏 
I'm attempting to address your questions below 😄.

As a reminder, what triggers an error by this tool requires fixing (to pass the CI and merge a PR), while warnings are simply encouragements to address issues but cannot block the PR from being merged. 

I aim to proceed with these criteria and tools to enhance the contribution experience, simplify the review process using more interactive tools like this one, and improve quality more simply (without striving for perfection and without necessarily creating large locales/calendars PRs that can't be completed quickly). We should shift away from creating intimidating big to-do lists every time a new contributor opens a PR and instead encourage small PRs or step-by-step contributions, depending on availability and motivation. This way, we can focus solely on what is localized (checking or challenging translations, verifying with sources, or leaving comments in the locale file about what still needs verification later - which you do very well).

- Proposed Tests:
  - **_Checking if only keys defined in `en` are defined in all other locales:_** Partially. All keys for celebrations and most associated metadata returned by Romcal must be localized in English (otherwise, it generates an error). For other locales, only celebrations from the Proper of Time and the GRC return a warning.
  - **_Checking if only keys defined in `en` are used in all calendar files:_** No. It is not required to translate everything into another language. However, missing locales for the Proper of Time and the GRD will be indicated as warnings. FYI, I plan to add more tools shortly to enable verification for any calendar and any locale, and to encourage and assist in prioritizing what needs to be translated in a specific locale.
  - **_Checking if all `en` localization keys (or parts of them) are defined in `martyrology.ts`:_** Partially. Locales explicitly used for an item in the martyrology are currently not utilized by Romcal (they do not appear in the metadata of a celebration). However, I consider this a missing feature. So, for now, it is already possible to localize each entity of the martyrology. This is a point I will evolve into a warning in a next PR, if an item is not localized (following the same criteria described above), and possibly later into an error when these translated metadata are returned by Romcal.

- **_Creating a separate script, instead of using Jest unit tests._** In fact, this verification script is not related to unit tests. It does not test specific Romcal features but rather checks the overall data integrity. This CLI is a first version that I intend to evolve (as mentioned at the end of the PR description) and possibly develop into an eslint plugin (with the possibility of automatically fixing certain points).

- **_Suggestion to consider renaming the `locale-checks` script for potential future check types like `checks:locale` or `checks:l10n`._** Thank you for the suggestion; I'll consider it based on how this tool evolves. In the short term, I'm leaning towards renaming it to `data:checks`.

- **_Including more saints/blesseds/venerables... in `martyrology.ts` even if not used by Romcal_**: This is actually out of scope, and I prefer to stick to what is strictly used by Romcal (i.e. defined from a calendar definition). If necessary, it's straightforward to extend Romcal's metadata through an external project or repository.

- **_Useless localized items: suggestion to flag for removal only those items not inscribed in `martyrology.ts`_**: The same comment as above applies. Anything not used by Romcal should not be included, so it should be removed.

lib/catalog/martyrology.ts
tukusejssirs: Consider this a request not blocking this PR from merging: I still want see event celebrations meatadata included in romcal, however, I don’t if they are moved into a separate file (e.g. `events.ts`).

lib/locales/fr.ts
tukusejssirs: Yeah, `mr` stands for _Missale Romanum_.

However, it could be ambiguous with `Martyrologium Romanum`, which I would shorten to `mar` (as in _MArtyrologium Romanum_); then we could have `mir` for _MIssale Romanum_, however, I still like `mr` better. YMMV.

We could also append `e` to the edition which would denote _emmendata_ (e.g. `mr_en_2011_ed3e`. For English Missals (which have two separate linguistic versions), we could also append `us` or `uk` (e.g. `mr_en_2011_ed3e_us`) in order to make it less ambiguous, because I presume some variations between the linguistic versions.

scripts/locale-checks.ts
tukusejssirs: You seem like not wanting to use an `enum` here. Any reaons? :pray:

tukusejssirs:
> leaving comments in the locale file about what still needs verification later - which you do very well).

While I did that in the past, I’m against using code comments for TODOs. I’d prefer creating an issue (per locale and per calendar) with the list of TODOs, potential issues and ideally with link to the related files.

>   - **_Checking if only keys defined in `en` are defined in all other locales:_** Partially. All keys for celebrations and most associated metadata returned by Romcal must be localized in English (otherwise, it generates an error). For other locales, only celebrations from the Proper of Time and the GRC return a warning.

Are you against adding this test (in full) and do you want to do it in a follow-up PR? IMHO this test totally makes sense.

We could create a dynamic type of `names` based on `en.ts` (all props optional, but no other prop is allowed) which would be then enforced in other l10n files; then this check would be done by TS compiler.

>   - **_Checking if only keys defined in `en` are used in all calendar files:_** No. It is not required to translate everything into another language. However, missing locales for the Proper of Time and the GRD will be indicated as warnings. FYI, I plan to add more tools shortly to enable verification for any calendar and any locale, and to encourage and assist in prioritizing what needs to be translated in a specific locale.

It seems to me like you are talking about a different thing.

I was talking about the l10n keys used in calendar files: all allowed keys are defined in `en.ts`, therefore, I suggest to check if all keys used in the GRC and all particular calendars are only those which are defined in `en.ts. Here I am not talking about _translat[ing] everything into another language_.

>   - **_Checking if all `en` localization keys (or parts of them) are defined in `martyrology.ts`:_** Partially. Locales explicitly used for an item in the martyrology are currently not utilized by Romcal (they do not appear in the metadata of a celebration). However, I consider this a missing feature. So, for now, it is already possible to localize each entity of the martyrology. This is a point I will evolve into a warning in a next PR, if an item is not localized (following the same criteria described above), and possibly later into an error when these translated metadata are returned by Romcal.

It sounds to me like you understood my suggestion the other way around. :wink: All I suggested is whether each `en.ts` key in `names` could be found in `martyrology.ts`.

However, on second thoughts, it might not be necessary after we restructure the files (possibly in YAML/JSON) with the following information per each l10n key (cf [this comment](https://github.com/romcal/romcal/discussions/445#discussioncomment-7287058)):

```ts
{
  /** Person/event ID from `martyrology.ts` or `events.ts` */
  id: string,
  /** The localised string */
  value: string,
  /** Reference of the source either where it can be found or based on which it was translated. */
  source: SourceReference
}

// Example: 
{
  id: 'epiphany_of_the_lord',
  value: 'The Epiphany of the Lord',
  source: {
    chapter: 'General Roman Calendar',
    page: 121,
    section: 'January',
    source: 'mr_en_2011_ed3e_us'
  }
}
```

> - **_Creating a separate script, instead of using Jest unit tests._** In fact, this verification script is not related to unit tests. It does not test specific Romcal features but rather checks the overall data integrity. This CLI is a first version that I intend to evolve (as mentioned at the end of the PR description) and possibly develop into an eslint plugin (with the possibility of automatically fixing certain points).

You’re right. :+1:

> - **_Suggestion to consider renaming the `locale-checks` script for potential future check types like `checks:locale` or `checks:l10n`._** Thank you for the suggestion; I'll consider it based on how this tool evolves. In the short term, I'm leaning towards renaming it to `data:checks`.

:+1:

> - **_Including more saints/blesseds/venerables... in `martyrology.ts` even if not used by Romcal_**: This is actually out of scope, and I prefer to stick to what is strictly used by Romcal (i.e. defined from a calendar definition). If necessary, it's straightforward to extend Romcal's metadata through an external project or repository.

Okay.

> - **_Useless localized items: suggestion to flag for removal only those items not inscribed in `martyrology.ts`_**: The same comment as above applies. Anything not used by Romcal should not be included, so it should be removed.

Okay.

tukusejssirs (approved):

lib/catalog/martyrology.ts
emagnier: You are mixing the entities from the martyrology and those from the calendar definitions. Currently, there are two calendar entities (`dedication_of_consecrated_churches_on_october_25` and `dedication_of_consecrated_churches_on_last_sunday_of_october`) that are well included in the data returned by Romcal. 

Just FYI, `dedication_of_consecrated_churches` corresponds only to the translation key defined from the `customLocaleId` property of the two definitions. So, here in this case (in the martyrology file), this entity is indeed unnecessary.

By the way, we already have an open issue https://github.com/romcal/romcal/issues/136, where I haven't had the chance to respond yet.

emagnier:
I've rephrased several responses, I hope it's clearer now 😉.
If not, please keep those questions coming. Also, don't hesitate to give a try to this new script added in this PR.

And thank you for the PR approval 👍 

- `I’m against using code comments for TODOs`: On the contrary, I strongly encourage this practice for everything related to data: calendar/locales/martyrology (however, I do have a different view when it comes to the codebase's TODOs). So, for example, if we don't have access to the sources or have the ability to quickly verify it, I highly encourage adding a TODO comment - as close to the data as possible - to avoid blocking the PR (as long as the contribution adds value). This makes it much easier for any future contributor who reviews the sources, to see what needs improvement or validation.
- `Are you against adding this test (in full) and do you want to do it in a follow-up PR?`: As I've mentioned before, I'm certainly not _against_ to adding these additional tests, and I plan to add them in a future PR.
- `We could create a dynamic type of names based on en.ts (all props optional, but no other prop is allowed) which would be then enforced in other l10n files; then this check would be done by TS compiler.`: I'm not particularly in favor of that. It will complicate the build process, and since we want to use agnostic data files, it doesn't really make sense.
- `all allowed keys are defined in en.ts, therefore, I suggest to check if all keys used in the GRC and all particular calendars are only those which are defined in en.ts`: The calendars are the sources of input data, so all locale keys come from there (the en.ts file is not the reference). And the check "_[4] There are useless localized items_" verifies that all localization keys correspond to a calendar entity (or its `customLocaleId` property when it apply) or a martyrology entity.
- `All I suggested is whether each en.ts key in names could be found in martyrology.ts`: No, because for example, a localization key can also correspond to a calendar entity for multiple persons at once (via the `customLocaleId` property). And the martyrology should only contain entities for one person at a time, as the martyrology is used to provide additional metadata specific to an individual (titles, birth or death dates, etc.).

Issue #457, opened 2023-09-19
Locale definitions are incompatible with Webpack 5


ciesiolka:
Using Localization package into the TypeScript project with Webpack 5 fails with the following errors:
```
ERROR in ./node_modules/@romcal/calendar.poland/esm/index.mjs 25:0-38
Module not found: Error: Can't resolve './cs' in 'C:\Users\Mat\Documents\spiewnik-organisty\node_modules\@romcal\calendar.poland\esm'
Did you mean 'cs.mjs'?
BREAKING CHANGE: The request './cs' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains 
'"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```
The same error is repeated 12 times, one for each language.

Installed with `npm i romcal@dev @romcal/calendar.general-roman@dev @romcal/calendar.poland@dev`.

tukusejssirs:
Could you give a MVE? This seems like you try to use ESM modules (`*.mjs` files) instead of CommonJS files (`*.js` files).

Look at these files:

```
node_modules/@romcal/calendar.general-roman/cjs/pl.js
node_modules/@romcal/calendar.general-roman/esm/pl.mjs
node_modules/@romcal/calendar.poland/cjs/pl.js
node_modules/@romcal/calendar.poland/esm/pl.mjs
node_modules/romcal/dist/cjs/romcal.js
node_modules/romcal/dist/esm/romcal.mjs
```

It might be related to you TS/Webpack configuration.

ciesiolka:
Webpack prioritizes mjs over js by default (see https://github.com/webpack/webpack/issues/5756) and in Webpack 5 a breaking change was introduced which makes the `.mjs` extension mandatory, unless the whole project is marked as module (see similar issue https://github.com/babel/babel/issues/12169).

I'll try to prepare MVE tomorrow.

TobiTenno:
@ciesiolka this should theoretically be fixed now, with some of the explicit exports added in dev.57. I'd love to collab with you if you need any help with an extra set of eyes

TobiTenno:
I can confirm this is working on at latest dev.60, if not earlier, as we're webpack 5 using on romcal-examples/react-app (see romcal/romcal-examples#2)

TobiTenno:
[...]

PR #465, opened 2023-10-20
Named exports for Romcal, and update hybrid NPM package setup (cjs, esm)

emagnier:
### Breaking Change:

- Now exporting Romcal (and other internal modules) as a named export. 
  - To import Romcal, you now should use:
      - `import { Romcal } from 'romcal'` instead of `import Romcal from 'romcal'`.
      - `const { Romcal } = require('romcal')` instead of `const Romcal = require('romcal')`.
  - This change reinforces the use of named imports, helps reduce potential issues with default exports (as some TypeScript compilers and bundlers do not work well with default exports), and will also make future code refactoring easier.
- Set Node 18+ as minimum version, and updating some npm dependencies.
- Updating hybrid NPM package setup (CJS + ESM), following the new recommended way by NPM.
  - This should address the issue described here: https://github.com/romcal/romcal/issues/457.


emagnier:
While awaiting the automation of the CHANGELOG, I have also added a note to document the breaking change (I will update the matching documented Romcal version, if needed).

lib/index.ts
```diff
@@ -329,7 +329,7 @@ class Romcal {
   static LOCALE_IDS = LOCALE_IDS;
 }
 
-export default Romcal;
+export { Romcal };
```
tukusejssirs, line 332: @emagnier, why don’t we simply export `Romcal` at its definition (L113) using `export class Romcal ...`? :thinking:

```suggestion
```

tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "target": "es2017",
-    "lib": ["ES2018", "ES2019", "ES2020", "ESNext"],
+    "module": "CommonJS",
+    "target": "ES2021",
```
tukusejssirs: In `scripts/build.ts`, you use `ES2022` L214 and on L240, but here you use `ES2021`. Don’t you want to use the same target? :thinking:

tsconfig.release.json
```diff
@@ -1,9 +1,8 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "module": "es2020",
-    "target": "es2020",
-    "lib": ["ES2020", "ESNext"],
+    "module": "ESNext",
+    "target": "ES2021",
```
tukusejssirs: In `scripts/build.ts`, you use `ES2022` L214 and on L240, but here you use `ES2021`. Don’t you want to use the same target? :thinking:

tukusejssirs (changes requested):
@emagnier, I’ve got just a few nitpicks, otherwise your work is awesome as always. :wink: :pray:

tsconfig.json
emagnier: Fixed 👍

tsconfig.release.json
emagnier: Fixed 👍

lib/index.ts
emagnier: Technically, it's the same thing. Here, it's just a matter of readability convenience: all the `export` are gathered in one place in this file.

tukusejssirs: I agree with you it is the same, however, for my point of view it is a useless line and extra 17 characters (` default Romcal;` plus a newline). I personally prefer exporting stuff where they are defined instead of exporting them separately.

As for the _readability convenience_, I personally see no advantage, as:

- IMHO developers use IDEs which use intellisense to gather all the info and has many _Go to_ features;
- even for other users, I believe that searching a file for `export` word is not that hard.

YMMV. If you disagree with me, feel free to merge the PR.

tukusejssirs (approved):

lib/index.ts
emagnier: I agree with your point of view, except for this specific case, for the reasons I've just mentioned. I prefer not to overlook the developer experience, rather than save characters (for code that is transpiled anyway).

PR #466, opened 2023-10-20
Checking and cleaning France calendar entries

emagnier:
- Validating calendar from the missal.
- Adding missing celebrations, or updating celebrations from the GRC.

<details>
  <summary>Sources (from <code>mr_fr_2021_ed3</code>)</summary>
  
</details>

### DIFF
--- a/lib/particular-calendars/france.ts
+++ b/lib/particular-calendars/france.ts
@@ -4,7 +4,6 @@ import { CalendarDef } from '../models/calendar-def';
 import { Inputs, ParticularConfig } from '../types/calendar-def';
 import { Europe } from './europe';
 
-// src: mr_fr_2021_ed3
 export class France extends CalendarDef {
   parentCalendar = Europe;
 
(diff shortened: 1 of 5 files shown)


lib/particular-calendars/france.ts
```diff
@@ -4,6 +4,7 @@ import { CalendarDef } from '../models/calendar-def';
 import { Inputs, ParticularConfig } from '../types/calendar-def';
 import { Europe } from './europe';
 
+// src: mr_fr_2021_ed3
```
tukusejssirs: I suggest to provide the sources of individual celebrations (or `LiturgicalDay` item if you wish). Providing a source like is ambiguous.

tukusejssirs (approved):
I have just a nitpick. I wish you’d update the PR to provide a source per celebration (`LiturgicalDay`), however, that is not that pressing and thus it is a non-blocking request.

lib/particular-calendars/france.ts
emagnier: Indeed, I agree.

tukusejssirs (approved):

PR #468, opened 2023-10-21
Re-introduce the OPTIONAL_MEMORIAL rank

### LINKED ISSUE #435: Add `optional memorial` rank
I have just noticed that there is no optional memorial in l10n files.

https://github.com/romcal/romcal/blob/5d3fc9c7be367bbb679bbcf63479a4db5e461241/lib/locales/en.ts#L54-L60

While the rank of a celebration in calendar files is defined using `Precedences.OptionalMemorial_12`, but in the output of romcal we need get `MEMORIAL` rank, `OPTIONAL_MEMORIAL_12` precedence, `isOptional` set to `true` and `rankName` is set to `memorial`.

```js
// Example
{
    id: 'most_holy_name_of_jesus',
    date: '2023-01-03',
    dateDef: { month: 1, date: 3 },
    precedence: 'OPTIONAL_MEMORIAL_12',
    rank: 'MEMORIAL',
    allowSimilarRankItems: false,
    isHolyDayOfObligation: false,
    isOptional: true,
    ...
}
```

Now, I think we should treat optional memorial rank as a different one from obligatory memorial, at the very least from the l10n point of view, as we should return different `rank` string.

While I get it that `isOptional` makes sense from one point of view, it does not make sense from l10n point of view. Actually, I consider this a bug, as _at the very least_ `rankName` should return `optional memorial` instead of `memorial`.

emagnier:
**Reintroduction of the `OPTIONAL_MEMORIAL` rank to:**
- Distinguish obligatory memorials from optional memorials using the `rank` property.
- It is still possible to distinguish them using the `precedence` or `isOptional` properties.
- Provide specific localization for the `OPTIONAL_MEMORIAL` value.

**Notes:**
- Up until Romcal v2, this value was titled `OPT_MEMORIAL`. I prefer to state it without abbreviation for clarity.
- The translations here come from the git history of this project.
- All the `rank` translations are now all in lowercase.

Resolves https://github.com/romcal/romcal/issues/435.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,15 @@ The romcal v3 is a major rewrite of the core library, with performance and archi
 As a result, romcal v3 is now up to 10x faster than v2 and v1.
 The new architecture helped to introduce new features without performance and code quality impacts.
 
-### Added
+### v 3.0.0@dev-52
+
+#### Breaking change
+
+- The `OPTIONAL_MEMORIAL` `rank` is being reintroduced, which was previously available as `OPT_MEMORIAL` until Romcal v2. This re-introduction includes related localizations for all currently supported locales.
+
+### v 3.0.0@dev-1
+
+#### Added
 
 - Locales & Internationalization:
   - introducing a new internationalization library, [i18next](https://www.i18next.com/).
@@ -29,7 +37,7 @@ The new architecture helped to introduce new features without performance and co
   - When computing a calendar, the corresponding metadata are also integrated in the final `LiturgicalDay` object.
   - This is just a start, new metadata could be added or refined.
 
-### Changed
+#### Changed
 
 - Build: the build process have been completely rewritten.
   - The core library do not contain anymore all the calendar and localization files, to lightweight the size of the library (it only contain a lightweight version of the General Roman Calendar, without translation and extra metadata).
@@ -58,7 +66,7 @@ The new architecture helped to introduce new features without performance and co
   - `seasons` are now following closely the _General Norms for the Liturgical Year and the Calendar_. There is no more a Holy Week season (moved to the period metadata), Holy Thursday have 2 possible liturgical days (the Holy Thursday which is the last day of Lent, and the Thursday of the Lord's Supper which start the Paschal Triduum), Easter Sunday take place in 2 seasons (the Paschal Triduum and the Easter Time).
   - `periods` include additional information, like early/late ordinary time, and special liturgical periods coming from the monastic traditions.
 
-### Removed
+#### Removed
 
 - Options to group and filters `LiturgicalDay` objects by predefined criteria is now removed. The reason is that is now very easy to group or filter a JavaScript object, and hard to cover all possible user requirement. However, produced data is by default sorted and grouped by dates.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The new architecture helped to introduce new features without performance and co
 
 #### Breaking change
 
-- The `OPTIONAL_MEMORIAL` `rank` is being reintroduced, which was previously available as `OPT_MEMORIAL` until Romcal v2. This re-introduction includes related localizations for all currently supported locales.
+- The `OPTIONAL_MEMORIAL` rank is being reintroduced, which was previously available as `OPT_MEMORIAL` until Romcal v2. This re-introduction includes related localizations for all currently supported locales.
 
 ### v 3.0.0@dev-1
 
--- a/lib/constants/ranks.ts
+++ b/lib/constants/ranks.ts
@@ -31,25 +31,25 @@ export enum Ranks {
   Feast = 'FEAST',
 
   /**
-   * Obligatory Memorials are liturgical commemorations of saints, events, or aspects of the faith.
-   * Their observance is mandatory and integrated into the celebration of the occurring weekday,
-   * following the liturgical norms outlined in the General Instruction of the Roman Missal and the
-   * Liturgy of the Hours.
-   * When an Obligatory Memorial falls on a weekday during the liturgical season of Lent or a
-   * privileged weekday of Advent, it must only be celebrated as an Optional Memorial, as Lent and
-   * Advent have their own specific liturgical observances that take precedence.
+   * **Obligatory memorials** are liturgical commemorations of saints, events, or aspects of the
+   * faith. Their observance is mandatory and integrated into the celebration of the occurring
+   * weekday, following the liturgical norms outlined in the General Instruction of the Roman Missal
+   * and the Liturgy of the Hours.
+   * When an **obligatory memorial** falls on a weekday during the liturgical season of Lent or a
+   * privileged weekday of Advent, it must only be celebrated as an **optional memorial**, as Lent
+   * and Advent have their own specific liturgical observances that take precedence.
    */
   Memorial = 'MEMORIAL',
 
   /**
-   * Optional Memorials are liturgical commemorations of saints, events, or aspects of the faith,
-   * but they are not obligatory.
+   * **Optional memorials** are liturgical commemorations of saints, events, or aspects of the
+   * faith, but they are not obligatory.
    * Their observance is integrated into the celebration of the occurring weekday, adhering to the
    * liturgical norms provided in the General Instruction of the Roman Missal and the Liturgy of
    * the Hours.
-   * In cases where multiple Optional Memorials are designated on the same day in the liturgical
+   * In cases where multiple **optional memorials** are designated on the same day in the liturgical
    * calendar, only one of them may be celebrated, and the others must be omitted (UNLY #14).
-   * This allows for some flexibility in choosing which Optional Memorial to commemorate when
+   * This allows for some flexibility in choosing which optional memorial to commemorate when
    * multiple options are available.
    */
   OptionalMemorial = 'OPTIONAL_MEMORIAL',
--- a/lib/constants/ranks.ts
+++ b/lib/constants/ranks.ts
@@ -31,15 +31,28 @@ export enum Ranks {
   Feast = 'FEAST',
 
   /**
-   * Memorials are either obligatory or optional; their observance is integrated into
-   * the celebration of the occurring weekday in accordance with the norms set forth in the
-   * General Instruction of the Roman Missal and of the Liturgy of the Hours.
-   * Obligatory Memorials which fall on weekdays of Lent may only be celebrated as
-   * Optional Memorials.
-   * If several Optional Memorials are inscribed in the Calendar on the same day, only
-   * one may be celebrated, the others being omitted. (UNLY #14)
+   * Obligatory Memorials are liturgical commemorations of saints, events, or aspects of the faith.
+   * Their observance is mandatory and integrated into the celebration of the occurring weekday,
+   * following the liturgical norms outlined in the General Instruction of the Roman Missal and the
+   * Liturgy of the Hours.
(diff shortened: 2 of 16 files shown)


tukusejssirs:
> All the rank translations are now all in lowercase.

While I agree with you, @emagnier, I’d like to emphasise that this should follow grammatical rules of a given locale, e.g. in German (not yet supported by romcal), all nouns are written with capital initial letter, therefore we should not assume the ranks has to be written in lowercase. That said, we should prefer lowercase to uppercase.

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
 
   /**
-   * Memorials are either obligatory or optional; their observance is integrated into
-   * the celebration of the occurring weekday in accordance with the norms set forth in the
-   * General Instruction of the Roman Missal and of the Liturgy of the Hours.
-   * Obligatory Memorials which fall on weekdays of Lent may only be celebrated as
-   * Optional Memorials.
-   * If several Optional Memorials are inscribed in the Calendar on the same day, only
-   * one may be celebrated, the others being omitted. (UNLY #14)
+   * Obligatory Memorials are liturgical commemorations of saints, events, or aspects of the faith.
```
tukusejssirs: In general, `obligatory memorial` (and any other celebration rank) is not a proper noun in English, therefore, it should not be written with capital initial letters, unless it is part of a title (heading, celebration name, etc).

We might want to move this to a separate PR, as this should be fixed in all files and I have no problem with doing so. Therefore, consider this comment as non-blocking. :wink:

```suggestion
   * Obligatory memorials are liturgical commemorations of saints, events, or aspects of the faith.
```

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
+   * Their observance is mandatory and integrated into the celebration of the occurring weekday,
+   * following the liturgical norms outlined in the General Instruction of the Roman Missal and the
+   * Liturgy of the Hours.
+   * When an Obligatory Memorial falls on a weekday during the liturgical season of Lent or a
+   * privileged weekday of Advent, it must only be celebrated as an Optional Memorial, as Lent and
```
tukusejssirs: Aren’t memorials in Lent and Advent (after 17 Dec including) celebrated rather as _commemorations_?

And again, `obligatory memorial` is not a proper noun in English _per se_.

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
-   * General Instruction of the Roman Missal and of the Liturgy of the Hours.
-   * Obligatory Memorials which fall on weekdays of Lent may only be celebrated as
-   * Optional Memorials.
-   * If several Optional Memorials are inscribed in the Calendar on the same day, only
-   * one may be celebrated, the others being omitted. (UNLY #14)
+   * Obligatory Memorials are liturgical commemorations of saints, events, or aspects of the faith.
+   * Their observance is mandatory and integrated into the celebration of the occurring weekday,
+   * following the liturgical norms outlined in the General Instruction of the Roman Missal and the
+   * Liturgy of the Hours.
+   * When an Obligatory Memorial falls on a weekday during the liturgical season of Lent or a
```
tukusejssirs:
```suggestion
   * When an obligatory memorial falls on a weekday during the liturgical season of Lent or a
```

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
+   * privileged weekday of Advent, it must only be celebrated as an Optional Memorial, as Lent and
+   * Advent have their own specific liturgical observances that take precedence.
    */
   Memorial = 'MEMORIAL',
 
+  /**
+   * Optional Memorials are liturgical commemorations of saints, events, or aspects of the faith,
```
tukusejssirs:
```suggestion
   * Optional memorials are liturgical commemorations of saints, events, or aspects of the faith,
```

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
+   * but they are not obligatory.
+   * Their observance is integrated into the celebration of the occurring weekday, adhering to the
+   * liturgical norms provided in the General Instruction of the Roman Missal and the Liturgy of
+   * the Hours.
+   * In cases where multiple Optional Memorials are designated on the same day in the liturgical
```
tukusejssirs:
```suggestion
   * In cases where multiple optional memorials are designated on the same day in the liturgical
```

lib/constants/ranks.ts
```diff
@@ -31,16 +31,29 @@ export enum Ranks {
...
+   * calendar, only one of them may be celebrated, and the others must be omitted (UNLY #14).
+   * This allows for some flexibility in choosing which Optional Memorial to commemorate when
```
tukusejssirs:
```suggestion
   * This allows for some flexibility in choosing which optional memorial to commemorate when
```

emagnier:
If the rule in a language is to keep these names capitalized, I am, of course, 100% in favor of following that rule (for example, I left "Sunday" capitalized in English). Otherwise, indeed, I also think it's better to keep them in lowercase as it makes them easier to use in string composition.

lib/types/locale.ts
```diff
@@ -52,6 +52,7 @@ export interface Locale {
     sunday?: string;
     feast?: string;
     memorial?: string;
+    optional_memorial?: string;
     weekday?: string;
   };
```
tukusejssirs, line 57: How about making `ranks`, `colors`, `weekdays` and `months` required? IMHO, they can be easily translated in any locale, as these strings are very common.

tukusejssirs (approved):
@emagnier, thanks for the PR. It LGTM, I have just some (regular) nitpicks. :wink:

If you want, you merge the PR as is.

lib/constants/ranks.ts
emagnier: Would you think that in these cases (during Lent and from 17 Dec), we should define a new `rank` type as `COMMEMORATION`? At first glance, I think it's a good idea  👍  Especially since the precedence property also provides useful information about the celebration.

lib/types/locale.ts
emagnier: No, because an XX-XX locale may extend an XX locale. So, at this level, it should remain optional (the script added here #461 will, however, ensure that all these items are correctly defined and present for a language).

lib/constants/ranks.ts
tukusejssirs: Well, defining a new `rank` for commemorations is a good idea, however, I am not sure if that is that necessary (there is no such type in the GIRM). IMHO adding a `LiturgicalDay` (celebration) property (e.g. `isCommemoration` of a `boolean` type) could suffice.

lib/types/locale.ts
tukusejssirs: You’re right, I have forgotten about the regional locales. :+1:

lib/constants/ranks.ts
emagnier: I like the convention of capitalizing these names to emphasize the importance of these names, which are the subject of the description here. This writing rule should only be used in such cases (for describing a property or an acronym, for example).

However, at the same time, and for the sake of consistency, I understand your point and your preference for following the general writing rules.

Before choosing one writing style over the other, what are your thoughts on this?

tukusejssirs: > to emphasize the importance of these names

Based on what you what to accomplish, I’d choose one of the following:
- reference a type: use [TSDoc links](https://tsdoc.org/pages/tags/link/). e.g. `{@link Ranks.OptionalMemorial}` (once per TSDoc block);
- emphasise some strings: use Markdown syntax for bold and/or italics;
- emphasise some code: use Markdown syntax for inline code snippets.

I’d always use Markdown syntax in TSDocs anyway. I’d also try to use grammatically correct code comments in simple English.

> acronym

I think we need to create a glossary or a list of terms where we would define some of the most commonly used stuff, including acronyms and shortenings (which would link to an entry with full name via `see $link`). Then, we could use `{@links $link}` in TSDocs when needed.

emagnier: Okay, and thank you for your recommendations. I will update these descriptions following your suggestions 👍

emagnier: > I think we need to create a glossary or a list of terms where we would define some of the most commonly used stuff

Indeed, this is clearly something I would like to see added to Romcal.

lib/constants/ranks.ts
emagnier: I'm not very enthusiastic about adding an `isCommemoration` property. 
Especially since memories, feasts, and other ranks are also commemorations, having an `isCommemoration: false` property for these celebrations seems counterintuitive to me. 

The term "commemoration" (for a rank) remains quite generic, so in our case here, it seemed appropriate (i.e. we specify that this element remains a commemoration, even though it no longer holds the official rank of a memory).

Let's save this discussion for later and for a dedicated topic because I believe we need to reexamine the missal instructions on this issue and let this idea mature a bit.

tukusejssirs: > I'm not very enthusiastic about adding an `isCommemoration` property.
> Especially since memories, feasts, and other ranks are also commemorations, having an `isCommemoration: false` property for these celebrations seems counterintuitive to me.

You might be right.

I was referring to this section of the GIRM:

> 

On the weekdays of Advent from December 17 to December 24, on days within the Octave of Christmas, and on the weekdays of Lent, except Ash Wednesday and during Holy Week, the Mass for the current liturgical day is to be used; but the collect may be taken from a memorial which happens to be listed in the General Calendar for that day, except on Ash Wednesday and during Holy Week. On weekdays of the Easter Season, memorials of Saints may rightly be celebrated fully.

Whenever a proper celebration is prohibited by this and only the _collect may be take from a memorial_, I considered this a _commemoration_. I have no idea where I found this term though.

> Let's save this discussion for later and for a dedicated topic because I believe we need to reexamine the missal instructions on this issue and let this idea mature a bit.

Agreed. :+1:

tukusejssirs (approved):

emagnier:
@tukusejssirs, FYI I have rebased the PR, and added a commit to document the breaking change in the changelog (while waiting for the changelog update to be automated someday).

CHANGELOG.md
```diff
@@ -8,7 +8,15 @@ The romcal v3 is a major rewrite of the core library, with performance and archi
 As a result, romcal v3 is now up to 10x faster than v2 and v1.
 The new architecture helped to introduce new features without performance and code quality impacts.
 
-### Added
+### v 3.0.0@dev-52
+
+#### Breaking change
+
+- The `OPTIONAL_MEMORIAL` `rank` is being reintroduced, which was previously available as `OPT_MEMORIAL` until Romcal v2. This re-introduction includes related localizations for all currently supported locales.
```
tukusejssirs: I probably use `rank` here as a term not as a code. In case you still want to use it as a code, IMHO it should be written as `Rank`.

```suggestion
- The `OPTIONAL_MEMORIAL` rank is being reintroduced, which was previously available as `OPT_MEMORIAL` until Romcal v2. This re-introduction includes related localizations for all currently supported locales.
```

tukusejssirs (approved):
I have a minor nitpick, however, it is nothing serious, therefore I approve the PR in case you want to merge it anyway. :wink: :pray:

CHANGELOG.md
emagnier: `rank` was meant to refer to the `rank` property, not the `Rank` type. But alright, I'm going with your suggestion 👍👍. Anyway, you have a great eye for detail (which is excellent), but let's merge this PR now and continue to make progress on other matters.

PR #471, opened 2023-12-03
fix: add export keys, file path for esm files

TobiTenno:
- fixes error found on romcal/romcal-examples#1:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'rest-api-with-fastify/node_modules/@romcal/calendar.france/esm/cs' imported from @romcal/calendar.france/esm/index.js
```
- export keys are added to help resolve an error where node resolution thinks romcal is only CJS on node 18 (at least according to the config in the example app)

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/hydrogen
--- a/package.json
+++ b/package.json
@@ -8,6 +8,11 @@
   "engines": {
     "node": ">=18.0.0"
   },
+  "exports": {
+    "types": "./dist/index.d.ts",
+    "import": "./dist/esm/romcal.js",
+    "require": "./dist/cjs/romcal.js"
+  },
   "scripts": {
     "clean": "rimraf coverage dist tmp",
     "lint": "eslint . --ext .ts",
--- a/scripts/bundle.ts
+++ b/scripts/bundle.ts
@@ -244,7 +244,7 @@ export const RomcalBundler = (): void => {
      */
     const indexImports = Object.entries(calVarObj).reduce((acc, [locale, varName]) => {
       const importFileName = locale.replace(/([A-Z])/g, '-$1').toLowerCase();
-      return acc + `import { ${varName} } from './${importFileName}';\n`;
+      return acc + `import { ${varName} } from './${importFileName}.js';\n`;
     }, '');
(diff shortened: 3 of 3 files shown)


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/hydrogen
```
emagnier, line 1: Could you exclude this file from your PR (or/and add it to `.gitignore`)?

That being said, in a separate PR, it might be a good idea to reintroduce it, while modifying the Github Actions files of this project to also use this same configuration file.

emagnier (changes requested):
Thank you for your contribution @TobiTenno, and welcome to this project! 

Indeed, ESM modules were still the neglected part of this project (I hadn't had the time to thoroughly test it yet). So, thank you very much for the fix 🙏 

Just one small thing to fix for this PR, and we'll be good to merge!

.nvmrc
TobiTenno: I'll happily do so. I think it's important to have since it provides (often near-automatic) setup information before trying to install.

emagnier: You're right, it's a free improvement. Let's keep it here.

In the future, using this config file into the Github Actions files would be a valuable addition (contributions are welcome for an upcoming PR).

emagnier (approved):

.nvmrc
TobiTenno: for sure! I'll do so posthaste

TobiTenno: @emagnier would the preference be to use the version that volta uses as well? it looks like volta pins to 20.8.1 right now. should it be pinning to an LTS?

emagnier: Good point. An LTS version is probably preferable.

TobiTenno: cool. i'll make sure i can pin them both. would it be problematic to _just_ use the nvmrc and not the volta config in the package.json?

i think volta supports .nvmrc (i'll submit a PR and i guess we can see?) or i can remove the .nvmrc and put a readme entry about using volta, then i can switch the CI build to use volta, if that's the desired source of truth.

emagnier: actually, I would lean towards droping Volta and relying solely on `.nvmrc` as the source of truth. I haven't been using Volta for some time now (it doesn't support PNPM, which I've been using more and more across various projects), and I believe NVM is much more mature and widely used than Volta.

PR #476, opened 2024-02-21
Feat/fix es string

miguelms95:
Fix the string "modaro" (word that doesn't exist) to "morado" https://dle.rae.es/morado

### DIFF
--- a/lib/locales/es.ts
+++ b/lib/locales/es.ts
@@ -61,13 +61,13 @@ export const locale: Locale = {
   },
 
   colors: {
-    black: 'negro',
-    gold: 'dorado',
-    green: 'verde',
-    purple: 'morado',
-    red: 'rojo',
-    rose: 'rosa',
-    white: 'blanco',
+    black: 'negro', // src: GIRM, 346e (https://www.santaclaradeestella.es/BIBLIOTECA/ENCICLICAS/INSTRUCCI%C3%93N_GENERAL_DEL_MISAL_ROMANO.htm)
+    gold: 'dorado', // src: GIRM, 346g (https://www.santaclaradeestella.es/BIBLIOTECA/ENCICLICAS/INSTRUCCI%C3%93N_GENERAL_DEL_MISAL_ROMANO.htm)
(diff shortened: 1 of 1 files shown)


lib/locales/es.ts
```diff
@@ -64,7 +64,7 @@ export const locale: Locale = {
     black: 'negro',
     gold: 'dorado',
     green: 'verde',
-    purple: 'modaro',
+    purple: 'morado',
```
tukusejssirs: It might be a good idea to add a source. I suggest to use GIRM where in 346d, there is defined the liturgical usage of the colour.

I suggest to use `// src: GIRM, 346d (https://www.santaclaradeestella.es/BIBLIOTECA/ENCICLICAS/INSTRUCCI%C3%93N_GENERAL_DEL_MISAL_ROMANO.htm)`, i.e. in the following format: `// src: GIRM, $paragraf[$point] ($url_in_a_locale)`.

You could do the same for the rest of the colours.

tukusejssirs (changes requested):
Thanks, @miguelms95!

While fixing the typo, I think we should also add the source for the colour, potentially for all the colours.

.vscode/settings.json
```diff
@@ -5,8 +5,8 @@
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true, // if developer has eslint extension enabled
-    "source.fixAll.tslint": false // if developer has tslint extension enabled
+    "source.fixAll.eslint": "explicit",
```
tukusejssirs: Could you remove the changes to this file please? 🙏

tukusejssirs (changes requested):
@miguelms95, thanks! :pray: LGTM, apart from some changes in the VSCode config. After you remove them, tgis PR id ready to get merged.

.vscode/settings.json
miguelms95: Sorry, it was an error. It's fixed now! 👍

tukusejssirs (approved):

tukusejssirs:
Thanks, @miguelms95, for your PRs! 🙏

PR #477, opened 2024-02-23
fix(l10n): hard-code seasonal week number in `.names` of every locale

### LINKED ISSUE #474: 21 January - spanish locale missing week number - third sunday ordinay time
Hi, 21 january, there is no translation or mapping for the third sunday of Ordinary time. Romcal returns {{week}}, see below:
image

i'm not sure where to update to solve this.

tukusejssirs:
Currently, `i18next` functions are not run on `.names` l10n object, however, for now, it is not necessary to do so, as the values for the related l10n items are always the same.

Fix #474


tests/locales.test.ts
TobiTenno, line 1: I like this, but we should consider moving that fixture array to a fixture file so it's cleaner

TobiTenno: https://github.com/tukusejssirs/romcal/pull/3 i submitted a PR to your fork with my recommendation

tests/locales.test.ts
```diff
@@ -2,6 +2,7 @@ import { Brazil_PtBr } from 'romcal/dist/bundles/brazil';
 import { France_Fr } from 'romcal/dist/bundles/france';
 
 import { Romcal } from '../lib';
+import { fixture as nameFixtures } from './fixtures/names.fixture';
```
tukusejssirs: @TobiTenno, how about naming the fixture as `NamesFixture` instead of naming it `fixture` and using an alias here? :thinking:

TobiTenno: this makes sense to me. I find that I make heavy use of aliases so as to use the same name out of a lot of exports so that the consumer (in this case, the test) can alias as needed and we don't have to make up unnecessarily complex unique names in each file/implementation/etc, especially where the name of the file generally gives this differentiation

tests/fixtures/names.fixture.ts
```diff
@@ -0,0 +1,143 @@
...
+  /** Localization tests definition */
+  t: ({
+    /** Liturgical celebration localization key name */
+    k: string,
+    /** Expected value of a localized liturgical celebration name */
+    v: string,
+  })[]
+})[]
+
+export const fixture: Fixture = [
```
TobiTenno:
```suggestion
export const namesFixture: Fixture = [
```

tests/locales.test.ts
TobiTenno:
```suggestion
import { namesFixture } from './fixtures/names.fixture';
```

TobiTenno (changes requested):

tests/locales.test.ts
tukusejssirs: > we don't have to make up unnecessarily complex unique names in each file/implementation/etc, especially where the name of the file generally gives this differentiation

While I get what you mean, however, I prefer the filename to be based on the export name in kebab-case (without the suffix when it is part of the export name) with a suffix and extension (e.g. `namesFixture` → `names.fixture.ts`).

That said, I am open to define any export naming style, even one that I dislike (e.g. I don’t use semicolons unless necessary, however, Etienne uses them, thus they are used in romcal).

TobiTenno: oh, i meant more that, like you, i have tendencies wherein i'm willing to do something different from them

tests/fixtures/names.fixture.ts
tukusejssirs: I have renamed it to `seasonal-week-number-localisation.fixture.ts`. I know it is a quite long filename, however, it better describes its contents.

WDYT? :thinking:

TobiTenno: i think it's far too long for reusage. at longest, i'd be ok with shortening each word, otherwise i think it should just use a single one of the names that's most identifiable....

`seasonal-week-num-l10n.fixture.ts` at the longest

the name alone, as it is, would cause a line wrap, or just about

TobiTenno: alternatively, we can nest fixtures by concept in various folders, so... `l10n/season-weeks.fixture.ts` and the fixture could be `weekNumber: Fixture`

tukusejssirs: > alternatively, we can nest fixtures by concept in various folders, so... `l10n/season-weeks.fixture.ts` and the fixture could be `weekNumber: Fixture`

I like this compromise. Actually, I think it is even better (I think I have suggested this for the particular calendars: move then to separate folders by calendar level, like the GRC, continental, regional, national, dioceasan, parish calendars).

Now, we might want to think about the actual folder structure under `tests/fixtures`. I think that the first division is to be `l10n`/`locales` and `calendars` (and maybe `global`). Then, we might want to divide `l10n` into `temporal` and `sactoral`, or maybe `generated` (for generated stuff, like the localised seasonal week number) and `static` (like the original tests for a key in French and Brazilian Portuguese).

Should we have any more subdivisions? How should we divide the fixtures related to calendars? :thinking:

TobiTenno: from how i understand the temporal vs sactoral, those are purely for what the user chooses to request at construction time, those aren't separate files... from what you're saying, i'm probably mistaken, but i don't remember seeing separate calendars for those.

if we need tests for specific locales, i think something's going wrong, as those rules should be built into the general library logic to be tested for each of them based on an individual configuration

tukusejssirs: > from how i understand the temporal vs sactoral

I always forget if the final `e` is used in English or not (it is actually). :man_facepalming:

Now, [_Sanctorale_](https://en.wikipedia.org/wiki/Sanctorale) is the Proper of Saints, while [_Temporale_](https://en.wikipedia.org/wiki/Temporale) is the Proper of Time.

In romcal, the Proper of Time is fully generated (calendar-wise; unless there are some extra rules in a particular calendar which override those in the GRC), while the Proper of Saints is defined per calendar (as celebrations), while also counting on the calendar inheritance (e.g. `united-states` inherite celebrations from the GRC, then from `americas`, finally overridden by those in `united-states`) to reduce defining celebrations multiple times in the codebase.

I hope I have answered your implicit question. :wink:

> if we need tests for specific locales, i think something's going wrong, as those rules should be built into the general library logic to be tested for each of them based on an individual configuration

Well, I think that some stuff needs to be verified for all locales/calendars, others are more specific only to one/some of them. I might be wrong though, as it all depends on where are the _unit_ boundaries.

Moreover, we want to move to monorepo and make create the data files (l10n + calendars/celebrations + sources + martyrology) code-agnostic (and they should be written without any code in general), thus we might want to wait until then, as testing the data files as static files is quite different from the current ones.

Also, you are right that _those rules should be built into the general library logic_ and after moving to monorepo, IMHO we’ll move some of the logic to the core library (`@romcal/core`).

In general, I think we should work less on the tests and more on the move to monorepo.

I still wish to finish at least #459, probably #126 and maybe #478, however, the less time I have, the more I think we should start working on the monorepo stuff ASAP. On monorepo, I think we should talk either in #445 or maybe in a separate GH Discussion.

TobiTenno:
your changes failed lint checks

TobiTenno (approved):

tukusejssirs:
> your changes failed lint checks

Yeah, I am about to merge the fix. :wink:

tests/fixtures/names.fixture.ts
TobiTenno: yes, regardless, we should have a monorepo discussion, not in the PR for something barely tangentially related

TobiTenno:
planning to merge at <time datetime="2024-03-14T00:00:00+0000">midnight UTC tonight</time>

tukusejssirs:
You can merge this rightaway. :wink:

PR #480, opened 2024-03-10
fix(la): use correct gender for ordinal numbers

tukusejssirs:
Previously, we used ordinal numbers in masculine gender for Sundays in Ordinary Time, Lent and Easter Time.


tukusejssirs:
@TobiTenno, could you review this PR please? :pray:

tests/locales.test.ts
```diff
@@ -2,15 +2,42 @@ import { Brazil_PtBr } from 'romcal/dist/bundles/brazil';
...
+            if (day) {
+              /* eslint-disable-next-line jest/no-conditional-expect */
+              expect((day)?.name).toBe(d.v);
+            } else {
+              /* eslint-disable-next-line no-console */
+              console.log(`\`${i.c.i18n.id}\`: \`${d.k}\` is not celebrated in ${new Date().getUTCFullYear()}`)
+            }
+          });
+        }
+      }
```
TobiTenno: I'm realizing i don't think the single-character names of fixture fields is very readable.. it's probably too late to fix that, but if we could, that would be much appreciated. this is probably the downside of reviewing in a browser vs coding in an ide.

TobiTenno (approved):
i like the point of the change, the clarity of fixtures is testy, so i might submit a PR after this to converse/propose some more expanded names for fixtures variables

tukusejssirs:
> i like the point of the change, the clarity of fixtures is testy, so i might submit a PR after this to converse/propose some more expanded names for fixtures variables

Go for it! :wink:

tests/locales.test.ts
tukusejssirs: I have no problem renaming the single-character names. It is not too late for this, as it is not yet merged into `dev` and it will never be too late, as this `type` is used in tests only, thus we can change them any time, even after merging it to `dev`.

How about using this `type` definition:

```ts
type Fixture = ({
  /** Base romcal bundle instance */
  calendar: BaseRomcalBundle,
  /** Localization tests definition */
  tests: ({
    /** Liturgical celebration localization key name */
    key: string,
    /** Expected value of a localized liturgical celebration name */
    value: string,
  })[]
})[]
```

tukusejssirs: Look at [this](https://github.com/romcal/romcal/pull/477/files#diff-03515950f846611766a0962c19a6c41f320aeb8b07fbe8d90607d6f2869dfef7). I fixed it in #477 fixed, then I want to rebase this PR on top of it.

TobiTenno:
planning to merge at <time datetime="2024-03-14T00:00:00+0000">midnight UTC tonight</time>

tukusejssirs:
I want to rebase this after merging #477 so that we import the `Fixture` type, before merging this.

TobiTenno:
@tukusejssirs #477 is merged now, so feel free to rebase

tukusejssirs:
Nearly all tests fail for some reason, for example:

```bash
  ● Testing localization functionality › Test the Latin locale › Testing whether Latin celebration names use correct gender for ordinal numbers › - `la`: `ordinary_time_33_sunday`

    expect(received).toBe(expected) // Object.is equality

    Expected: "Dominica trigesima tertia per annum"
    Received: "Dominica trigesimum tertium per annum"

      25 |     describe('Testing whether Latin celebration names use correct gender for ordinal numbers', () => {
      26 |       laOrdinalNumberGenderFixture.forEach((i) => {
    > 27 |         const r = new Romcal({ localizedCalendar: i.calendar });
         |                                                       ^
      28 |
      29 |         i.tests.forEach((d) => {
      30 |           test(`- \`${i.calendar.i18n.id}\`: \`${d.key}\``, async () => {

      at Object.<anonymous> (tests/locales.test.ts:27:55)
```

AFAIK the following should work (i.e. `i18n` should use the feminine form → `33_feminine`):

https://github.com/tukusejssirs/romcal/blob/b6716778ef2006279e3283e95e9f7a0115677503/lib/locales/la.ts#L26

```ts
      sunday: 'Dominica $t(ordinals:{{week}}, { "context": "feminine" }) per annum',
```

https://github.com/tukusejssirs/romcal/blob/b6716778ef2006279e3283e95e9f7a0115677503/lib/locales/la.ts#L166

```ts
    '33_feminine': 'trigesima tertia',
```

tukusejssirs:
But the `la` tests pass in GitLab Actions. :thinking:

The test suit fails using either PNPM or NPM on my machine. All tests passed when I first created them few days ago.

TobiTenno:
yeah, i'm trying to repro it failing, as I'm not seeing failures on my end

tukusejssirs:
Okay, it seems like some caches caused all these issues. After ditching them all and starting from scratch, everything works as expected.

@TobiTenno, the PR is ready to be merged. Note that I have added another to fix some lint and format issues. I know it is out of the PR scope, so if you want to do it in a separate PR, just say so. :wink:

TobiTenno:
nah, i like fixing format & lint issues in progress, so this is great :+1:

tukusejssirs:
> nah, i like fixing format & lint issues in progress, so this is great 👍

But the changed files are unrelated to this PR. :man_shrugging:

TobiTenno:
yeah, but sometimes this happens when we update lint/prettier cause we don't have a format check in our CI... which i'm probably gonna change shortly

tukusejssirs:
> yeah, but sometimes this happens when we update lint/prettier cause we don't have a format check in our CI... which i'm probably gonna change shortly

I am for manual checking of the lint/format changes, as sometimes they mess some stuff up.

On the other hand, I agree to add lint/format _checks_ to the CI and when they fail, the PR checks would fail.

WDYT? :thinking:

TobiTenno:
technically we already have lint checks, as they're part of the prebuild, so `npm run build` triggers `npm run lint`.

the format checks (in #493) don't fix the problems, they just let you know about them.

tukusejssirs:
> technically we already have lint checks, as they're part of the prebuild, so npm run build triggers npm run lint.

Yeah, however, I disagree with having them there. IMHO we should be able to build (locally) an unlinted and unformatted code. IMHO `prebuild` should be used for build preparation (like cleanup or some files preparation), not for linting or formatting.

Lint and format checks could be part of the pre-commit checks (e.g. using Husky) though.

TobiTenno:
hmmm.... i'll do this pr first add them to ci, then i'll do a followup for setting up husky and remove them from prebuild. how does that sound?

tukusejssirs:
It sounds perfect to me, thanks! :pray:

PR #493, opened 2024-03-13
ci: run prettier checks in ci

TobiTenno:
lint runs as part of prebuild

i didn't add the prettier check to the prod release build because if it's already there, i'd rather let the thing succeed


tukusejssirs:
→ https://github.com/romcal/romcal/pull/480#issuecomment-1996010405

TobiTenno:
well those errors are odd

.github/workflows/romcal-ci.yml
```diff
@@ -24,6 +24,7 @@ jobs:
 
       # using -q to silent info outputs
       - run: npm -q --no-progress ci --no-audit
+      - run: npm run prettier
```
tukusejssirs:
```suggestion
      - run: npm run lint
      - run: npm run prettier
```

tukusejssirs (changes requested):
How about adding `lint` too? :thinking:

tukusejssirs:
> well those errors are odd

Why? :thinking:

.github/workflows/romcal-ci.yml
TobiTenno: > → [#480 (comment)](https://github.com/romcal/romcal/pull/480#issuecomment-1996010405)

TobiTenno:
> > well those errors are odd
> 
> Why? 🤔

i don't get them locally. I was referring to the ones in my actual build

tukusejssirs:
> i don't get them locally. I was referring to the ones in my actual build

I’m not sure I get you now.

Are you saying that you don’t errors _for this PR_ locally or in general you didn’t experienced such errors? :thinking:

I was talking about my previous experiences (not related to this PR) that the by linting and formatting the codebase _could_ make the code either unbuildable or cause some runtime errors. Granted, it happens seldom.

TobiTenno:
on this PR, i'm talking about just the errors i'm seeing on this PR. I get plenty of formatting errors if i don't format my code.

tukusejssirs:
Regarding [these errors](https://github.com/romcal/romcal/actions/runs/8272505932/job/22634450971?pr=493#step:5:12), i think it would be much safer to provide a list of globs (with file extentions) and a list of filenames which should be formatted.

TobiTenno:
> Regarding [these errors](https://github.com/romcal/romcal/actions/runs/8272505932/job/22634450971?pr=493#step:5:12), i think it would be much safer to provide a list of globs (with file extentions) and a list of filenames which should be formatted.

I'm not entirely sure why we don't just run it against everything. Given that the main source files should already be linted from the previous step, checking formatting would be "double" work, but would help us to simplify the prettier invocation.

tukusejssirs:
> I'm not entirely sure why we don't just run it against everything. Given that the main source files should already be linted from the previous step, checking formatting would be "double" work, but would help us to simplify the prettier invocation.

I usually lint and format only specific files (using globs), however, I don’t mind to lint and format them all. Do it if you want. :wink:

.github/workflows/romcal-ci.yml
```diff
@@ -23,10 +23,20 @@ jobs:
           node-version-file: '.nvmrc'
 
       # using -q to silent info outputs
-      - run: npm -q --no-progress ci --no-audit
+      - name: Install dependencies
```
tukusejssirs: Inconsistent capitalisation of job names: either use _Title Case_ or _Sentence case_ everywhere.

.github/workflows/romcal-ci.yml
```diff
@@ -23,10 +23,20 @@ jobs:
...
+        run: npm -q --no-progress ci --no-audit
 
-      - run: npm run build
+      - name: Lint
+        run: npm run lint
 
-      - run: npm run test:without-coverage
+      - name: Check Formatting
```
tukusejssirs: Inconsistent naming style: either change the name on L29 to `Check linting` (or something similar with `Check`) or naming this name (L32) to `Format`.

tukusejssirs (approved):
I’ve got some nitpicks. Feel free to disregard them if you don’t want to change the job names. :wink:

.github/workflows/romcal-ci.yml
TobiTenno: definitely a nitpick, but i'm open to it :)

.github/workflows/romcal-ci.yml
TobiTenno: the problem with a change like these is is that `Format` is connotative of "changing", whereas `Lint` is the generalized word for static analysis that doesn't change the code.

so: `Check Lint` is redundant, and `Format` is inaccurate, because we're doing read-only checking

tukusejssirs: I agree that `Format` is inaccurate and `Check Format` is better.

> whereas `Lint` is the generalized word for static analysis that doesn't change the code.

While you are right by definition, however, ESLint _can_ and will change the code using `--fix`. But that is not the case in here.

---

I still prefer using `Check` for both of them, but that is my preference. We don’t like lint or format the code in the CI, thus naming them `Lint` or `Formatting` is inaccurate too. However, I don’t want to argue with you or talk much about this, as it is not an important thing. :wink:

PR #495, opened 2024-03-14
ci: pre-commit lint, prettier, commitlint

TobiTenno:
- set up husky
- add lint & fmt checks to pre-commit
- add commit linting to commit msg hook
- remove prebuild lint script
- add commit linting to ci
- dedupe package-lock on commit

### DIFF
--- a/.commitlintrc
+++ b/.commitlintrc
@@ -1,3 +1,0 @@
-{
-  extends: ["@commitlint/config-conventional"]
-}
--- a/.commitlintrc
+++ b/.commitlintrc
@@ -1,0 +1,3 @@
+{
+  extends: ["@commitlint/config-conventional"]
+}
--- a/.commitlintrc.ts
+++ b/.commitlintrc.ts
@@ -1,0 +1,52 @@
+import { UserConfig } from '@commitlint/types';
+import { LOCALE_IDS } from './lib/constants/locales';
+import { CALENDAR_IDS } from './lib/constants/calendars';
+
+const empty = [null];
+const calendarScopes = [...CALENDAR_IDS, ...LOCALE_IDS, 'l10n', 'calendar'];
+
+const typesEnumScoped = {
+  // Updates to the automation or release process
+  ci: empty,
+  // Changes to build scripts
+  build: empty,
+  // Documentation only changes
+  docs: empty,
+  // Features, net-new functionality
+  feat: calendarScopes,
+  // Enhancing existing functionality
+  enh: calendarScopes,
+  // Bug fixes
+  fix: calendarScopes,
+  // Refactoring, no behavior changes
+  refactor: calendarScopes,
+  // Performance improvements
+  perf: calendarScopes,
+  // Add or correct tests
+  test: calendarScopes,
+  // Changes that affect
+  style: calendarScopes,
+};
+
+export default {
+  extends: ['@commitlint/config-conventional'],
+  formatter: '@commitlint/format',
+  helpUrl: 'https://github.com/romcal/romcal/blob/dev/docs/contribute-to-romcal.md#committing-changes',
+  parserPreset: {
+    parserOpts: {
+      issuePrefixes: ['#'],
+    },
+  },
+  plugins: ['selective-scope'],
+  rules: {
+    'scope-enum': [
+      2,
+      'always',
+      // From `commitlint` point of view, allow all scopes regardless of the type
+      [...calendarScopes].filter(Boolean),
+    ],
+    // Use `selective-scope` to make sure that the scopes are only allowed when they are defined for a particular type
+    'selective-scope': [2, 'always', typesEnumScoped],
+    'type-enum': [2, 'always', Object.keys(typesEnumScoped)],
+  },
+} as UserConfig;
--- a/.commitlintrc.ts
+++ b/.commitlintrc.ts
@@ -12,7 +12,7 @@ const typesEnumScoped = {
   build: empty,
   // Documentation only changes
   docs: empty,
-  // Features, net-new functionality
+  // Features, new functionality
   feat: calendarScopes,
   // Enhancing existing functionality
   enh: calendarScopes,
--- a/.commitlintrc.ts
+++ b/.commitlintrc.ts
@@ -3,7 +3,7 @@ import { LOCALE_IDS } from './lib/constants/locales';
 import { CALENDAR_IDS } from './lib/constants/calendars';
 
 const empty = [null];
-const calendarScopes = [...CALENDAR_IDS, ...LOCALE_IDS, 'l10n', 'calendar'];
+const scopes = [...CALENDAR_IDS, ...LOCALE_IDS, 'l10n', 'calendar'];
 
 const typesEnumScoped = {
   // Updates to the automation or release process
@@ -13,19 +13,19 @@ const typesEnumScoped = {
   // Documentation only changes
   docs: empty,
   // Features, new functionality
-  feat: calendarScopes,
+  feat: scopes,
   // Enhancing existing functionality
-  enh: calendarScopes,
+  enh: scopes,
   // Bug fixes
-  fix: calendarScopes,
+  fix: scopes,
   // Refactoring, no behavior changes
-  refactor: calendarScopes,
+  refactor: scopes,
   // Performance improvements
-  perf: calendarScopes,
+  perf: scopes,
   // Add or correct tests
-  test: calendarScopes,
-  // Changes that affect
-  style: calendarScopes,
+  test: scopes,
+  // Changes that affect style, linting, formatting and grammar
+  style: scopes,
 };
 
 export default {
@@ -43,7 +43,7 @@ export default {
       2,
       'always',
       // From `commitlint` point of view, allow all scopes regardless of the type
-      [...calendarScopes].filter(Boolean),
+      scopes,
     ],
     // Use `selective-scope` to make sure that the scopes are only allowed when they are defined for a particular type
     'selective-scope': [2, 'always', typesEnumScoped],
--- a/.github/workflows/romcal-ci.yml
+++ b/.github/workflows/romcal-ci.yml
@@ -6,10 +6,37 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Install Dependencies
+        run: npm -q --no-progress ci --no-audit
+
+      - name: Validate with commitlint (push)
+        if: github.event_name == 'push'
+        run: npx commitlint --last --verbose
+
+      - name: Validate PR commits with commitlint
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.head.sha }}~${{ github.event.pull_request.commits }} --to ${{ github.event.pull_request.head.sha }} --verbose
+
   build:
     runs-on: ubuntu-latest
-
+    needs: commitlint
     env:
       CI: true
 
--- a/.github/workflows/romcal-ci.yml
+++ b/.github/workflows/romcal-ci.yml
@@ -67,3 +67,20 @@ jobs:
 
       - name: Ensure Data Integrity
         run: npm run data:checks
+
+  check-dedupe:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Git checkout dev branch
+        uses: actions/checkout@v4
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Check for dependency duplicates
+        run: |
+          npm dedupe
+          git diff --exit-code || (echo "package-lock.json is not up-to-date" && exit 1)
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,4 +5,4 @@ echo "Running PRE-COMMIT hook"
 npx lint-staged
 npm dedupe
 
-echo "Don't for get to commit your deduped package-lock.json!"
+echo "Don't forget to commit your deduped package-lock.json!"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -3,6 +3,3 @@
 echo "Running PRE-COMMIT hook"
 
 npx lint-staged
-npm -q --no-progress dedupe
-
-echo "Don't forget to commit your deduped package-lock.json!"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -3,6 +3,6 @@
 echo "Running PRE-COMMIT hook"
 
 npx lint-staged
-npm dedupe
+npm -q --no-progress dedupe
 
 echo "Don't forget to commit your deduped package-lock.json!"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+echo "Running PRE-COMMIT hook"
+
+npx lint-staged
+npm dedupe
+
+echo "Don't for get to commit your deduped package-lock.json!"
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -144,16 +144,16 @@ General format/terminology:
 
 Here are some guardrails to follow:
 
-When "type" one of the keys, the scopes allowed are as follows:
+When the "type" is one of the keys, the scopes allowed are as follows:
 
 - `ci` (updates to automation or the release process): No scope allowed
 - `build` (updates to build scripts): No scope allowed
 - `docs` (documentation): No Scope allowed
 - `feat` (net-new feature):
-  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
-  - Any of the locale ids (`en`, `fr`, `es`, etc.)
-  - 'l10n' for changes to localization, either multiple locales or not specific to one
-  - 'calendar' for changes to calendars, either multiple or not specific to one
+  - Any of the calendar IDs (`general` or `europe`, or `slovakia`, etc.)
+  - Any of the locale IDs (`en`, `fr`, `es`, etc.)
+  - `l10n` for changes to localization, either multiple locales or not specific to one
+  - `calendar` for changes to calendars, either multiple or not specific to one
 - `enh` (enhancement): see `feat`
 - `fix` (bug fix): see `feat`
 - `refactor` (refactoring of code): see `feat`
@@ -165,11 +165,11 @@ When "type" one of the keys, the scopes allowed are as follows:
 
 (of staged files)
 
-Files you stage & commit are automatically linted, and fixes applied.
+Files you stage & commit are automatically linted, and fixes are applied to your commit.
 
 If it fails linting, your commit will fail.
 
 ### Dependency Deduping
 
-Will be performed automatically, but you are responsible for committing it at your convenience.
+Dependency deduping will be performed automatically, but you are responsible for committing it at your convenience.
 Dedupe checking is performed in GitHub Actions, and if it fails, the build will fail.
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -144,7 +144,7 @@ General format/terminology:
 
 Here are some rules to follow:
 
-When the "type" is one of the keys, the scopes allowed are as follows:
+When the "type" is one of the keys, the scopes allowed follow the successive rules:
 
 - `ci`:
   - scopes: No scope allowed
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -125,3 +125,47 @@ _pass the `-t` flag and the path of the file_
 ## Coverage reporting
 
 A coverage report is generated for every core branch build and published to [gh-pages]().
+
+## Committing changes
+
+### Message Format
+When you are ready to commit your changes, your commit message should follow the conventional commit format
+
+General format/terminology:
+```
+<type>[optional scope]: <description>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+Here are some guardrails to follow:
+
+When "type" one of the keys, the scopes allowed are as follows:
+- `ci` (updates to automation or the release process): No scope allowed
+- `build` (updates to build scripts): No scope allowed
+- `docs` (documentation): No Scope allowed
+- `feat` (net-new feature):
+  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
+  - Any of the locale ids (`en`, `fr`, `es`, etc.)
+  - 'l10n' for changes to localization, either multiple locales or not specific to one
+  - 'calendar' for changes to calendars, either multiple or not specific to one
+- `enh` (enhancement): see `feat`
+- `fix` (bug fix): see `feat`
+- `refactor` (refactoring of code): see `feat`
+- `perf` (performance improvements): see `feat`
+- `test` (adding or updating tests): see `feat`
+- `style` (formatting, missing semi-colons, etc; no code change): see for `feat`
+
+### Automatic linting
+(of staged files)
+
+Files you stage & commit are automatically linted, and fixes applied.
+
+If it fails linting, your commit will fail.
+
+### Dependency Deduping
+
+Will be performed automatically, but you are responsible for committing it at your convenience.
+Dedupe checking is performed in GitHub Actions, and if it fails, the build will fail.
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -129,9 +129,11 @@ A coverage report is generated for every core branch build and published to [gh-
 ## Committing changes
 
 ### Message Format
+
 When you are ready to commit your changes, your commit message should follow the conventional commit format
 
 General format/terminology:
+
 ```
 <type>[optional scope]: <description>
 
@@ -143,6 +145,7 @@ General format/terminology:
 Here are some guardrails to follow:
 
 When "type" one of the keys, the scopes allowed are as follows:
+
 - `ci` (updates to automation or the release process): No scope allowed
 - `build` (updates to build scripts): No scope allowed
 - `docs` (documentation): No Scope allowed
@@ -159,6 +162,7 @@ When "type" one of the keys, the scopes allowed are as follows:
 - `style` (formatting, missing semi-colons, etc; no code change): see for `feat`
 
 ### Automatic linting
+
 (of staged files)
 
 Files you stage & commit are automatically linted, and fixes applied.
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -142,9 +142,7 @@ General format/terminology:
 [optional footer(s)]
 ```
 
-Here are some rules to follow:
-
-When the "type" is one of the keys, the scopes allowed follow the successive rules:
+In the following list, you can find what scopes can be used for each allowed type.
 
 - `ci`:
   - scopes: No scope allowed
--- a/docs/contribute-to-romcal.md
+++ b/docs/contribute-to-romcal.md
@@ -1,6 +1,6 @@
 **:warning: The documentation above is written for romcal v2, and has not been updated yet to romcal v3.**
 
-# Install, Build, Run and Test romcal locally
+# Install, Build, Run and Test romcal Locally
 
 This section describes how to run the romcal source code in your machine, and how to contribute to this project.
 You might also be interested in reading the general [contribution guide](../CONTRIBUTING.md).
@@ -12,7 +12,7 @@ You might also be interested in reading the general [contribution guide](../CONT
 - [Test](#test)
 - [Coverage reporting](#coverage-reporting)
 
-## Retrieve the codebase
+## Retrieve the Codebase
 
 1. [Fork the repo](https://help.github.com/en/github/getting-started-with-github/fork-a-repo#fork-an-example-repository).
 
@@ -122,11 +122,11 @@ npm run test -t ./src/utils/object/object.test.ts
 
 _pass the `-t` flag and the path of the file_
 
-## Coverage reporting
+## Coverage Reporting
 
 A coverage report is generated for every core branch build and published to [gh-pages]().
 
-## Committing changes
+## Committing Changes
 
 ### Message Format
 
@@ -142,26 +142,46 @@ General format/terminology:
 [optional footer(s)]
 ```
 
-Here are some guardrails to follow:
+Here are some rules to follow:
 
 When the "type" is one of the keys, the scopes allowed are as follows:
 
-- `ci` (updates to automation or the release process): No scope allowed
-- `build` (updates to build scripts): No scope allowed
-- `docs` (documentation): No Scope allowed
-- `feat` (net-new feature):
-  - Any of the calendar IDs (`general` or `europe`, or `slovakia`, etc.)
-  - Any of the locale IDs (`en`, `fr`, `es`, etc.)
-  - `l10n` for changes to localization, either multiple locales or not specific to one
-  - `calendar` for changes to calendars, either multiple or not specific to one
-- `enh` (enhancement): see `feat`
-- `fix` (bug fix): see `feat`
-- `refactor` (refactoring of code): see `feat`
-- `perf` (performance improvements): see `feat`
-- `test` (adding or updating tests): see `feat`
-- `style` (formatting, missing semi-colons, etc; no code change): see for `feat`
+- `ci`:
+  - scopes: No scope allowed
+  - use: Updates to automation or the release process
+- `build`:
+  - scopes: No scope allowed
+  - use: updates to build scripts
+- `docs`:
+  - scopes: No Scope allowed
+  - use: documentation changes, updates, or additions
+- `feat`:
+  - scopes:
+    - Any of the calendar IDs (`general` or `europe`, or `slovakia`, etc.)
+    - Any of the locale IDs (`en`, `fr`, `es`, etc.)
+    - `l10n` for changes to localization, either multiple locales or not specific to one
+    - `calendar` for changes to calendars, either multiple or not specific to one
+  - use: new features
+- `enh`:
+  - scopes: see `feat`
+  - use: enhancements to existing features
+- `fix`:
+  - scopes: see `feat`
+  - use: bug fixes
+- `refactor`:
+  - scopes: see `feat`
+  - use: refactoring code, maintaining the same behaviors
+- `perf`:
+  - scopes: see `feat`
+  - use: performance improvements, but not changing behavior
+- `test`:
+  - scopes: see `feat`
+  - use: adding or updating tests
+- `style`:
+  - scopes: see for `feat`
+  - use: formatting, missing semicolons, etc.; no code change
 
-### Automatic linting
+### Automatic Linting
 
 (of staged files)
 
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
   "devDependencies": {
     "@commitlint/cli": "^19.1.0",
     "@commitlint/config-conventional": "^19.1.0",
+    "@commitlint/types": "^19.0.3",
     "@jsdevtools/npm-publish": "^1.4.3",
     "@types/cli-progress": "^3.11.5",
     "@types/humanize-duration": "^3.27.4",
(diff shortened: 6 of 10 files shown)


TobiTenno:
@emagnier @tukusejssirs i can't think of anything else we really need at first glance for the initial round of semantic release message support, i'd like to do the actual switch to semantic release in a separate PR, since it could get pretty complicated

.github/workflows/romcal-ci.yml
```diff
@@ -6,10 +6,37 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
tukusejssirs, line 11: Does this cancel previous jobs in progress or if there are other jobs in progress, it cancels this one? 🤔

.husky/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+echo "Running PRE-COMMIT hook"
+
+npx lint-staged
+npm dedupe
+
+echo "Don't for get to commit your deduped package-lock.json!"
```
tukusejssirs: What is the reason of this comment/log?

It `package-lock.json` should not be committed, we could run `git checkout package-lock.json`.

However, I am not sure it shouldn’t be committed. I’d actually go further and made sure the lock file is up to date with `package.json` and force-re-install the packages ignoring the lock file and committing the lock file if there are any changes.

WDYT? 🤔

package.json
```diff
@@ -26,7 +25,8 @@
     "test": "TZ=UTC jest --coverage",
     "test:without-coverage": "TZ=UTC jest",
     "test:watch": "TZ=UTC jest --watch",
-    "test:snapshot:update": "TZ=UTC jest -u"
+    "test:snapshot:update": "TZ=UTC jest -u",
+    "prepare": "husky"
```
tukusejssirs, line 29: Is this actually required in the latest Husky version? I slightly remember it is not used anymore. I don’t have in my private package, but Husky should work without preparation. 🤔

package.json
```diff
@@ -101,5 +104,14 @@
     "type-fest": "^4.10.3",
     "typescript": "~5.2.2"
   },
+  "lint-staged": {
```
tukusejssirs, line 109: I have never used `lint-staged` (`nx affected` does similar job, probably even better), however, why is it required to duplicate the config for TS and MD/YAML? ESLint skips files that are not matched by the rules. 🤔

.commitlintrc
```diff
@@ -0,0 +1,3 @@
+{
+  extends: ["@commitlint/config-conventional"]
```
tukusejssirs: I believe you wanted to customise the config more. :wink:

Consider this comment just a note without any action required.

tukusejssirs (changes requested):
Some nitpicks and questions.

tukusejssirs:
> i can't think of anything else we really need at first glance for the initial round of semantic release message support, i'd like to do the actual switch to semantic release in a separate PR, since it could get pretty complicated

Okay, thanks for the PR! 🙏

.github/workflows/romcal-ci.yml
TobiTenno: it cancels any job that isn't the latest

package.json
TobiTenno: this was automatically added by husky, not me

package.json
TobiTenno: this doesn't use the precise same config as lint/prettier as it also fixes. this is the only way i've used it, but i'm open to trying it with other configurations, this is simply the most clear for me.

.commitlintrc
TobiTenno: i actually like the default conventional config, but i'm happy to config more. what configurations would you like to see differently configured?

.husky/pre-commit
TobiTenno: no, this is saying it _should_ be committed, because `npm dedupe` ensures the package-lock.json is deduped and minimized to decrease the installs necessary. I was debating also immediately `git add`'ing package-lock.json to accomplish this, but i felt this would be inconsistent with how lint-staged will modify files but also not add them to git.

package.json
tukusejssirs: I don’t want to force you to follow my preferences, however, IMHO in CI, we should only _check_ the style (ESLint and Prettier), never fix it. I get it that it might be easier for devs, however, I still think that all the changes (except for release-related changes) should be done by the dev, incl the code styling and formatting. If there are any issues (code, test, lint or format), it should be fixed by the dev.

.commitlintrc
tukusejssirs: Well, we might use the default config, I just thought you wanted to configure it more.

I prefer to list all allowed types and scopes, and use the [`selective-scope`](www.npmjs.com/package/commitlint-plugin-selective-scope) plugin to make sure that some scopes are used only for specific types (e.g. all package names could be used as scopes for any type but `chore`; for `chore` type, I only use `[null, 'build', 'ci', 'deps']` where `null` means no scope).

I could show you my `commitlint.config.js` if you want.

.husky/pre-commit
tukusejssirs: Then, why don’t we actually check it is committed? For example using this:

```sh
if [ -f package-lock.json ]; then
  echo "Don't for get to commit your deduped package-lock.json!"
  exit 2
fi
```

Or even better: why don’t we first install the packages (thus the lock file is created), dedupe it and commit the lock file if it has any meaningful changes? :thinking:

> i felt this would be inconsistent with how lint-staged will modify files but also not add them to git

I have never used `lint-staged`, therefore, that is up to you, however, it might be a good idea to commit the changed lock file if there are any meaninful changes.

package.json
TobiTenno: this isn't CI, this is pre-push

.commitlintrc
TobiTenno: i'm happy to take a look, i don't think i'd agree that we should use selective-scope without good reason

package.json
tukusejssirs: Okay, if it is pre-push, linting and formatting should be already checked in the CI before we can merge the PR. :thinking:

.husky/pre-commit
TobiTenno: ok, but this is in husky, meaning it's on the developer side before the commit takes place. the lock file is already created, dedupe requires that packages are already installed, or it would fail. i don't feel the need to do _everything_ for the developer, contributors should be expected to at least set up the project if they're doing anything that would change either the node modules in package.json or package-lock.json

package.json
TobiTenno: yeah, this ensures that won't be a problem when the commit gets up to the PR. 

husky can't protect web-based commits anyway, so the CI action to check the above are still needed to protect anyone
- making web-based commits
- bypassing pre-commit checks (`git commit -n`)

.commitlintrc
tukusejssirs: <details>
  <summary><code>commitlint.config.js</code></summary>

```js
const {existsSync, readdirSync} = require('node:fs')

const appsLibsScopes = [null]

// For `chore`, these are the only allowed scopes.
const choreScopes = [null, 'build', 'ci', 'deps']

// List all apps and libraries, and all Nest modules in them
// Note: I use Nx monorepo and Nest.js app.
// Note: I still want to fix this to select Nest modules only. For now, all it does is select subfolders under `{apps,libs}/$app/src`. I also want to get the apps and libs list programmatically using Nx.
// Examples: `app1`, `app2`, `app1|module1`.
;[`${__dirname}/apps`, `${__dirname}/libs`].forEach(f => {
  if (existsSync(f)) {
    readdirSync(f, {withFileTypes: true})
      .filter(dirent => dirent.isDirectory())
      .map(dirent => dirent.name)
      .forEach(s => {
        if (existsSync(`${f}/${s}/src`)) {
          const modules = readdirSync(`${f}/${s}/src`, {withFileTypes: true})
            .filter(dirent => dirent.isDirectory())
            .map(dirent => dirent.name)

          appsLibsScopes.push(s, ...modules.map(m => `${s}|${m}`))
        } else {
          appsLibsScopes.push(s)
        }
      })
  }
})

// `docs` location
// Note: That I am in the process of moving the root-level `docs` folder to a Vitepress app under `apps/`.
const docsFolder = `${__dirname}/docs`

// List app docs subfolders which are treated the same way as the modules of apps/libs.
// Examples: `docs`, `docs/contribution`, `docs/usage`.
const docsScopes = [
  null,
  ...(existsSync(docsFolder)
    ? readdirSync(docsFolder, {withFileTypes: true}).map(i =>
        i.isFile() ? i.name.replace(/\.md/, '') : i.name
      )
    : [])
]

// Define all allowed scopes for each type
// Note: I know I could simplify this using a loop/`Array#map()`, however, this is more readable.
const typesEnumScoped = {
  // Dev tools related stuff, not related to the code itself
  // Commits, that affect build components like build tool, CI pipeline, dependencies, project version, ...
  // Miscellaneous commits, e.g. modifying `.gitignore`
  chore: choreScopes,

  // Commits, that affect documentation only
  docs: docsScopes,

  // Commits, that improves an existing feature (minor version bump)
  enh: appsLibsScopes,

  // Commits, that adds a new feature (minor version bump)
  feat: appsLibsScopes,

  // Commits that fix a bug (patch version bump)
  fix: appsLibsScopes,

  // Commits are special `refactor` commits, that improve performance
  perf: appsLibsScopes,

  // Commits, that rewrite/restructure your code, however, does not change any behaviour
  refactor: appsLibsScopes,

  // Commits that revert previous commits
  revert: appsLibsScopes,

  // Commits that do not affect the meaning (white-space, formatting, missing semicolons, etc)
  style: appsLibsScopes,

  // Commits, that add missing tests or correcting existing tests
  test: appsLibsScopes
}

module.exports = {
  extends: ['@commitlint/config-conventional'],
  formatter: '@commitlint/format',
  // TODO: Change this URL to our Wiki page with the description of the commit message format.
  helpUrl:
    'https://github.com/conventional-changelog/commitlint/#what-is-commitlint',
  parserPreset: {
    parserOpts: {
      issuePrefixes: ['#']
    }
  },
  plugins: ['selective-scope'],
  rules: {
    'body-case': [2, 'always', 'sentence-case'],
    'body-empty': [0, 'never'],
    'body-full-stop': [2, 'always', '.'],
    'body-leading-blank': [2, 'always'],
    'body-max-line-length': [0, 'never', 0],
    'footer-empty': [0, 'never'],
    'footer-leading-blank': [2, 'always'],
    'footer-max-line-length': [0, 'never', 0],
    'header-full-stop': [2, 'never', '.'],
    'header-max-length': [0, 'never', 0],
    'scope-enum': [
      2,
      'always',
      // From `commitlint` point of view, allow all scopes regardless of the type
      [...appsLibsScopes, ...choreScopes, ...docsScopes].filter(i => i !== null)
    ],
    // Use `selective-scope` to make sure that the scopes are only allowed when they are defined for a particular type
    'selective-scope': [2, 'always', typesEnumScoped],
    'subject-case': [1, 'always', 'lower-case'],
    'subject-empty': [2, 'never'],
    'subject-full-stop': [2, 'never', '.'],
    'type-case': [2, 'always', 'lower-case'],
    'type-enum': [2, 'always', Object.keys(typesEnumScoped)]
  }
}
```

</details>

tukusejssirs: > i don't think i'd agree that we should use selective-scope without good reason

And I don’t want to force you to use it. All I wanted to say is that without it, `commitlint` allows any `type` and `scope` matches which IMHO might not be what one wants.

TobiTenno: this seems like something we could do as we work on the monorepo, but i don't think it's necessary for the current state, since we don't have several of these distinctive locations for most things, without reorganizing significantly, unless we just want specific scopes for constants, models, and calendars, but i don't feel those would be significantly impactful right now.

.husky/pre-commit
TobiTenno: I'm still looking at the right way to potentially add anything staged so the experience is that the fixes "just happens" and we don't have to worry about committing twice or readding files or anything like that

TobiTenno: i say that, then i find that lint-staged automatically handles all of that, so it'd just be for the dedupe part that i wanna sort it out

.commitlintrc
TobiTenno: that said, you do give a really good template even if we just want to do those 3 for now, and otherwise allowing scopes as empty for other files... what do you think of that?

tukusejssirs: > unless we just want specific scopes for constants, models, and calendars, but i don't feel those would be significantly impactful right now

If you look at my commit messages in romcal, for scope of `feat` and `fix` types I try to use the following:

- `l10n`: related to l10n changes either of multiple locales or unrelated to a single l10n file;
- `calendar`: related to calendar changes either of multiple calendars or unrelated to a single l10n files;
- `$localeId` (e.g. `la` or `en`): related to a single locale;
- `$calendarId` (e.g. `general` or `europe` or `slovakia`): related to a single calendar.

For style changes, I used to use `chore(style)`, however, not anymore. Now, I use `style` type with either no scope or one of the scopes listed above.

The rest could be the same as I suggested in the `commitlint.config.js` … unless you disagree.

.husky/pre-commit
tukusejssirs: > I'm still looking at the right way to potentially add anything staged so the experience is that the fixes "just happens" and we don't have to worry about committing twice or readding files or anything like that

I don’t like the idea to stage anything automatically, as I usually have multiple files changed, but I might commit them separately.

.commitlintrc
TobiTenno: yeah, i'm pretty fine with these, but i think the distinction of scope and type are important here. I'm fine with having restricted scopes, but not restricted types. I think specifying `feat`/`fix`/etc are still critical for determination of change type, and not just making everything either a minor or patch change.

specifically referring to this structure [[src](https://www.conventionalcommits.org/en/v1.0.0/)]:
```
<type>[optional scope]: <description>

[optional body]

[optional footer(s)]
```

tukusejssirs: > I'm fine with having restricted scopes, but not restricted types.

I’m not sure I understand you, however, both types and scopes are restricted (albeit separately; see the [type list in `@commitlint/config-conventional`](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum)) by `commitlint` by default. :thinking: All I suggest to restrict the possible type-scope pairs.

.husky/pre-commit
TobiTenno: yeah, this only stages those files you've already included in your commit (from lint-staged), and the only one i'm looking to add is package-lock to be automatically staged, but as-is, we're only linting & adding the files you've already staged

.commitlintrc
TobiTenno: i think i'm willing to adapt some of what you have into a config for us. I'll work on that in a few hours. I misread 

> for scope of `feat` and `fix` types I try to use

to mean something you are not saying. Thank you for your clarification.

.husky/pre-commit
tukusejssirs: As I have already said, I disagree with staging any files automatically. I might have some unstaged changes to `package*.json` files, however, I might want commit those changes in a later commit.

TobiTenno: right, what i'm saying is i'm not changing that yet, and given your input, i won't automatically add package*.json

TobiTenno: I am considering, that said, requiring a "dedupe" check after the test that would fail CI if your package-lock isn't deduped so it would need to be deduped before the PR could be merged

.commitlintrc
tukusejssirs: > i think i'm willing to adapt some of what you have into a config for us. I'll work on that in a few hours.

Okay, I’ll review it then, probably tomorrow CET/UTC+01.

> I misread […]

Well, yeah, I wrote it fast. All I wanted to say by _for scope of `feat` and `fix` types I try to use_ is that for those types I have already used those scopes, however, I didn’t say anything about other types which should be kept as is, or possibly, we might want to move feature enhancements to `enh` from `feat` and keep `feat` for new features only. We also might want to merge/reduce `build`, `chore` and `ci` (I prefer using only `chore` with `build` and `ci` to be scopes for `chore`, however, we could keep them as they are).

.husky/pre-commit
tukusejssirs: > 

I am considering, that said, requiring a "dedupe" check after the test that would fail CI if your package-lock isn't deduped so it would need to be deduped before the PR could be merged

Go for it if you want, I have nothing against that. :wink:

.commitlintrc
TobiTenno: i think, `feat` should stay to cover both scenarios as-is, as they both produce the same minor version bump. Would you see `enh` as a patch rather than a minor version change?

for `build`/`chore`/`ci`, i'm down reducing chore out and only allowing `build` and `ci` for specific differences, as i think anything that's a `chore` in this case would really be either a `build` or `ci` change, and `chore` could be eliminated entirely. I'm open to your opinion on this.

tukusejssirs: > i think, `feat` should stay to cover both scenarios as-is, as they both produce the same minor version bump. Would you see `enh` as a patch rather than a minor version change?

From release point of view, both `feat` and `enh` would cause a `minor` bump, however, from the changelog point of view, I suggest to group them separately (just like any other type).

> for `build`/`chore`/`ci`, i'm down reducing chore out and only allowing `build` and `ci` for specific differences, as i think anything that's a `chore` in this case would really be either a `build` or `ci` change, and `chore` could be eliminated entirely. I'm open to your opinion on this.

As you can see in the shared `commitlint.config.js`, I use `chore` with three scopes (`build`, `ci`, and `deps`, the last one used when updating the deps). I also use a scopeless `chore` type for other changes, e.g. when updating/adding `.gitignore` or `commitlint.config.js` or `update_dependencies.sh` and for initial commit (`chore: bootstrap Nx monorepo`). I also modified the MR (in GitLab; in GitLab) message to the following:

```bash
chore: merge branch `$featBranch` to `dev`
```

Changing the merge message is irrelevant to romcal, as we squash the changes.

.husky/pre-commit
TobiTenno: i've got that part added

.commitlintrc
TobiTenno: i'm open to the `feat`/`enh` distinction on release. For the sake of enabling that, i'll include the distinction in our config.

As for the `chore`/`build`/`ci` difference, i find [angular's reasoning](https://gist.github.com/brianclements/841ea7bffdb01346392c?permalink_comment_id=3102984#gistcomment-3102984) for removing chore and splitting it off to cover these cases...

one case of this... most CI already include skipping ci (how romcal does on `ci:` commits) via `[skip ci]`, which i think makes more sense.

`build`: covering build script changes, or just someone doing a manual build to deploy

`ci`: covering ci; such as actions, scripts for releases, configuration that doesn't directly tie to `build` updates

TobiTenno: changes/configs added for this, i also moved `dedupe` to be run only if you actually modify the package-lock, so this would keep it more in line with now the lint works, and keeps it a bit faster so you don't dedupe every time

TobiTenno: @tukusejssirs how do you prevent your commitlint from failing to load `selective-scope`? I have it as a dev dependency, but i'm not quite sure why it's being weird

.commitlintrc.ts
```diff
@@ -0,0 +1,52 @@
+import { UserConfig } from '@commitlint/types';
+import { LOCALE_IDS } from './lib/constants/locales';
+import { CALENDAR_IDS } from './lib/constants/calendars';
+
+const empty = [null];
+const calendarScopes = [...CALENDAR_IDS, ...LOCALE_IDS, 'l10n', 'calendar'];
```
tukusejssirs: As we have there `LOCALE_IDS`, I don’t think they are `calendarScopes`, but `romcalScopes` or just `scopes`.

.commitlintrc.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const typesEnumScoped = {
+  // Updates to the automation or release process
+  ci: empty,
+  // Changes to build scripts
+  build: empty,
+  // Documentation only changes
+  docs: empty,
+  // Features, net-new functionality
```
tukusejssirs: Why I assume `net-new` means `totally new`, IMO `new` is sufficient.

```suggestion
  // Features, new functionality
```

.commitlintrc.ts
```diff
@@ -0,0 +1,52 @@
...
+      issuePrefixes: ['#'],
+    },
+  },
+  plugins: ['selective-scope'],
+  rules: {
+    'scope-enum': [
+      2,
+      'always',
+      // From `commitlint` point of view, allow all scopes regardless of the type
+      [...calendarScopes].filter(Boolean),
```
tukusejssirs: I’m just curious: why do we need to `filter()` the array when it does not contain `null` (whichhave moved to `empty`)? 🤔

.commitlintrc.ts
```diff
@@ -0,0 +1,52 @@
...
+  enh: calendarScopes,
+  // Bug fixes
+  fix: calendarScopes,
+  // Refactoring, no behavior changes
+  refactor: calendarScopes,
+  // Performance improvements
+  perf: calendarScopes,
+  // Add or correct tests
+  test: calendarScopes,
+  // Changes that affect
```
tukusejssirs:
```suggestion
  // Changes that affect style, linting, formatting and grammar
```

.commitlintrc.ts
```diff
@@ -0,0 +1,52 @@
...
+  style: calendarScopes,
+};
+
+export default {
+  extends: ['@commitlint/config-conventional'],
+  formatter: '@commitlint/format',
+  helpUrl: 'https://github.com/romcal/romcal/blob/dev/docs/contribute-to-romcal.md#committing-changes',
```
tukusejssirs, line 34: I’d personally move the `Comitting changes` to `docs`, maybe to [`docs/contribute-to-romcal.md`](https://github.com/romcal/romcal/blob/dev/docs/contribute-to-romcal.md), however, that is just a preference.

Actually, you did document it there, however, you link to `README.md`. I get you want to link to the entire contributing docs, however, in this case, I’d prefer using a direct link to the commintlint types and scopes definition.

.commitlintrc
tukusejssirs: I have had issues with _loading_ the plugin. All I did is to install it and configure it in `commitlint.config.js`.

What it do when it fails to load? 🤔

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+```
+<type>[optional scope]: <description>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+Here are some guardrails to follow:
```
tukusejssirs: I’d probably replace `guardrails` with a simpler word, as the romcal contributors have ranging English skills, therefore, we should try to use simple language.

Just a note: I have never heard about this word before. :wink:

Maybe `rules`? 🤔

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+When "type" one of the keys, the scopes allowed are as follows:
```
tukusejssirs: The first clause is missing a verb, maybe `is`. 🤔

And what _keys_? l10n keys? 🤔

```suggestion
When the "type" is one of the keys, the scopes allowed are as follows:
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+- `ci` (updates to automation or the release process): No scope allowed
```
tukusejssirs: In this list, I’d probably  use lowercase initial letter and append semicolon to each list item. Again, just a preferencd. :wink:

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+- `build` (updates to build scripts): No scope allowed
+- `docs` (documentation): No Scope allowed
+- `feat` (net-new feature):
+  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
+  - Any of the locale ids (`en`, `fr`, `es`, etc.)
+  - 'l10n' for changes to localization, either multiple locales or not specific to one
+  - 'calendar' for changes to calendars, either multiple or not specific to one
```
tukusejssirs:
```suggestion
  - `calendar` for changes to calendars, either multiple or not specific to one
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+Here are some guardrails to follow:
+
+When "type" one of the keys, the scopes allowed are as follows:
+
+- `ci` (updates to automation or the release process): No scope allowed
+- `build` (updates to build scripts): No scope allowed
+- `docs` (documentation): No Scope allowed
+- `feat` (net-new feature):
+  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
+  - Any of the locale ids (`en`, `fr`, `es`, etc.)
```
tukusejssirs:
```suggestion
  - Any of the locale IDs (`en`, `fr`, `es`, etc.)
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+Here are some guardrails to follow:
+
+When "type" one of the keys, the scopes allowed are as follows:
+
+- `ci` (updates to automation or the release process): No scope allowed
+- `build` (updates to build scripts): No scope allowed
+- `docs` (documentation): No Scope allowed
+- `feat` (net-new feature):
+  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
```
tukusejssirs:
```suggestion
  - Any of the calendar IDs (`general` or `europe`, or `slovakia`, etc.)
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+  - Any of the locale ids (`en`, `fr`, `es`, etc.)
+  - 'l10n' for changes to localization, either multiple locales or not specific to one
```
tukusejssirs:
```suggestion
  - `l10n` for changes to localization, either multiple locales or not specific to one
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+  - 'calendar' for changes to calendars, either multiple or not specific to one
+- `enh` (enhancement): see `feat`
+- `fix` (bug fix): see `feat`
+- `refactor` (refactoring of code): see `feat`
+- `perf` (performance improvements): see `feat`
+- `test` (adding or updating tests): see `feat`
+- `style` (formatting, missing semi-colons, etc; no code change): see for `feat`
```
tukusejssirs: I usually include changes related to formatting, linting and grammar.

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+- `ci` (updates to automation or the release process): No scope allowed
+- `build` (updates to build scripts): No scope allowed
+- `docs` (documentation): No Scope allowed
+- `feat` (net-new feature):
+  - Any of the calendar ids (`general` or `europe`, or `slovakia`, etc.)
+  - Any of the locale ids (`en`, `fr`, `es`, etc.)
+  - 'l10n' for changes to localization, either multiple locales or not specific to one
+  - 'calendar' for changes to calendars, either multiple or not specific to one
+- `enh` (enhancement): see `feat`
```
tukusejssirs: Where in `feat`? 🤔

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+- `refactor` (refactoring of code): see `feat`
+- `perf` (performance improvements): see `feat`
+- `test` (adding or updating tests): see `feat`
+- `style` (formatting, missing semi-colons, etc; no code change): see for `feat`
+
+### Automatic linting
+
+(of staged files)
+
+Files you stage & commit are automatically linted, and fixes applied.
```
tukusejssirs: I usually prefer typing ampersand out unless used for styling.

```suggestion
Files you stage and commit are automatically linted with fixes applied.
```

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+If it fails linting, your commit will fail.
+
+### Dependency Deduping
```
tukusejssirs, line 190: Consistency nitpick: use the same the heading casing everywhere, either sentence case or title case. 😉

docs/contribute-to-romcal.md
```diff
@@ -125,3 +125,51 @@ _pass the `-t` flag and the path of the file_
...
+
+Will be performed automatically, but you are responsible for committing it at your convenience.
```
tukusejssirs: Missing subject in the first clause.

```suggestion
Dependency deduping will be performed automatically, but you are responsible for committing it at your convenience.
```

tukusejssirs (changes requested):
Here are some suggestions and, OFC, nitpicks. 😉

.commitlintrc
tukusejssirs: > As for the `chore`/`build`/`ci` difference, i find [angular's reasoning](https://gist.github.com/brianclements/841ea7bffdb01346392c?permalink_comment_id=3102984#gistcomment-3102984) for removing chore and splitting it off to cover these cases...

So, changes to `.gitignore` (and other, similar changes) are made under the `build` type, right? :thinking:

.commitlintrc.ts
TobiTenno: fair. i'll refactor that

.commitlintrc.ts
TobiTenno: uhhh, committing changes is in docs... I'm not linking to the readme, i'm linking to contribute-to-romcal.md

.commitlintrc.ts
TobiTenno: i just left the filter. i don't think there's a reason not to remove this unless we add in null and forget to filter, but that's probably excessive pre-optimization.
```suggestion
      calendarScopes,
```

docs/contribute-to-romcal.md
TobiTenno:
```suggestion
Here are some rules to follow:
```

docs/contribute-to-romcal.md
TobiTenno: keys are the types in the UL below that indicate which you're checking

docs/contribute-to-romcal.md
TobiTenno: this is markdown, not js :thinking:

docs/contribute-to-romcal.md
TobiTenno: the least-nested list item before this

docs/contribute-to-romcal.md
TobiTenno: `etc.`

docs/contribute-to-romcal.md
TobiTenno:
```suggestion
Files you stage & commit are automatically linted, and fixes are applied to your commit.
```

docs/contribute-to-romcal.md
TobiTenno: i think this is a bit too nit-picky, but i'll update it

docs/contribute-to-romcal.md
TobiTenno: it's literally already the title, but ok.... this is a common shorthand.

.commitlintrc
TobiTenno: `ci`, you're preventing more files from being checked in and affecting ci, not changing build scripts or how the build works. the build will still use anything in `.gitignore`

TobiTenno: > I have had issues with loading the plugin. All I did is to install it and configure it in commitlint.config.js.
> 
> What it do when it fails to load? 🤔

I've seen 
```diff
- Failed to load plugin commitlint-plugin-selective-scope.
```

so that's probably just a weirdness issue where i might just have to have `npm install` before lint-staged to make sure everything's enforced

.commitlintrc.ts
tukusejssirs: My bad then, sorry.

docs/contribute-to-romcal.md
tukusejssirs: Sorry, I still don’t follow. What is _UL_?

If _keys_ are _types_, which use two different names of one thing? :thinking:

I really thing I don’t get what you wanted to say by that sentence. Even though I think the problem is with me, just imagine if it is read by someone who don’t understand English that much. And we have such contributors, as we try to localise romcal data (let’s say) to any language and even the calendar data needs to be acquired from different languages. :thinking:

docs/contribute-to-romcal.md
tukusejssirs: Yeah, it is MD, however:

- I personally don’t use semicolons after commands in JS (unless they are necessary);
- I treat MD files as text in articles/books, i.e. they should follow some style (as in _manual of style_, e.g. Chicago Manual of Style or MLA.

Therefore, I suggest to choose a style and use it consistently. I personally (and you don’t need to accept it) prefer ending lists with either a comma (unless a comma is used at least once in at least one of the list items) or a semicolon (which can be used even there is no comma in any of the list items).

Sorry, I am a _formalist_, as in _I love the form (in contrary to the content)_. Feel free to ignore my and keep it as is. :wink:

docs/contribute-to-romcal.md
tukusejssirs: I’d rather write more examples, however, feel free to keep it as it. :wink:

docs/contribute-to-romcal.md
tukusejssirs: Can’t help myself. However, feel free to ignore my nitpicks, esp when I myself call a comment a nitpick.

Anyway, thanks! :pray:

docs/contribute-to-romcal.md
tukusejssirs: Agreed that it is a common shorthand, however, I am for using full sentences.

.commitlintrc.ts
TobiTenno: you're ok, it's a lot of docs!

docs/contribute-to-romcal.md
TobiTenno: the keys of the "map" below the description (`When "type" is one of the keys, the scopes allowed are as follows:`), i.e. `ci`, and the "values" are the descriptions with the scopes

how exactly would you say, without being very vague, "these are the allowed types and their allowed scopes"?

.commitlintrc
tukusejssirs: > I have **never** had issues with _loading_ the plugin. All I did is to install it and configure it in `commitlint.config.js`.

For some reason, I have missed the _never_ word in here. :thinking:

> so that's probably just a weirdness issue where i might just have to have `npm install` before lint-staged to make sure everything's enforced

It is still weird, because that error seems to come from `commitlint`, thus some packages need have been installed, while the plugin was still missing. :thinking:

Does it still happen to you? :thinking:

.commitlintrc.ts
tukusejssirs: Or maybe I review the PR too quickly (and I have done it a 6" screen of my phone). :man_shrugging:

docs/contribute-to-romcal.md
tukusejssirs: How about this: _In the following list, you can find what scopes can be used for each allowed type._

tukusejssirs (approved):
LGTM, @TobiTenno, thanks for bearing with me! :pray:

docs/contribute-to-romcal.md
TobiTenno: i can add some more while i clarify the description of the whole list

tukusejssirs: That said, maybe I should start to be less nit-picky, as there are much more important things to do with romcal.

.commitlintrc
TobiTenno: interestingly, only after commits where a dedupe happens

docs/contribute-to-romcal.md
TobiTenno: i think i've got the wording fixed as described. lemme know!

.commitlintrc
tukusejssirs: Maybe `npm dedupe` removes the plugin, as it can’t find it use anywhere in the code (`lib/`?). 🤔

docs/contribute-to-romcal.md
tukusejssirs: Thanks! 🙏

.commitlintrc
TobiTenno: 🤔  i don't think dedupe does tree shaking in that way

TobiTenno: hmm, trying to repro after manually deduping & running commitlint didn't cause issues for me

TobiTenno: i'm gonna chalk it up to me mistaking different machines that i was developing on at different times

PR #504, opened 2024-03-22
ci: correct commitlint usage

TobiTenno:
old docs outdated conventional-changelog/commitlint#3947

### DIFF
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Validate with commitlint
         if: github.event_name == 'push'
-        run: npx commitlint --from HEAD~1 --to HEAD --verbose
+        run: npx commitlint --last --verbose
 
   dev_branch_publishing:
     runs-on: ubuntu-latest
--- a/.github/workflows/dev-publishing.yml
+++ b/.github/workflows/dev-publishing.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Validate with commitlint
         if: github.event_name == 'push'
-        run: npx commitlint --last --verbose
+        run: npx commitlint --from HEAD~1 --to HEAD --verbose
 
   dev_branch_publishing:
     runs-on: ubuntu-latest
--- a/.github/workflows/romcal-ci.yml
+++ b/.github/workflows/romcal-ci.yml
@@ -13,8 +13,6 @@ concurrency:
 jobs:
   commitlint:
     runs-on: ubuntu-latest
-    # only lint pull requests here, pushes are linted in dev-publishing
-    if: github.event_name == 'pull_request'
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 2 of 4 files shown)


TobiTenno:
this time i tested locally to make sure this substitution from the removed `--last` flag worked

TobiTenno:
@tukusejssirs @emagnier trying to fix up the last of what i broke, i've tested this a bit more on other repos since i last changed this

tukusejssirs:
I’m just curious: what was wrong with `--last`?

In `commitlint --help` (using `@commitlint/cli@19.2.1`), I can see that `--last`/`-l` is also possible:

```ts
-l, --last           just analyze the last commit; applies if edit=false
```

That said, `-f HEAD~1 -t HEAD` should work exactly the same way.

TobiTenno:
@tukusejssirs 
that's what I thought was supposed to work as well, but when i tested locally, `--last` didn't exist for me

it's also not listed on their [CLI reference](https://commitlint.js.org/reference/cli.html)

TobiTenno:
ope! looks like they [just readded it](https://github.com/conventional-changelog/commitlint/releases/tag/v19.2.1)

TobiTenno:
i'll retest locally and see if i can sort it out, i'll add the commitlint version print as well

TobiTenno:
looks like the one i had installed is old. i'll get that updated and re-use `--last`. thanks for the extra eyes, @tukusejssirs

TobiTenno:
@tukusejssirs let me know if you're satisfied by this current state.

to explain a bit, the prettier format -> dedupe -> sort would have to be in both the lint-staged cleanup and the dedupe on the check to ensure they both format out the same

tukusejssirs:
At first, I’ve thought you sorted `package*.json` file alphabetically, however, after reading the `README.md` file of `sort-package-json`, I know it does it in a different order. I personally usually sort JSON files alphabetically using `jq -S`, however, I don’t about this much (I always use features of my IDE or `grep` to find the properties/values I want.

However, now the PR name and description does not match the changes made. You seem to prettify and sort `package*.json` files in the CI after deduping them. After vhanging them, I see no reason not to commit this PR.

tukusejssirs (approved):

tukusejssirs:
Oh, and here the auto-merge goes: I wanted you change the PR desc, but not to slow you down by waiting on me to approve the changes … I have forgotten about the auto-merge being enabled.

TobiTenno:
I'm happy to do another, but i'm not sure that I understand what you're asking. I change them in CI after deduping (as mentioned above) so that it's the same treatment they'd get from the commit hook to format -> dedupe -> sort so there'd be no git differences, which would fail the dedupe check

TobiTenno:
I think the dedupe check (specifically the check, not the local part) just may not work out due to how it interferes with dependabot updates, and we have no way of forcing dependabot to dedupe

tukusejssirs:
> I'm happy to do another, but i'm not sure that I understand what you're asking. I change them in CI after deduping (as mentioned above) so that it's the same treatment they'd get from the commit hook to format -> dedupe -> sort so there'd be no git differences, which would fail the dedupe check

Nothing to do anymore. I was suggesting to update the OP which could be used in the MR commit message which is not the case on GitHub when squashing the changes (https://github.com/romcal/romcal/commit/0c9211233ef6e17e777f066dee893ef1766d1cde).

Anyway, I still don’t really like the the auto-merge option being enabled, as whenever I (or someone else) approve a PR, it might still have some nitpicks which I raise without requiring them, thus I let the PR author decide whether they implement them or not. Sometimes another reviewer (e.g. Etienne) might disagree with me, however, the PR would be already merged, thus a revert would be potentially required.

PR #517, opened 2024-04-12
ci: disable max-body-line-len commitlint

TobiTenno:


.commitlintrc.ts
```diff
@@ -39,6 +39,7 @@ export default {
   },
   plugins: ['selective-scope'],
   rules: {
+    'body-max-line-length': [RuleConfigSeverity.Disabled] as const,
```
tukusejssirs, line 42: Can’t it be simply disabled by _removing_ this line or at least by setting it to `[0, 'never', 0]` (I’ve just checked I have it set to this, just like `footer-Mmax-line-length` and `header-max-length`)? :thinking:

tukusejssirs: Is this `RuleConfigSeverity.Disabled` is preferred to setting it to `[0, 'never', 0]`? :thinking:

TobiTenno: well, the default in [`config-conventional`](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/src/index.ts) that we extend has it declared, so we can't just remove the line

and i think how you and I do it are functionally the same i'm just using the types for it

tukusejssirs (approved):

PR #535, opened 2024-04-27
refactor: converting from monolith to monorepo, starting package decomposition

emagnier:
It's been a while since we're talking about monorepos, so let's do it!

This PR is really just a first step. There's still quite a bit of work to modularize things further, but let's save that for future PRs.

Basically, in step 1, the **romcal** library is converted into a package within the monorepo `@romcal/rite-roman1969`. Later on, we can continue to refactor certain elements into different packages (for future PRs). I've already made a first one: `@romcal/easter` (which contains the code to calculate the date of Easter).

I think the key here is to stabilize the current configuration. I've continued to use `npm` to set up this monorepo. I'm more used to using `pnpm` than `npm` in monorepo projects, so I'd appreciate some help fine-tuning the configs and package.json with `npm` (for now, it's pretty basic).

Here's the start of the new structure:
```
packages/
    config/     # contains configuration files for current and future packages, defined once in this package
    easter/     # a first package, as an example
rites/
    roman1962/   # a new package, currently empty, waiting for future contributors ;-)
    roman1969/   # the 'romcal' library, moved here
```

For now, everything remains functional in the same way, using the npm package `romcal` (to avoid breaking changes). 

If you'd like to contribute to this PR as well, feel free to do so 😉


emagnier:
Ping @TobiTenno or @tukusejssirs, I just did a rebase. 

Since this PR was opened in April until now, there have been quite a few changes, so the rebase is starting to get tedious... This PR is a first step in reorganizing this codebase into a monorepo, so indeed, there will still be work to do after this. But it would be good to move forward with this PR, to avoid blocking future development or making the rebases even more tedious.

TobiTenno:
@emagnier i'm down for it assuming we can still release as desired, i'll give it another thorough review this morning and ask for changes or approve in a few hours.

packages/easter/jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('@romcal/config/jest.config'),
+};
```
TobiTenno, line 3:
```suggestion
module.exports = require('@romcal/config/jest.config');
```

TobiTenno:
```suggestion
module.exports = require('@romcal/config/jest.config');
```

rites/roman1969/__tests__/calendar-builder.test.ts
```diff
@@ -1,13 +1,11 @@
...
-import { Slovakia_Sk } from 'romcal/dist/bundles/slovakia';
-
-import { CalendarScope, Id, LiturgicalCalendar, LiturgicalDay, LiturgicalDayDef, Period, Romcal, Season } from '../lib';
-import { dateDifference } from '../lib/utils/dates';
+import { England_En } from '../dist/bundles/england';
+import { GeneralRoman_En } from '../dist/bundles/general-roman';
+import { Germany_En } from '../dist/bundles/germany';
+import { Hungary_En } from '../dist/bundles/hungary';
+import { Ireland_En } from '../dist/bundles/ireland';
+import { Slovakia_Sk } from '../dist/bundles/slovakia';
```
TobiTenno: if we're going to reference these bundle paths so much, i think we should wire up these paths in the tsconfig and resolve them as `@bundles/{calendar}` so that we can avoid quite so many messy paths.

rites/roman1969/__tests__/calendar-builder.test.ts
```diff
@@ -1,13 +1,11 @@
...
+import { CalendarScope, Id, LiturgicalCalendar, LiturgicalDay, LiturgicalDayDef, Period, Romcal, Season } from '../src';
```
TobiTenno: as above, we should alias `src` and `src/utils` unless we're just going to reference `@src` all the time, that way if we move files, we don't have to manually update the relative pathing

rites/roman1969/build/build.ts
```diff
@@ -1,6 +1,9 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable import/no-unused-modules */
```
TobiTenno: these are things we can configure in config/eslintrc, i.e. allowing `devDependencies` in `/build` folders...

ex: 
```js
    'import/no-extraneous-dependencies': [
      'error',
      { devDependencies: ['build/**/*.ts'] },
    ],
```

rites/roman1969/build/bundle.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
TobiTenno: see other comment about this config

rites/roman1969/build/data-checks.ts
```diff
@@ -20,22 +20,24 @@
  * to distinguish them from the keys that are directly defined in the source files.
  */
 
+/* eslint-disable no-console */
+/* eslint-disable import/no-extraneous-dependencies */
```
TobiTenno: see other comment about this type of configuration

no-console could be just ignored in an override in a similar manner in `build` folders

rites/roman1969/build/publish.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-extraneous-dependencies */
+
```
TobiTenno: see other comments about this type of configuration

rites/roman1969/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@romcal/rite-roman1969",
```
TobiTenno: are we also releasing this as the general roman calendar?

tsconfig.base.json
```diff
@@ -0,0 +1,33 @@
...
+    "downlevelIteration": true,
+    "importHelpers": true,
+    "allowJs": true,
+    "baseUrl": ".",
+    "paths": {
+      "@romcal/rite-roman1962": ["./rites/roman1962"],
+      "@romcal/rite-roman1969": ["./rites/roman1969"],
+      "@romcal/config": ["./packages/config"],
+      "@romcal/easter": ["./packages/easter"],
+      "@romcal/package.json": ["./package.json"]
```
TobiTenno: could we consider not having the `@romcal/` prefix, as these will conflict with future package names?

(and add some others i mentioned elsewhere)

i.e.
```suggestion
      "@rite-roman1962": ["./rites/roman1962"],
      "@rite-roman1969": ["./rites/roman1969"],
      "@config": ["./packages/config"],
      "@easter": ["./packages/easter"],
      "@package.json": ["./package.json"]
```

TobiTenno (changes requested):
not massive changes, but some that i think will keep us from having as much boilerplate text and move more to configuration

emagnier:
Thanks @TobiTenno 🙏 

Indeed, I haven’t tested the script/publication flow yet. But everything should work the same way. For now, I’ve simply rewired everything so that the published packages work identically, without any breaking changes (and thus, everything remains transparent for users using romcal through npm packages).

There will probably be some reflection in the future about how romcal's npm packages should be used according to different rites (like `roman1962` or `roman1969`). But I thought we could address this topic later, when we have a clearer picture of the new codebase reorganization, and if another type of liturgical calendar (other than `roman1969`) comes into play.

tukusejssirs:
@emagnier, sorry, I haven’t reviewed this awesome PR yet. I’ll try to find some time this weekend for it. I feel you with the tedious rebases.

rites/roman1969/package.json
emagnier: Not for now, at the moment it's only for internal use. But eventually in the future, it's something to consider.

tsconfig.base.json
emagnier: I like keeping the `@romcal/` prefix to clearly identify that these are romcal-specific packages. And I think that if some of these packages are published someday, we can always adjust the naming without any issues.

Or maybe we should use a different prefix to identify the romcal packages that are published and those that are not.

emagnier:
Thank you very much for your initial feedback. I don't have much free time right now, but I'll try to make progress on this in the next few days. This way, we'll have the first steps in place to keep moving forward on this 🔥.

tsconfig.base.json
TobiTenno: so, problem is that these aren't packages, these are in-type aliases, and the actual packages can be referenced directly in the system by their full package name. I think what you're doing makes sense, but now how you're doing it.

rites/roman1969/package.json
TobiTenno: so, what's going to be releasing as the main general roman calendar? i don't think we want to break that pathway

tsconfig.base.json
TobiTenno: i.e. `@romcal/config` shouldn't be usable outside of this and this might lead someone to believe this will also be a separate stand-alone released package... but given that it's purely internal to the monorepo, this should be an internal config reference, not something I could consume in another project.

rites/roman1969/__tests__/calendar-builder.test.ts
emagnier: I introduced paths alias like `@dist/rite-roman1969/bundles/germany` and `@src/rite-roman1969`. We'll see how it works in practice, or if there are better suggestions.

tsconfig.base.json
emagnier: The issue here is that, packages like `@romcal/rite-roman1962` are actual NPM packages (they have their own `package.json` files), they're just not published on npmjs.org. But for everything to work well in the monorepo, I also had to define them as aliases. Also, the configuration varies slightly depending on the package manager (npm, yarn, pnpm...), and I haven't fully explored how to optimize the setup here (so inputs are welcome here).

For now, I’ve renamed the @romcal/* packages to @internal/* while these packages remain unpublished. If some of them do get published, then we can rename them back to @romcal/*, which should reduce the risk of confusion.

rites/roman1969/package.json
emagnier: For now, it’s the `romcal` package that continues to be published (as well as all the `@romcal/calendar.*`). Nothing more.

To avoid breaking anything, the `romcal` package could become an alias for a future package `@romcal/rite-roman1969` (or vice versa). Or maybe we will do it differently. 
Currently, other types of rites are not yet supported, and I think we still need to think about how to organize (and publish) these packages properly (including all the `@romcal/calendar.*`). So indeed, for now, I’m focusing on preparing the codebase in the direction of multi-rites + greater modularity, without introducing any breaking changes.

package.json
```diff
@@ -48,21 +48,21 @@
...
   "scripts": {
-    "build": "npm run build -w=@romcal/easter && npm run build -w=@romcal/rite-roman1969",
-    "clean": "rimraf coverage dist tmp && npm run clean -w=@romcal/easter && npm run clean -w=@romcal/rite-roman1969",
-    "data:checks": "npm run data:checks -w=@romcal/rite-roman1969",
-    "doc": "npm run doc -w=@romcal/rite-roman1969",
-    "docs:check-links": "npm run docs:check-links -w=@romcal/rite-roman1969",
-    "docs:sort-glossary": "npm run docs:sort-glossary -w=@romcal/rite-roman1969",
-    "lint": "npm run lint -w=@romcal/easter && npm run lint -w=@romcal/rite-roman1969",
-    "lint:fix": "npm run lint -w=@romcal/easter && npm run lint:fix -w=@romcal/rite-roman1969",
+    "build": "npm run build -w=@internal/easter && npm run build -w=@internal/rite-roman1969",
```
TobiTenno: if we're building or cleaning on multiple, they can all be listed, e.g.: `npm run test --workspace=a --workspace=b` ([ref](https://docs.npmjs.com/cli/v10/using-npm/workspaces))

```suggestion
    "build": "npm run build -w=@internal/easter -w=@internal/rite-roman1969",
```

package.json
```diff
@@ -40,85 +40,43 @@
...
-    "lint": "eslint .",
-    "lint:fix": "npm run lint -- --fix",
+    "build": "npm run build -w=@internal/easter && npm run build -w=@internal/rite-roman1969",
+    "clean": "rimraf coverage dist tmp && npm run clean -w=@internal/easter && npm run clean -w=@internal/rite-roman1969",
+    "data:checks": "npm run data:checks -w=@internal/rite-roman1969",
+    "doc": "npm run doc -w=@internal/rite-roman1969",
+    "docs:check-links": "npm run docs:check-links -w=@internal/rite-roman1969",
+    "docs:sort-glossary": "npm run docs:sort-glossary -w=@internal/rite-roman1969",
+    "lint": "npm run lint -w=@internal/easter && npm run lint -w=@internal/rite-roman1969",
+    "lint:fix": "npm run lint -w=@internal/easter && npm run lint:fix -w=@internal/rite-roman1969",
```
TobiTenno: as above here (for running multiple workspaces at once), but it looks like the first command lost the fix flag, so we should have...
```suggestion
    "lint:fix": "npm run lint -w=@internal/easter -w=@internal/rite-roman1969  -- --fix",
```

package.json
```diff
@@ -40,85 +40,43 @@
...
-    "build": "node -r ts-node/register scripts/build.ts",
-    "clean": "rimraf coverage dist tmp",
-    "data:checks": "node -r ts-node/register scripts/data-checks.ts",
-    "doc": "node -r ts-node/register scripts/bundle-doc.ts",
-    "docs:check-links": "node scripts/docs/check-broken-links.mjs ./docs",
-    "docs:sort-glossary": "node scripts/docs/sort-glossary.mjs ./docs",
-    "lint": "eslint .",
-    "lint:fix": "npm run lint -- --fix",
+    "build": "npm run build -w=@internal/easter && npm run build -w=@internal/rite-roman1969",
+    "clean": "rimraf coverage dist tmp && npm run clean -w=@internal/easter && npm run clean -w=@internal/rite-roman1969",
```
TobiTenno:
```suggestion
    "clean": "rimraf coverage dist tmp && npm run clean -w=@internal/easter -w=@internal/rite-roman1969",
```

package.json
```diff
@@ -40,85 +40,43 @@
...
+    "data:checks": "npm run data:checks -w=@internal/rite-roman1969",
+    "doc": "npm run doc -w=@internal/rite-roman1969",
+    "docs:check-links": "npm run docs:check-links -w=@internal/rite-roman1969",
+    "docs:sort-glossary": "npm run docs:sort-glossary -w=@internal/rite-roman1969",
+    "lint": "npm run lint -w=@internal/easter && npm run lint -w=@internal/rite-roman1969",
```
TobiTenno:
```suggestion
    "lint": "npm run lint -w=@internal/easter -w=@internal/rite-roman1969",
```

package.json
```diff
@@ -40,85 +40,43 @@
...
     "prettier:fix": "prettier -w .",
-    "test": "TZ=UTC jest --coverage",
-    "test:snapshot:update": "TZ=UTC jest -u",
-    "test:watch": "TZ=UTC jest --watch",
-    "test:without-coverage": "TZ=UTC jest"
-  },
-  "dependencies": {
-    "i18next": "^21.6.0",
-    "ts-deepmerge": "^7.0.0"
+    "test": "npm run test -w=@internal/easter && npm run test -w=@internal/rite-roman1969",
```
TobiTenno:
```suggestion
    "test": "npm run test -w=@internal/easter -w=@internal/rite-roman1969",
```

rites/roman1962/jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('@internal/config/jest/jest.config'),
+};
```
TobiTenno: no need to spread
```suggestion
module.exports = require('@internal/config/jest/jest.config');
```

rites/roman1962/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@internal/rite-roman1962",
+  "version": "0.0.0",
+  "description": "Liturgical calendar for the Roman Rite according to the 1962 reform",
```
TobiTenno, line 4: So, throwing a wrench, the calendar is technically the 1960 calendar, it was incorporated into the 1962 edition of the Roman Missal. so we probably need to refactor as `roman1960` (i know we don't have any calendars in here yet, but just hoping to keep it accurate)

```suggestion
  "description": "Liturgical calendar representing the General Roman Calendar of 1960",
```

rites/roman1969/build/data-checks.ts
```diff
@@ -23,19 +23,19 @@
...
+import { PROPER_CYCLES, PSALTER_WEEKS, SUNDAY_CYCLES, WEEKDAY_CYCLES } from '../src/constants/cycles';
+import { MONTHS } from '../src/constants/months';
+import { RANKS } from '../src/constants/ranks';
+import { SEASONS } from '../src/constants/seasons';
+import { WEEKDAYS } from '../src/constants/weekdays';
+import { GeneralRoman } from '../src/general-calendar/proper-of-saints';
+import { locales } from '../src/locales';
+import { CalendarDef } from '../src/models/calendar-def';
+import { particularCalendars } from '../src/particular-calendars';
+import { toPackageName } from '../src/utils/string';
```
TobiTenno, line 38: could we have internal aliases in each distinct package for it's src folder so we can simplify these, or for the distinct src directories so they're more easily referenced in a given package?
i.e. in order to produce

```suggestion
import { Locale } from '@src';
import { Martyrology } from '@src/catalog/martyrology';
import { COLORS } from '@src/constants/colors';
import { PROPER_CYCLES, PSALTER_WEEKS, SUNDAY_CYCLES, WEEKDAY_CYCLES } from '@src/constants/cycles';
import { MONTHS } from '@src/constants/months';
import { RANKS } from '@src/constants/ranks';
import { SEASONS } from '@src/constants/seasons';
import { WEEKDAYS } from '@src/constants/weekdays';
import { GeneralRoman } from '@src/general-calendar/proper-of-saints';
import { locales } from '@src/locales';
import { CalendarDef } from '@src/models/calendar-def';
import { particularCalendars } from '@src/particular-calendars';
import { toPackageName } from '@src/utils/string';
```

TobiTenno (changes requested):
- we misnamed GRC 1960 as 1962 ([whoops](https://en.wikipedia.org/wiki/General_Roman_Calendar_of_1960))
- internal aliases inside packages for ease of reference and less shifting of relative references with refactors

rites/roman1962/package.json
emagnier: Thank you for your comment and the historical reminder! The 1962 reform was also much more significant (compared to the one in 1960), and it’s the one we use today as a reference when we talk about the "extraordinary form" of the Roman rite, or the "Tridentine Mass," or the "Traditional Mass." So, I’d be still inclined to keep 1962 as the reference for that reason. What do you think?

rites/roman1969/build/data-checks.ts
emagnier: I think it's a great idea. However, when defining an `@src` alias in the `tsconfig.json` of the current package, it completely overwrites the `paths` property from the parent `tsconfig.json` (instead of merging it). So, we would need to copy the `paths` config from the parent into the sub-package's `tsconfig.json` (in addition to the custom `@src`), which I find unfortunate.

I’m a bit torn between copying some config on all sub-packages to have cleaner paths in the code, or keeping the paths as they are now. But maybe I’m missing something, and there could be a better solution to handle these paths.

TobiTenno: ah, that's a fair problem... i wish we could use js configs instead of json configs :( let's keep it how you have for now, and we can iterate on it later.

TobiTenno (approved):

rites/roman1962/package.json
TobiTenno: i agree with the idea, but we are also calling the current calendar (in code) the `rite-roman1969`, which corresponds to the original promulgation of the calendar, not the updates (i.e. 2002, 2000, or 1975)...

I'm fine with your wording and i think it suffices to assuage my issues 👍

emagnier: Thank you, and since these packages are not released for now, we still have time to test the naming.

PR #536, opened 2024-05-01
feat(philippines): raise the rank of `ezequiel_moreno_bishop`

tukusejssirs:
According to [this article](https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/), the celebration rank was raised to an obligatory memorial in the Philippines in late January 2024.

I’ve learnt about this change via an [update](https://en.wikipedia.org/w/index.php?title=National_calendars_of_the_Roman_Rite&diff=prev&oldid=1221643797) to the Wikipedia article of [National calendars of the Roman Rite](https://en.wikipedia.org/wiki/National_calendars_of_the_Roman_Rite).

### DIFF
--- a/lib/particular-calendars/philippines.ts
+++ b/lib/particular-calendars/philippines.ts
@@ -50,7 +50,17 @@ export class Philippines extends CalendarDef {
     },
 
     ezequiel_moreno_bishop: {
-      precedence: Precedences.OptionalMemorial_12,
+      /*
+       * src: https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/;
+       * old:
+       *   rank: Precedences.OptionalMemorial_12
+       *   since: unknown
+       *   until: 28 Jan 2024
+       * new:
+       *   rank: Precedences.ProperMemorial_11b
+       *   since: 29 Jan 2024
+       */
+      precedence: Precedences.ProperMemorial_11b,
       dateDef: { month: 8, date: 19 },
     },
 


lib/particular-calendars/philippines.ts
```diff
@@ -50,7 +50,7 @@ export class Philippines extends CalendarDef {
     },
 
     ezequiel_moreno_bishop: {
-      precedence: Precedences.OptionalMemorial_12,
+      precedence: Precedences.ProperMemorial_11b, // src: https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/; since: 29 Jan 2024; until then it was an optional memorial
```
TobiTenno: proposal: 
```suggestion
      /*
        src: https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/;
        since 29 Jan 2024; until then it was an optional memorial
       */
      precedence: Precedences.ProperMemorial_11b,
```

not really a substantive change in my comment, this would just help avoid extra long lines

tukusejssirs: I really don’t care about the comment format here, as I consider them as notes for the fure future source and reference objects.

That said, until now, we added only `src` and `based on` comments after the particular l10n key value. Here, I have added a note of the previous rank and when it was changed.

We could also use a JSON string comment instead, however, as I have already said, it does not matter to me much … until we add actual sources and references into romcal.

TobiTenno: yes, i'm not debating what the content should be, precisely, just that a block comment is better for reading and formatting this long text than a line comment after the content

tukusejssirs: Okay, I’ll accept your suggestion. :+1:

tukusejssirs: I have changed it to a YAML within a block comment. I hope you are okay with this. :wink:

```ts
      /*
        src: https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/;
        old:
          rank: Precedences.OptionalMemorial_12
          since: unknown
          until: 28 Jan 2024
        new:
          rank: Precedences.ProperMemorial_11b
          since: 29 Jan 2024
       */
```

tukusejssirs: Actually, I am thinking about changing it to this:

```ts
      /*
       * src: https://recoletos.ph/2024/02/01/cbcp-elevates-st-ezekiel-moreno-memorial-as-obligatory/;
       * old:
       *   rank: Precedences.OptionalMemorial_12
       *   since: unknown
       *   until: 28 Jan 2024
       * new:
       *   rank: Precedences.ProperMemorial_11b
       *   since: 29 Jan 2024
       */
```

tukusejssirs: Changed.

TobiTenno (approved):

TobiTenno:
@tukusejssirs @emagnier i think we'll have to change some settings for PRs to push version updates back into the branch. I believe the merge from this PR & 1a244c86a21432b098a8dac42e595b208d11790f will end up with the same failing state of not being able to push to the branch, due to 
image

because this rule conflicts with
image

tukusejssirs:
> @tukusejssirs @emagnier i think we'll have to change some settings for PRs to push version updates back into the branch. I believe the merge from this PR & https://github.com/romcal/romcal/commit/1a244c86a21432b098a8dac42e595b208d11790f will end up with the same failing state of not being able to push to the branch

Do you see a way out of this situation? :thinking:

While I think we should have push changes without a PR, we still need to allow pushing changes from GitHub CI. :thinking:

TobiTenno:
[re: @tukusejssirs]

> > @tukusejssirs @emagnier i think we'll have to change some settings for PRs to push version updates back into the branch. I believe the merge from this PR & [1a244c8](https://github.com/romcal/romcal/commit/1a244c86a21432b098a8dac42e595b208d11790f) will end up with the same failing state of not being able to push to the branch
> 
> Do you see a way out of this situation? 🤔
> 
> While I think we should have push changes without a PR, we still need to allow pushing changes from GitHub CI. 🤔

i think our current settings will allow for this, as we only currently allow @emagnier and @romcalproject to bypass any merge requirements, one of which is still PRs... I don't think we're at risk if we remove it.

For now, i'm going to turn it off, and we can keep chatting on it if we can find a way around. In my attempts with semantic-release and similar (like how we do), i've never found a way with that setting on

PR #438, opened 2023-05-12
fix(util): fix `toScreamingSnakeCase()` function

tukusejssirs:
At the same time, create a test for the function.

PR #538, opened 2024-05-01
style(commitlint): add additional Commitlint scopes

tukusejssirs:


tukusejssirs:
I have added `util` Commitlint scope, as I think we should a scope for stuff related to neither calendars not l10n. If you think it should have a different name, feel free to suggest a different one.

This would be also used for the commit message of #438.

Moreover, I am not 100 % sure what Commitlint type and scope to use for this PR. I’d personally use `chore` type with no scope, however, @TobiTenno didn’t want to add that type, thus I chore `style`. Tell me if we want to use a different one for changing Commitlint configuration.

Thanks in advance! :pray:

tukusejssirs:
Moreover, the `commitlint` PR check fails, as ATM we don’t allow `style` without scope.

ATM `style` must have scope (one of `[...CALENDAR_IDS, ...LOCALE_IDS, 'calendar', 'l10n', 'util']`; `util` being added here). I suggest to add the following scopes for `style`:

- `commitlint`;
- `editorconfig`;
- `eslint`;
- `prettier`.

These scopes would be used in commit messages where the changes are only related to the configuration of a particular tool.

WDYT, @TobiTenno?

.commitlintrc.ts
TobiTenno, line 1: for this PR to pass/merge, you need to add `null` to `style`.

I think most of the rest are self-explanatory. could you specify what `'util'` should apply to as a scope?

TobiTenno: Style applies to this PR very nicely.

Allowing null scopes seems tricky because we are trying to avoid people using style lazily without telling us what it applies to.

In my opinion, those additional scopes make sense for `style`, but I'm not sure that I would use `util` as a "catch-all" scope

tukusejssirs: > for this PR to pass/merge, you need to add `null` to `style`.

I get this, however, I wanted to discuss this further, possible to revise the scope of this PR to update the Commitlint configuration more. :man_shrugging:

> could you specify what 'util' should apply to as a scope?

I’d use it for changes to utilities, i.e. changes that are not related to calendars or l10n or tests or docs. Just look at #438: I fixes a function (a utility) which related to neither calendars or l10n.

> Allowing null scopes seems tricky because we are trying to avoid people using style lazily without telling us what it applies to.

Okay, thus no additional empty scopes allowed. :+1:

> but I'm not sure that I would use util as a "catch-all" scope

If you want to suggest a different scope name, just tell me. However, there is no scope for changes not related to calendars/l10n/tests/docs yet IMO.

TobiTenno: I'm down for this if we restrict it to changes in the `utils` folder

tukusejssirs: > I'm down for this if we restrict it to changes in the utils folder

I’m okay with this. I believe there is no way to configure Commitlint to restrict the usage of a type to a folder (or is there?), thus we will need to pay attention to this.

I think I need to add a comment on this into the Commitlint config.

TobiTenno: i'll go update that for you so we can roll this through. i'll (likewise) add the discussed allowed scopes

TobiTenno (approved):

Issue #201, opened 2020-07-17
Titles definition within the calendar definition


tukusejssirs:
Related: #48

---

Currently, the patron titles (like `patron of Europe`) are broken because we rely on the localised key and not on the calendar definition. The very saint can be celebration as patron in one country, but not in the other.

I suggest we remove all patron saints from from the l10n files and define them in the respective particular calendar.

As there are many saints that are patrons of whole continents, I think we should create _continental calendars_ (like `europe.ts`) which would contain only the continental patron saints. In the continental calendars, there should be listed all countries that are on that particular continent OR the particular calendars should contain a variable/key that would define on which continent they are, and then we would use these continental calendars in the calendar inheritence (see #48).

---

Example how it could look like from the celebration definition point of view. I have already shared this example [here](https://github.com/romcal/romcal/issues/175#issuecomment-675938978) and [here](https://github.com/romcal/romcal/issues/175#issuecomment-676575446), but as one celebration can have multiple celebration titles, now I suggest that the `patronage` element should be an array.

<details>
  <summary>patronage titles definition in celebration definition</summary>

```js
// This would output `principal patrons of Canada`
metadata: {
	titles: [
		martyr,
	],
	patronage: [
		{
			territory: 'Canada',  // Any string
			level: 'principal',
		},
	],
	sex: 'male',  // male | female
	number: 'plural',  // singular | plural
	capitalize: false,  // `true` should be used for Our Lady or Jesus only; when true, `Patron`/`Patroness` should be written with capital `P`; there should be a romcal option to override this (i.e. permanently set to `false`) per locale, because e.g. in Slovak we don’t capitalise the first letter
}

// This would output `patroness of Italy and Europe`
metadata: {
	titles: [
		martyr,
	],
	patronage: [
		{
			territory: 'Italy',
		},
		{
			territory: 'Europe',
		},
	],
	sex: 'female',
	number: 'singular',
	capitalize: false,
}
```
</details>

I think that the implementation should be something like this:

1. Based on `level`, it would use either `principal` or `secondary` or nothing (empty string);

2. Based on the `sex` and `number`, it would use `patron`, `patrons`, `patroness` or `patronesses`;

3. Based on the `capitalize`, it would change the letter of the `patron` (either sex and either number) to `Patron`;

4. Based on the `territory`, it would add `of {{territory}}`.

[...]

Issue #154, opened 2020-01-26
Separation of localisation keys into separate metadata


tukusejssirs:
Related #26 #12 #24 #3 #43

---

### Tasks (current; assigned to @tukusejssirs) <a name="tasks_ts"></a>

- [x] fill `original key name` column:
   - where there are multiple keys, they are comma-space-separated;
   - `abachum`, `anastasius`, `audifax`, `emerentiana`, `hyginus`, `marcellusI`, `marius`, `martina`, `paulTheFirstHermit`, `peterNolasco`, `prisca`, `teophorus` are currently not in romcal, as they was celebrated in `general` before 1969; it was added as a minor work on the Extra-Ordinary Form calendar;
- [x] lowercase all non-proper words:
   - [x] saint/blessed;
   - [x] titles;
- [x] change number `1000` to `many`;
- [ ] merge `firstName` and `additionalFirstName` columns and `surname` and `additionalSurname` and separate the values by comma-space (`, `);
- [ ] separate `notes` column into:
   - [ ] `alternativeName` and `alternativeSurname`;
   - [x] `name distinguisher` used to add an additional part to the key name to differentiate the keys from each other;
   - [ ] `originalName` and `originalSurname` (name/surname in original language and form; this might be an array or string; this should be protected and nobody should be able to override the value unless done in the sanctoral);
   - [ ] `collectiveName` (usual name for the particular person(s); like `Vietnamese Martyrs`; this might be unnecessary);
- [x] change the dates format to `Jan 1` (or `Jan 1, 2020`);
- [x] corrected some data (like `Alponsus` instead of `Alphonsus`);
- [x] merge `events` sheet into `persons`;
- [x] rename `persons` sheet to `sanctoral`;
- [x] remove unnecessary columns:
   - [x] `not celebrated in`;
   - [x] `celebrated in`;
   - [x] `celebration date (general calendar)`;
   - [x] `general or national calendar?`;
   - [x] `celebration date (cal1962 or EOC)`;
   - [x] `celebration date (cal1969 or GRC)`;
- [x] fill the `name distinguisher` when required;

### Tasks for all locales

These tasks should be done separately for each and every locale.

- [ ] separate the localisation keys of a particular locale into separate metadata;
- [ ] save the metadata values into DB or `.po` files (or whatever else according to the chosen localisation tool);

### Tasks done only once

These tasks are to be done only once and are (mostly) connected with the processing of the localisation data.

- [ ] create auxiliary functions that would process the localisation data;
[...]

PR #546, opened 2024-05-05
enh(costa-rica): our lady of angels

TobiTenno:
our lady of angels, split from #459

### DIFF
--- a/lib/locales/en.ts
+++ b/lib/locales/en.ts
@@ -790,11 +790,7 @@ export const locale: Locale = {
     our_lady_mother_of_divine_providence_patroness_of_puerto_rico:
       'Our Lady, Mother of Divine Providence, Patroness of Puerto Rico',
     /*
-     * src: https://gcatholic.org/calendar/2024/CR-en.htm
-     */
-    our_lady_of_angels: 'Our Lady of Angels',
-    /*
-     * src: @TobiTenno direct translation of the Spanish title
+     * src: https://gcatholic.org/calendar/2024/CR-en.htm + @TobiTenno direct translation of the Spanish title
      */
     our_lady_of_angels_patroness_of_costa_rica: 'Our Lady of Angels, Patroness of Costa Rica',
     our_lady_of_aparecida_patroness_of_brazil: 'Our Lady of Aparecida, Patroness of Brazil',
--- a/lib/locales/en.ts
+++ b/lib/locales/en.ts
@@ -793,6 +793,10 @@ export const locale: Locale = {
      * src: https://gcatholic.org/calendar/2024/CR-en.htm
      */
     our_lady_of_angels: 'Our Lady of Angels',
+    /*
+     * src: @TobiTenno direct translation of the Spanish title
+     */
+    our_lady_of_angels_patroness_of_costa_rica: 'Our Lady of Angels, Patroness of Costa Rica',
     our_lady_of_aparecida_patroness_of_brazil: 'Our Lady of Aparecida, Patroness of Brazil',
     our_lady_of_bethlehem: 'Our Lady of Bethlehem',
     our_lady_of_china: 'Our Lady of China',
--- a/lib/locales/en.ts
+++ b/lib/locales/en.ts
@@ -789,6 +789,10 @@ export const locale: Locale = {
     our_lady_mother_of_christian_unity: 'Our Lady, Mother of Christian Unity',
     our_lady_mother_of_divine_providence_patroness_of_puerto_rico:
       'Our Lady, Mother of Divine Providence, Patroness of Puerto Rico',
+    /*
+     * src: https://gcatholic.org/calendar/2024/CR-en.htm
+     */
+    our_lady_of_angels: 'Our Lady of Angels',
     our_lady_of_aparecida_patroness_of_brazil: 'Our Lady of Aparecida, Patroness of Brazil',
     our_lady_of_bethlehem: 'Our Lady of Bethlehem',
     our_lady_of_china: 'Our Lady of China',
--- a/lib/locales/es.ts
+++ b/lib/locales/es.ts
@@ -478,7 +478,8 @@ export const locale: Locale = {
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
     /*
-     * src: https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96
+     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
+     * page: 96
      */
     our_lady_of_angels_patroness_of_costa_rica: 'Nuestra Señora de los Ángeles, patrona de Costa Rica',
 
--- a/lib/locales/es.ts
+++ b/lib/locales/es.ts
@@ -477,6 +477,15 @@ export const locale: Locale = {
     odile_of_alsace_abbess_patroness_of_alsace: 'Santa Odilia, abadesa, Patrona de Alsacia',
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
+    /*
+     * src: https://gcatholic.org/calendar/2024/CR-en.htm
+     */
+    our_lady_of_angels: 'Nuestra Señora de los Ángeles',
+    /*
+     * src: https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96
+     */
+    our_lady_of_angels_patroness_of_costa_rica: 'Nuestra Señora de los Ángeles, patrona de Costa Rica',
+
     our_lady_of_aparecida_patroness_of_brazil: 'Nuestra Señora de Aparecida, Patrona de Brasil',
     our_lady_of_bethlehem: 'Nuestra Señora de Belén',
     our_lady_of_fatima: 'Nuestra Señora de Fátima',
--- a/lib/locales/es.ts
+++ b/lib/locales/es.ts
@@ -478,10 +478,6 @@ export const locale: Locale = {
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
     /*
-     * src: https://gcatholic.org/calendar/2024/CR-en.htm
-     */
-    our_lady_of_angels: 'Nuestra Señora de los Ángeles',
-    /*
      * src: https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96
      */
     our_lady_of_angels_patroness_of_costa_rica: 'Nuestra Señora de los Ángeles, patrona de Costa Rica',
--- a/lib/locales/es.ts
+++ b/lib/locales/es.ts
@@ -478,6 +478,7 @@ export const locale: Locale = {
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
     /*
+     * Ordo Temporis 2023-2024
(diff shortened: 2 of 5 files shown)


tukusejssirs:
We need to find a source for this celebration before we merge into romcal.

TobiTenno:
i've found general references to it being the largest national solemnity of the year, but i've not found an ordo or similar reference.

[this](https://gcatholic.org/calendar/2024/CR-zjos0-en.htm) is the best overall reference for a calendar i've found, and i can't find anything else with its sources

TobiTenno:
@tukusejssirs this is the closest thing to a decent source i'm finding

tukusejssirs:
@TobiTenno, I’ve found a better source via the [website of the Episcopical Conference of Costa Rica](https://www.iglesiacr.org/comisiones/) (at the bottom of the section named `Comisión Nacional de Liturgia`, there is a link to `Ordo Temporis 2023-2024`): https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96

Based on this source, could you add the Spanish name of the celebration to `es.ts` please?

IMO we should refer to this source in `costa-rica.ts` and `es.ts`, but not in `en.ts` where we could keep the gCatholic link.

BTW, I have added this source to [my gDrive](https://drive.google.com/drive/folders/1TMxeHcZrpCK74wt_8wYcNvqBKuDZnheu).

Thanks! :pray:

TobiTenno:
happily! thanks for finding it, @tukusejssirs

TobiTenno:
Given the separation of feasts and the elevation in Costa Rica for a regional patroness, i'm splitting the localization defs between the two parts. let me know if i'm doing this right!

lib/locales/en.ts
```diff
@@ -789,6 +789,14 @@ export const locale: Locale = {
     our_lady_mother_of_christian_unity: 'Our Lady, Mother of Christian Unity',
     our_lady_mother_of_divine_providence_patroness_of_puerto_rico:
       'Our Lady, Mother of Divine Providence, Patroness of Puerto Rico',
+    /*
+     * src: https://gcatholic.org/calendar/2024/CR-en.htm
+     */
+    our_lady_of_angels: 'Our Lady of Angels',
```
tukusejssirs: This should be removed, as it is not used in any calendar in romcal.

```suggestion
```

lib/locales/es.ts
```diff
@@ -477,6 +477,15 @@ export const locale: Locale = {
     odile_of_alsace_abbess_patroness_of_alsace: 'Santa Odilia, abadesa, Patrona de Alsacia',
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
+    /*
+     * src: https://gcatholic.org/calendar/2024/CR-en.htm
+     */
+    our_lady_of_angels: 'Nuestra Señora de los Ángeles',
```
tukusejssirs: This should be removed, as it is not used in any calendar in romcal.

```suggestion
```

tukusejssirs (changes requested):
> Given the separation of feasts and the elevation in Costa Rica for a regional patroness, i'm splitting the localization defs between the two parts. let me know if i'm doing this right!

I have forgotten about the patronage patronage in the l10n key name! :man_facepalming:

Well, we include the patronage title (or how do you call it? I want to differentiate it from regular titles) in the l10n key name. Therefore, the new key name should be `our_lady_of_angels_patroness_of_costa_rica` (this you have done correctly). However, `our_lady_of_angels` should be kept only if it is used; as this celebration is not used in romcal (for now) without the patronage title, it should be removed from the l10n files (at least this is how we did it until now; it won’t be the case after we generate the celebration names from metadata, see #154, as well as after we define all titles in celebrations in calendar files, see #201). `our_lady_of_angels` should be kept in `martyrology.ts` and `costa-rica.ts` (as you have already done it :+1:).

Moreover, I am not sure if the English celebration name in English should include a definite article or not (`Our Lady of [the] Angels`). I am not sure if She is venerated of some specific angels or of all angels. You might want to read on this. :thinking: I believe you told me you understand a bit of Spanish, probably more than me. :wink:

lib/locales/en.ts
tukusejssirs: That said, the `data:checks` script [passes](https://github.com/romcal/romcal/actions/runs/9568581543/job/26379148706?pr=546#step:9:12), which is strange to me.

Should it not raise a warning/error that `our_lady_of_angels` is not used? :thinking: Do the celebration IDs in calendars make these celebrations consider to be used? If so, we should update the `data:checks` script to:

- include all celebration IDs (the props of the `CalendarDef#inputs` object);
- exclude IDs of those celebration which have `customLocaleId` defined → then include the l10n ID which is set as the values of `customLocaleId`.

/cc @emagnier

TobiTenno: no, it is used, it's just overridden in the one place it is technically used

lib/locales/es.ts
TobiTenno: but it's the main l10n key, so i believe this would cause issues (eventually)

TobiTenno: imo we should remove the necessity of custom l10n keys for title amendments and those titles should be localized, rather than doubling those translations

lib/locales/en.ts
TobiTenno: re: definite article in the name, yes. Both because the source uses "the angels" and the spanish literal uses `los Angeles` not `de Angeles`.

from the theological standpoint, doubly so. while we don't know why the Fallen angels fell, we know it had something to do with the Incarnation, to which the Theotokos is essential.

TobiTenno:
@tukusejssirs I think i've got it all updated, let me know if i missed something

lib/locales/en.ts
tukusejssirs: Re definite article: I’ve read a bit about it. The actual (unofficial?) venerated title of Our Lady is _La Negrita_ (the Black One or Black Lady), however, the celebration name is actually coming from Porciuncula (now the Basilica of Our Lady of the Angels / Santa Maria degli Angeli) in Italy, via fransciscans.

I believe that Our lady is the Queen of _all_ (good/non-fallen) angels. I believe that while in English, we usually use the definite article which concreate stuff (among other things), we also use it when we talk about _all angels_. I could be wrong though. :wink:

tukusejssirs:
Indeed you have updated everything.

I have just noticed that the source link does not work (anymore?) and requires me to log in with my Microsoft account. I think we should replace it with [this link](https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR) (the one shared by the Costa Rican Episcopal Conference) which works without any issues, thus I think we should prefer it.

One downside is that we cannot append `page=96`, the I’d suggest to append `, p96` after this link itself.

lib/locales/en.ts
TobiTenno: Hmmm, to me the distinction when the noun is pluralized (of Angels/of the Angels) is moot, given its already applying to the plural. They would, in my reading, mean the same thing. A parallel is the Latin term `Panis Angelicus` denotes, to an English translation, Bread of Angels and Bread of the Angels equivalently.

tukusejssirs: Indeed, the distiction is minimal and depends much on the context and the person who uses a particular phrase.

> A parallel is the Latin term `Panis Angelicus` denotes, to an English translation, Bread of Angels and Bread of the Angels equivalently.

While I get what you mean, I think there is no such _explicit_ distinction in Latin (as there is no in-/definite article in Latin), thus the parallel is in the _translation_ level only IMHO. That said, you are a native English speaker, not me. :wink:

lib/locales/es.ts
```diff
@@ -477,6 +477,11 @@ export const locale: Locale = {
     odile_of_alsace_abbess_patroness_of_alsace: 'Santa Odilia, abadesa, Patrona de Alsacia',
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
+    /*
+     * src: https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96
```
TobiTenno:
```suggestion
     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
     * page: 96
```

lib/particular-calendars/costa-rica.ts
```diff
@@ -14,5 +15,16 @@ export class CostaRica extends CalendarDef {
       dateDef: { dateFn: 'pentecostSunday', addDay: 4 },
       properCycle: ProperCycles.ProperOfTime,
     },
+
+    our_lady_of_angels: {
+      precedence: Precedences.GeneralSolemnity_3,
+      /*
+       * Ordo Temporis 2023-2024
+       * src: https://lacecor-my.sharepoint.com/personal/repositorioweb_iglesiacr_org/_layouts/15/download.aspx?UniqueId=c6e7f2ef-edfa-445d-8a73-b640bc704bc8#page=96
```
TobiTenno:
```suggestion
     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
     * page: 96
```

TobiTenno:
@tukusejssirs done

lib/particular-calendars/costa-rica.ts
```diff
@@ -14,5 +15,17 @@ export class CostaRica extends CalendarDef {
       dateDef: { dateFn: 'pentecostSunday', addDay: 4 },
       properCycle: ProperCycles.ProperOfTime,
     },
+
+    our_lady_of_angels: {
+      precedence: Precedences.GeneralSolemnity_3,
+      /*
+       * Ordo Temporis 2023-2024
```
tukusejssirs, line 22: I am not against of adding the source title here, however, you didn’t add it in the other place.

lib/particular-calendars/costa-rica.ts
```diff
@@ -14,5 +15,17 @@ export class CostaRica extends CalendarDef {
...
+     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
+     * page: 96
```
tukusejssirs:
```suggestion
       * Ordo Temporis 2023-2024
       * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
       * page: 96
```

tukusejssirs (changes requested):

lib/particular-calendars/costa-rica.ts
TobiTenno: sure, i'm trying to make a better structure to make our sources clear without having to necessarily go into them so that if the link goes down someday, we know what we were looking at

TobiTenno: i've added it to the other.

lib/particular-calendars/costa-rica.ts
```diff
@@ -14,5 +15,18 @@ export class CostaRica extends CalendarDef {
       dateDef: { dateFn: 'pentecostSunday', addDay: 4 },
       properCycle: ProperCycles.ProperOfTime,
     },
+
+    our_lady_of_angels: {
+      precedence: Precedences.GeneralSolemnity_3,
+      /*
+       * Ordo Temporis 2023-2024
+       * Ordo Temporis 2023-2024
```
TobiTenno:
```suggestion
```

lib/locales/es.ts
```diff
@@ -477,6 +477,12 @@ export const locale: Locale = {
     odile_of_alsace_abbess_patroness_of_alsace: 'Santa Odilia, abadesa, Patrona de Alsacia',
     olga_of_kiev: 'Santa Olga',
     our_lady_help_of_christians: 'Nuestra Señora, Auxilio de los Cristianos',
+    /*
+     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
```
TobiTenno, line 482:
```suggestion
     * Ordo Temporis 2023-2024
     * src: https://lacecor-my.sharepoint.com/:b:/g/personal/repositorioweb_iglesiacr_org/Ee_y58b67V1EinO2QLxwS8gBb3zFtlgpRFG0_KBU9l_kCQ?e=TVe9YR
```

TobiTenno:
@tukusejssirs done (2/?)

tukusejssirs (approved):

Issue #511, opened 2024-04-01
holy_thursday wrong color, purple, instead of white.


miguelms95:
The Holy Thursday (holy_thursday) color must be white, and in romcal returns purple.

Source: https://en.wikipedia.org/wiki/Liturgical_colours#Roman_Catholic_Church

tukusejssirs:
Good catch, @miguelms95! :pray:

That said, `purple` is the liturgical colour of the Chrism Mass, while `white` is assigned to the evening Mass.

Now, I see only one way to fix this: generate two liturgical celebrations (array items) for the day (i.e. for the year 2024, `romcal['2024-03-28'].length` would return `2`). I think that is reasonable.

IMHO, we should split Holy Saturday exactly the same way:
1. the first array item would be `holy_saturday` on which there is no Mass celebrated;
2. the second array item would be used for the Easter Vigil.

And if @TobiTenno and @emagnier agree with me, I’d even further: add additional items for vigils of important celebrations (usually solemnities) which have special vigil rules (e.g. first evening prayer / vespers in the Liturgy of the Hours). We will definitely need that in order to provide some data suggested in #20.

WDYT, @TobiTenno and @emagnier? :thinking:

TobiTenno:
i think that makes sense

emagnier:
Indeed, for Holy Thursday, Romcal already provides 2 elements (2 objects in the array) for these 2 consecutive liturgical times (on the same civil day): Holy Thursday (purple), which is the last day of Lent, and Thursday of the Lord’s Supper (white), marking the beginning of the Easter Triduum.

Regarding Holy Saturday, I am not entirely convinced by the suggestion. This is because the vigil office and/or Mass are part of the same liturgical day, which is Easter Sunday, starting at nightfall. Traditionally in the church, the liturgical "day" began at nightfall and lasted until the next evening. Today this tradition is still followed for some major celebrations (like Christmas, Easter Sunday, Pentecost Sunday), as well as in monastic offices. From a data perspective, the "liturgical day" entity should remain a single object. For instance, Easter Sunday (which includes the Vigil, the Easter Sunday Mass, etc.) should be considered as one single object, not multiple.

To address the need to determine whether for a given civil day "A," the vigils of the following liturgical day "B" can be celebrated, I would suggest adding a property to the "A" object indicating that the vigils of the liturgical day "B" can also be celebrated. For example, on `holy_saturday`, we could have a property like `precedingVigil: [ "easter_sunday" ]`.

[...]

Issue #297, opened 2021-09-29
Add a `subtitle` property to `LiturgicalDay` object


tukusejssirs:
As talked about in #295, we should add a property of `subtitle` to `LiturgicalDay` object, which could be used for alternative celebration names, for example for:
- Sunday of the Word of God (3rd Sunday of Ordinary Time);
- Sunday of the Divine Mercy (2nd Sunday of Easter);
- Mission Sunday (second last Sunday of October);
- Good Shepherd Sunday (5th Sunday of Lent);
- Commemoration of the Lord’s Supper (Holy/Maundy Thursday);
- Eighth Day within the Octave of the Nativity of the Lord (1 January):
  - I have no idea if `mary_mother_of_god` key value should be the celebration name and `Eighth Day within the Octave of the Nativity of the Lord` subtitle, or vice versa.

The main/standard celebration name would be the general name of the celebration (e.g. `nth Sunday in Ordinary Time` and key name would be based on it.

emagnier:
Thank you for the feature request. But we should discuss before adding it to a milestone, especially that I'm the only one developer on this project (unfortunately) and I can't do yet miracle 😄

tukusejssirs:
I’m sorry for being to eager. :wink:

tukusejssirs:
I’ve updated the issue description to include the foollowing two celebrations:

- Mission Sunday (second last Sunday of October);
- Good Shepherd Sunday (5th Sunday of Lent).

I’m sure the list is not complete, but it might be helpful when you’ll work on this feature.

TobiTenno:
@emagnier @tukusejssirs i'd like to start working on some of the blockers. Is there any current objection to this? I'd like to pick back up on some of these now that my busy summer is over.

emagnier:
I think this one is a good candidate.

What’s your take on this? Currently, the localization files are mainly key-value pairs, with the values being strings used to name the corresponding liturgical day definition. Instead of being a string, it should probably be an object. Maybe something like this?
```
"liturgical_day_key": {
  title: "The liturgical day name", // or leave it undefined if this is generated automatically, e.g.: 2nd Sunday of Easter
  subtitle: "The liturgical day subtitle" // the optional subtitle
}
```

To push the ideas a bit further (this is off-topic here, but just to give a vision of how things could evolve later, maybe to guide the initial choices), I would consider merging the metadata from the martyrology file with the English locale file [...].
[...]

Issue #20, opened 2018-03-07
feature request: ‘Liturgical-calendar-style’ output


tukusejssirs:
First of all, the ‘direktórium’ (which I mentioned in [this](https://github.com/pejulian/romcal/issues/10#issuecomment-370244697) comment issue #10) can be anglicised as ‘directory’, but in America it is probably simply called as ‘liturgical calendar’ ([src](http://www.usccb.org/about/divine-worship/liturgical-calendar/index.cfm)). Otherwise, see [this](https://en.m.wikipedia.org/wiki/Directorium) definition of ‘directorium.’

What I mean in this feature request is to add some more information to the feast/memorials/etc. I know that this would require some time to code, it might be a milestone to v2.0.0.

Here are my suggestions / feature requests:

- add Mass readings (1st & 2nd readings, psalm and gospel) coordinates
- add the text of the readings (I knowm there might be some troubles with copyright holders)
- add an info if Gloria and/or Credo is said that day during the Holy Mass
- add some notes (e.g. if a Sunday/solemnuty/etc preceded a memorial, output here that ‘A <type> <key> will not be celebrated this year.’
- add Holy Mass prayers (Collect Prayer, Prayer above the Offerings, Prayer after Communion) – for now at least from where it is (like from 3rd ord. Sunday), later the might be good too.

agbloomfield:
This is an excellent addition, and part of another project I'm working on.

tukusejssirs:
> another project
What project do you mean? You mean one on `romcal` or something else? :)

agbloomfield:
Another project -- can you email me at agbloomfield <at> gmail <dot> com -- I'd like to share the history with you.

tukusejssirs:
I have just found out that what is called _Direktórium_ in Slovak, it is called _Ordo_ in Latin. I have found two proofs of that:
- _Ordo 2020_ (for GRC) can be ordered for example [here](https://www.libreriadelsanto.it/libri/9788826602868/ordo-missae-celebrandae-et-divini-officii-persolvendi.html) (full name: _Ordo missæ celebrandæ et divini officii persolvendi – Secundum calendarium romanum generale pro anno liturgico 2019-2020_);
- _Ordo 1904_ for Ireland (Irish dioceses) can be found [here](https://archive.org/details/OrdoDiviniOfficii1904) (full name: _Ordo divini officii recitandi missæque celebrandæ in annum MDCCCCIV – A singulis episcopis Hibernicis approbatus_).

[...]

PR #550, opened 2024-05-06
enh(calendar): separate easter vigil

### LINKED ISSUE #199: Separate `Easter Vigil` from `Holy Saturday`
Athough both Easter Vigil and Holy Saturday fall on the same day, I think the l10n key should be separated into two parts, or maybe the `Easter Vigil` parts should be removed altogether.

TobiTenno:
closes #199


lib/locales/en.ts
```diff
@@ -378,6 +378,8 @@ export const locale: Locale = {
     dunstan_of_canterbury_bishop: 'Saint Dunstan, Bishop',
     dyfrig_of_wales_bishop: 'Saint Dyfrig, Bishop',
     easter_sunday: 'Easter Sunday of the Resurrection of the Lord',
+    // TODO: ref needed, checking exact name
+    easter_vigil: 'Easter Vigil',
```
tukusejssirs: What reference is needed? Do you mean a _source_ reference when the key value can be found?

---

In [`mr_en_2011_ed3_us`](https://drive.google.com/file/d/12aeIw1559RyEPlBpGf6_o6Me5NxqjoLR/view?usp=drive_link) (private link) on p339 (PDF page, don’t look at the printed page numbers), the name of the Sunday is `Easter Sunday of the Resurrection of the Lord`. Next on p341, we have `The Easter Vigil in the Holy Night`.

Now IMHO, a celebration name should be the same regardless whether it is celebrated in the eve of the solemnity or on the day. I think the vigil name (if any) should be added as a celebration subtitle (see #297). Note that in the future we might need to differentiate between celebration subtitles (as described in #297) and Mass formulars (which we didn’t implemented yet), and the Masses `in the vigil`/`at dawn`/`during the day` are actually Mass formulars of a particular celebration.

This needs to be further discussed, however, IMHO we should implement either #297 (which partially solves this issue) or Mass formulars (related to #20; this might be better suited to fix this particular problem).

lib/locales/en.ts
```diff
@@ -498,7 +500,7 @@ export const locale: Locale = {
     holy_family_of_jesus_mary_and_joseph: 'The Holy Family of Jesus, Mary and Joseph',
     holy_guardian_angels: 'The Holy Guardian Angels',
     holy_innocents_martyrs: 'The Holy Innocents, Martyrs',
-    holy_saturday: 'Holy Saturday/Easter Vigil',
+    holy_saturday: 'Holy Saturday',
```
tukusejssirs:
```suggestion
    holy_saturday: 'Holy Saturday', // src: mr_en_2011_ed3_us, Holy Saturday, p339
```

lib/locales/pl.ts
```diff
@@ -233,6 +233,8 @@ export const locale: Locale = {
     divine_mercy_sunday: '2 Niedziela Wielkanocna czyli Niedziela Miłosierdzia Bożego',
     dominic_de_guzman_priest: 'Św. Dominika, prezbitera',
     easter_sunday: 'Niedziela Wielkanocy Zmartwychwstania Pańskiego',
+    // TODO: get a proper (non-google translate) translation
+    easter_vigil: 'Wigilia Wielkanocna',
```
tukusejssirs: [src](https://drive.google.com/file/d/1KQ0XEoiQjX79sR1mfRc6OaRfeoulPcv5/view?usp=drive_link)

```suggestion
    easter_vigil: 'Wigilia Paschalna', // src: Kalendarz liturgiczny Diecezji Tarnowskiej, Wigilia Paschalna, p135 (https://drive.google.com/file/d/1KQ0XEoiQjX79sR1mfRc6OaRfeoulPcv5/view)
```

lib/locales/sk.ts
```diff
@@ -371,6 +371,8 @@ export const locale: Locale = {
     dunstan_of_canterbury_bishop: 'Svätého Dunstana z Canterbury, biskupa',
     dyfrig_of_wales_bishop: 'Svätého Dubrícia z Walesu, biskupa',
     easter_sunday: 'Nedeľa Pánovho zmŕtvychvstania',
+    // TODO: get @tukusejssirs feedback on this translation
+    easter_vigil: 'Veľkonočná vigília',
```
tukusejssirs: I need to tend to this. I want to check it in `mr_sk_2021_ed3` in my local church. Bear with me please.

tukusejssirs (changes requested):
I have a few change requests (related to your TODO code comments). The biggest issue I see is an unimplemented way to add celebration subtitles and/or Mass formulars (I am not sure if it is actually called a Mas _formular_ in English; it’d be awesome if you confirm it :wink:).

emagnier (changes requested):
This PR should not be merged, or at least not in its current state. Because it goes against what Romcal currently produces (`LiturgicalDay` vs. `Celebration`, which are not the same level of data and abstraction).

Indeed, it aligns with the comment I had written [here](https://github.com/romcal/romcal/issues/199#issuecomment-660554577) in 2020. However, in the meantime, the Romcal codebase and functionalities have evolved significantly.

And today, this PR goes against the points I mentioned here: https://github.com/romcal/romcal/issues/511.

I will take more time this evening to explain the reasons (as I'm not too available at the moment).

lib/locales/pl.ts
TobiTenno: i'll reformat w/ a block comment

lib/locales/en.ts
TobiTenno: will do.... not sure we need to add a specific source for something that already exists

lib/locales/en.ts
TobiTenno: pretty much this one you just gave me

TobiTenno:
@emagnier 
If I shouldn't do this, we should close #199 as no longer needed and reopen once whichever issue we end up doing between #297/#20 (or both!), as it doesn't make sense in current context if there are still several parts to sort out the why and how, as the days we describe (and already have, the Holy Thursday split is not new) in #511 clearly already allows for this kind of exception, it just doesn't make it easy to create

emagnier:
I rather think that the spec of https://github.com/romcal/romcal/issues/199 needs to be redefined ([see my comment ](https://github.com/romcal/romcal/issues/199#issuecomment-2097052129) in this issue).

https://github.com/romcal/romcal/issues/297 and https://github.com/romcal/romcal/issues/20 are not in contradiction to this topic (and can be addressed separately).

About https://github.com/romcal/romcal/issues/511, it's not exactly the same thing, as Holy Thursday is the day when there's a change of liturgical season within the civil day (Lent to Paschal Triduum at evening mass). So, they have not the same metadata. Whereas Easter Vigils are intimately linked to Easter Sunday.

emagnier:
I think this PR should be closed.
The concept of `LiturgicalDay` absolutely needs to be preserved (it's the foundation for building the liturgical calendar) and shouldn't be mixed with the concept of `Celebration`.

The ability to output `Celebrations` (instead of `LiturgicalDays`) isn’t currently supported, but it would definitely be a feature worth adding.

TobiTenno:
@emagnier in the future, I'd update the ticket first to change the detail to somehow output celebrations, because that's pretty different than what was intended before. Since that's the desired go-forward, I agree, but i don't think this PR would've broken the concept of LiturgicalDay, personally.

emagnier:
Sorry, I may have been a bit too quick and radical here. You're right, it’s definitely possible to update the details of this PR while still preserving the concept of a `LiturgicalDay` (feel free to reopen it if needed, or start a new branch/thread, whichever you prefer).

That said, I can confirm that this change would have broken the concept as it stands, because the Easter Vigil and Easter Sunday both belong to the same *liturgical* day (in the strictly liturgical sense, not the civil 24-hour day starting at midnight). 

To avoid spreading the discussion too much, I'm responding more broadly here  [https://github.com/romcal/romcal/issues/199#issuecomment-2893634247](https://github.com/romcal/romcal/issues/199#issuecomment-2893634247), with a proposed better approach to tackle this topic.

TobiTenno:
well, _Liturgically_ Easter Saturday has no liturgies proper to it apart from the Easter Vigil. I agree that it would've added some distinctness, but i don't think this would've radically broken our concept. I'll revisit and try a different way

emagnier:
The Roman Rite indeed does not provide a Mass on Holy Saturday. So it's completely logical — and even important — that Romcal doesn't return anything for that liturgical day, in keeping with the spirit of the liturgy for that day and the rules of the Roman Missal.

The issue you're describing is mostly a display issue (being able to show the Easter Vigil on Saturday evening). So in this case, we shouldn't try to bend the concept of `LiturgicalDay`, which should remain as-is (as a low-level core layer). Instead, we should build a layer on top of `LiturgicalDay` that can also generate entities like `Celebration` or `Mass`, which can meet needs that are strictly calendrical (in the sense of the civil calendar) or practical (in terms of visual display or user presentation, outside of a purely liturgical concept). See my comment here for a proposal in that direction: [https://github.com/romcal/romcal/issues/199#issuecomment-2893634247](https://github.com/romcal/romcal/issues/199#issuecomment-2893634247)

lib/locales/sk.ts
TobiTenno: @tukusejssirs whenever you have a chance, this would be most welcome, I'm planning on modifying the PR. not sure on the final direction yet, but these changes, i believe, will still be included.

PR #579, opened 2024-06-14
fix(general-roman): the Passion of Saint John the Baptist should be celebrated in red

emagnier:
On August 29, Saint John the Baptist is celebrated as a martyr, so the celebration should be in red (unlike the Nativity of Saint John the Baptist, which is in white).

Therefore, I am adding the missing title for this liturgical day (so that Romcal automatically colors it red).

See § 308.b. of the [GIRM](https://www.catholicculture.org/culture/library/view.cfm?recnum=337): 
> Red is used on Passion Sunday (Palm Sunday) and Good Friday, Pentecost, celebrations of the Lord's passion, birthday feasts of the apostles and evangelists, and celebrations of martyrs.

### DIFF
--- a/lib/catalog/martyrology.ts
+++ b/lib/catalog/martyrology.ts
@@ -3322,6 +3322,7 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
     },
     passion_of_saint_john_the_baptist: {
       name: 'Passion of Saint John the Baptist',
+      titles: [Title.Martyr],
     },
     patrick_of_ireland_bishop: {
       canonizationLevel: CanonizationLevels.Saint,
--- a/lib/catalog/martyrology.ts
+++ b/lib/catalog/martyrology.ts
@@ -3059,6 +3059,8 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       dateOfDeath: 1917,
     },
     nativity_of_john_the_baptist: {
+      // Note: The title of Martyr is not indicated here, as Saint John the Baptist is not yet a Martyr at his birth.
+      // Therefore, this celebration will be celebrated in white (src: GIRM, 308a).
       name: 'The Nativity of Saint John the Baptist',
     },
     nativity_of_the_blessed_virgin_mary: {
@@ -3321,6 +3323,8 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       titles: [Title.Religious],
     },
     passion_of_saint_john_the_baptist: {
+      // Note: unlike the celebration of the Nativity of Saint John the Baptist, here his title of Martyr is indeed indicated.
+      // Therefore, this celebration will be celebrated in red (src: GIRM, 308b).
       name: 'Passion of Saint John the Baptist',
       titles: [Title.Martyr],
     },


lib/catalog/martyrology.ts
```diff
@@ -3322,6 +3322,7 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
     },
     passion_of_saint_john_the_baptist: {
       name: 'Passion of Saint John the Baptist',
+      titles: [Title.Martyr],
```
TobiTenno, line 3325:
```suggestion
     // § 308.b. of the [GIRM](https://www.catholicculture.org/culture/library/view.cfm?recnum=337)
      titles: [Title.Martyr],
```

TobiTenno, line 3329:
```suggestion
     // § 308.b. of the [GIRM](https://www.catholicculture.org/culture/library/view.cfm?recnum=337)
      titles: [Title.Martyr],
```

TobiTenno (changes requested):
just a smol one for us to work on adding citations

lib/catalog/martyrology.ts
TobiTenno: will definitely approve once this is in

emagnier: If we add this comment here, we should add it to all entities with the title of Martyr. In my view, this doesn't make much sense because we are simply following the missal's rule.

However, since this saint is celebrated twice a year for different reasons (which is much less common and limited to a few saints), you make a good point about clarifying this case. We celebrate him as a martyr here, while for the Nativity of Saint John the Baptist, this is not the case (hence the celebration in white for that feast). So I'm adding a comment here. This way, in the future, we can avoid mistakenly adding the "Martyr" title to `nativity_of_john_the_baptist`.

Thank you @TobiTenno  🙏 .

TobiTenno: my point is that we should add it as we're modifying things so that we can add it when we add something and notice an inconsistency

TobiTenno (approved):

lib/catalog/martyrology.ts
```diff
@@ -3321,7 +3323,10 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       titles: [Title.Religious],
     },
     passion_of_saint_john_the_baptist: {
+      // Note: unlike the celebration of the Nativity of Saint John the Baptist, here his title of Martyr is indeed indicated.
+      // Therefore, this celebration will be celebrated in red (src: GIRM, 308b).
       name: 'Passion of Saint John the Baptist',
+      titles: [Title.Martyr],
```
tukusejssirs, line 3329: I believe this is not the best solution.

While St John the Baptist was a martyr, the name of this celebration does not include the `Martyr` title, [1] therefore IMHO we should not return `Title.Martyr` for this celebration. At least not until we (1) assign/choose all the titles in each celebration definition in calendars; and (2) we generate the celebration names from separated l10n data.

IMHO defining `color` explicitly is a better solution in this case. Or we could check for the celebration ID and implicitly define the color for this celebration.

[1] The celebration is, as I call it, one of an _event_ (the passion/martyrdom of St John the Baptist) and not a personal celebration.

lib/catalog/martyrology.ts
```diff
@@ -3059,6 +3059,8 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       dateOfDeath: 1917,
     },
     nativity_of_john_the_baptist: {
+      // Note: The title of Martyr is not indicated here, as Saint John the Baptist is not yet a Martyr at his birth.
```
tukusejssirs, line 3062: Just a nitpick.

/cc @TobiTenno, maybe this could be written more naturally / better.

```suggestion
      // Note: The title of a martyr is not indicated here, as Saint John the Baptist is not yet a martyr at his birth.
```

lib/catalog/martyrology.ts
```diff
@@ -3321,7 +3323,10 @@ export const Martyrology: { catalog: MartyrologyCatalog } = {
       titles: [Title.Religious],
     },
     passion_of_saint_john_the_baptist: {
+      // Note: unlike the celebration of the Nativity of Saint John the Baptist, here his title of Martyr is indeed indicated.
```
tukusejssirs, line 3326: Just a nitpick.

/cc @TobiTenno, maybe this could be written more naturally / better. Should we use `specified` instead of `indicated`?


```suggestion
      // Note: Unlike the celebration of the Nativity of Saint John the Baptist, here the title of a martyr must be indicated.
```

tukusejssirs (changes requested):

lib/catalog/martyrology.ts
emagnier: I didn't have the same reasoning, but your point of view is interesting.

On my side, I tried to strictly follow the rules of the Missal to be as objective as possible: if a saint is a martyr, they are celebrated in red; if not, it is in white. Therefore, if a memorial is celebrated in red, it means that the person is indeed a martyr (unless there is an explicit exception described in the Missal).

Indeed, the word "Martyr" is not directly in the title of this memorial, but in this context, the term "Passion" (used in Latin and English) remains a synonym for "Martyr". FYI, the word "Martyr" is explicitly used in French. So, I relied on the traditional practices of the Church—which celebrates this memorial in red—to corroborate adding the title "Martyr".

So, I would say, regardless of which rule we want to follow here: based on the terminology of the memorial's title, or based on the analysis of the Missal's rules (although I prefer the latter, which I find more objective), as long as this liturgical day is celebrated in red, it works for me. I'll leave the final word to you on this 😉

lib/catalog/martyrology.ts
emagnier: In this context, I believe "indicated" could mostly mean "highly recommended," whereas "specified" has more of a connotation of specs. So indeed, "specified" would be more appropriate here. But yes, this is really nitpicking!

If you see other things like this, feel free to open PRs and suggest improvements (teamwork!). On my side, unless I'm writing something completely wrong or writing incomprehensible things (English is not my native language), I prefer to focus on functional improvements to Romcal. Otherwise, I won't have time to do more on this project. 😇

lib/catalog/martyrology.ts
tukusejssirs: > If we add this comment here, we should add it to all entities with the title of Martyr.

As I see it, we should add a source reference to any and all data we use in romcal, regardless of the source type (first-party/second-party/third-party source), although first-party sources should be always prefered (like the Roman Missal, incl GIRM).

lib/catalog/martyrology.ts
tukusejssirs: > On my side, I tried to strictly follow the rules of the Missal to be as objective as possible: if a saint is a martyr, they are celebrated in red; if not, it is in white. Therefore, if a memorial is celebrated in red, it means that the person is indeed a martyr (unless there is an explicit exception described in the Missal).

I have no objection to this reasonsing of yours. All I disagree is to add a title to a celebration which does not use it in its celebration name.

While your solution does not have any downside in the current romcal implementation, however, once we generate the celebration name from its metadata (see #154), it will be an issue. That’s I think your solution (adding a `Title.Martyr` to this celebration) is not the best solution.

In the end, IMHO we should add a custom rule to the romcal core (or however you call it; the place where we do the calendar generation itself), as this celebration is inscribed in the GRC, thus celebration everywhere. That said, an easier solution is to use `color` explicitly to override this.

> Indeed, the word "Martyr" is not directly in the title of this memorial, but in this context, the term "Passion" (used in Latin and English) remains a synonym for "Martyr". FYI, the word "Martyr" is explicitly used in French. So, I relied on the traditional practices of the Church—which celebrates this memorial in red—to corroborate adding the title "Martyr".

Yeah, I agree with `martyr` and `passion` in this sense are related terms, however, they are not synonyms. In the end, it does matter whether they are synonyms on not, the the _usage_ is what matters here IMHO:

- `passion` is used in the celebration name;
- `martyr` (and thus `Title.Martyr`) is used in the celebration metadata.

The latter is incorrect, as the celebration is not a _personal_ celebration. We assign titles to persons (saints, blesseds), not to a non-personal celebrations (i.s. events).

I agree that St John the Baptist _is_ a martyr. Therefore, we should add `Title.Martyr` to `john_the_baptist_martyr` in `martyrology.ts` (or the future file with all the celebration metadata), however, IMHO it should not be used for this celebration.

lib/catalog/martyrology.ts
tukusejssirs: I am nitpicking about the semantics! :wink: Sorry! :sweat_smile:

---

From my POV, the celebration definitions in calendars (which have the `LiturgicalDayInput` type) are composed of the celebration ID (a property of the `Inputs`) and the props of `LiturgicalDayInput` are the _specs_ (as you sort of called it) or _metadata_ (as I call it). Thus, from my POV, they are specified, or defined for a particular celebration inscribed in a calendar. :wink:

> If you see other things like this, feel free to open PRs and suggest improvements (teamwork!).

Yeah, I will, however, unfortunately, I’ve been really busy lately, unfortunately. I wish I could find more time for romcal.

> On my side, unless I'm writing something completely wrong or writing incomprehensible things (English is not my native language), I prefer to focus on functional improvements to Romcal. Otherwise, I won't have time to do more on this project. 😇

I get it, Etienne. Thank you very much for you do for romcal! :pray:

Neither am I a native English speaker, however, I tried to do my best (as you do). :wink:

I agree the functional improvements are more important that this, that is why I label the less imporant (PR) comments as nitpicks: they are not necessary to be implemented/fixed. They could be completely ignored, however, I still want to raise them. Sorry about the noise. :wink:

lib/catalog/martyrology.ts
TobiTenno: probably, but i think it's poignant when referencing the title to have it specifically called out in this way.

I think `the title of Martyr` is the most precise way of saying it, as `a martyr` is not a title, you'd be in the state of being a martyr (the second case), but `the title of Martyr` is a universal, like a tag... so it's like saying `the tag of dependency` or `the dependency tag`, as opposed to "the tag of a dependency" sounds like you're speaking of a particular dependency.... sorry, way too much English nuance in this, but this is the most natural way I thought it could be done

lib/catalog/martyrology.ts
TobiTenno: i generally agree w/ @emagnier on this one. titles are personal, not ephemerally disassociated from merit given a saint, so the possessive "his title of Martyr" is the most natural in English grammar.

TobiTenno: I think the best resolution for the future on these is to include metadata/specifications/whatever you wanna call them as an object field, like "sources" so that we can reference the sources programmatically and enforce them using linting or typing

emagnier: Including a way to indicate rules/sources is an important topic that we need to explore 💯.  And having linting or typing would be very cool!

We also need to keep some flexibility for special cases like this one. In this case, the main issue is the lack of an explicit mention in the Roman Missal. Especially considering that the Missal's rules specify that only martyrs are commemorated in red, along with a few other clearly mentioned celebrations, but without mentioning the Passion of Saint John the Baptist. The title "Martyr" is not explicitly given for this commemoration (at least in Latin and English, but in French, that's the case), and we don't want to add it. However, we know that in practice, this commemoration is indeed celebrated in red.

I didn't think there'd be so much talk when I opened this PR 😄. But we're getting into some interesting topics 👍

lib/catalog/martyrology.ts
emagnier: I'm totally on board with you on the terminology and localization front. So, let's go ahead and use the `color` property here. I'll open a PR to get that sorted.

emagnier: The fix is here: https://github.com/romcal/romcal/pull/584

lib/catalog/martyrology.ts
tukusejssirs: You’re right, @TobiTenno, there should be no indefinite artitle. :+1: Moreover, in English grammar, you describe different types of _phrases_ (e.g. noun phrase) as opposed to the Czech or Slovak grammar (where we deal more with individual words instead). That is a bit hard to understand and remember for me. As the phrase `title of martyr` is considered a single noun phrase (and not two nouns), there should be only a single (definite) article (and not two).

Just a final question regarding the phrase. I know that [titles](https://en.wikipedia.org/wiki/Title) (e.g. Mr/Mrs/Ms, PhD, Saint/Blessed, Martyr/Piest/Religious) are (sort of) part of name of a person in English. Now, as I see it, in this phrase, we are talking about the `martyr` title _in general_, but we also use the definite article before the phrase to denote that we are talking about this specific assignment of the title. **Do you still think we should use a capital initial letter in `Martyr`? Should we not lower-case it instead? :thinking:

Thank you very much! :pray:

lib/catalog/martyrology.ts
tukusejssirs: > i generally agree w/ @emagnier on this one. titles are personal, not ephemerally disassociated from merit given a saint, so the possessive "his title of Martyr" is the most natural in English grammar.

While I understand the titles are personal, this celebration (from my POV) is not a personal celebration.

Re the possessive pronoun: for me it’s a bit strange, however, it is not incorrect. YMMV. I have no problem to keep the pronoun though. :wink:

> I think the best resolution for the future on these is to include metadata/specifications/whatever you wanna call them as an object field, like "sources" so that we can reference the sources programmatically and enforce them using linting or typing

I think we should provide (at least in the each liturgical day object returned by `Romcal#generateCalendar()`) an object with all celebration metadata from `martyrology.ts`. However, I am not persuaded to add (personal) titles to non-personal celebrations.

I also think we should add metadata of all sources used by a celebration to each liturgical day item returned by `Romcal#generateCalendar()`. I understand that this is not possible yet, as the sources are not implemented yet.

> We also need to keep some flexibility for special cases like this one.

IMHO this is not a case special enough to justify an extra-ordinary way to define the liturgical colour. IMHO the best solution is to bake this in the computation logic, as it is used in the GRC, thus in the whole world.

That said, it might be useful for other celebrations not inscribed in the GRC; then IMHO we should use the `color` property.

> I didn't think there'd be so much talk when I opened this PR 😄. But we're getting into some interesting topics 👍

Sorry, Etienne! :pray: You know my stubbornness and my love for greate detail.

However, I think this conversation gave us all at least one thing: we might better understand how each of us think and feel about this particular issue. Think could help us do better decisions in the future. :wink:

lib/catalog/martyrology.ts
TobiTenno: when speaking of Martyr in such a way that the definite article (`the`) would be used, a capital can be appropriate if you're speaking of the thing with particular importance. It's the reason names of things in English, like `English`, `Spanish`, are not lower-cased like they are in many Romance (Latin-derivative) languages.

PR #584, opened 2024-06-16
fix(general-roman): the Passion of Saint John the Baptist in red, without using a Martyr title

emagnier:


TobiTenno (approved):

TobiTenno:
well, that's confusing. commitlint passed fine on your PR, but not on the release build :-/

edit: oh, it's cause your title that was forced in was too long.... yeah, we need to not try to use that long of titles. @emagnier you'll have to manually fix locally and force push your last commit to fix the title. i'm gonna find a way to lint the title length to be < 101 chars.... unless we should just remove this restriction entirely?

emagnier:
I just fixed it on `dev`. Indeed, when merging, GitHub appended the PR reference " (#584)" at the end, which made the title longer.

I'm not sure if there's a way to add validations at this level, but I don't think so... 🤔

TobiTenno:
i'll do some tests, but we can also remove the max limit on it easily, too

TobiTenno:
@emagnier fix for that https://github.com/romcal/romcal/pull/585

PR #585, opened 2024-06-17
ci: remove max header length

TobiTenno:


tukusejssirs:
I have no opinion on this.

While I am a proponent of soft-wrapping, I also understand that the [50/72 rule](https://dev.to/noelworden/improving-your-commit-message-with-the-50-72-rule-3g79) makes sense.

In the end, every commit message can be re-written to abide any length rule.

Therefore, if Etienne agrees with the commit message length check removal, I have no problem with that. :wink:

TobiTenno:
my opposition to soft-wrapping on titles is that they get really unruly and hard to read when you're trying to visually grok what's been changed, having the soft-wrap instead of just adding a body to the commit message gets annoying

tukusejssirs:
99 % of time, I do that in my terminal or IDE, which both soft-wrap the commit message (header and body).

However, as I said, I have no problem with hard-wrapping the message. :wink:

emagnier:
I also lean towards your opinion @TobiTenno on soft-wrapping, and I think it's good to keep a limit on the title (not too short, not too long). It forces us to be precise and concise, and it's easier to read in a changelog. For more details, the body of the commit message can be used for that.

Maybe we could have a middle ground, like keeping the limit in our PRs, but not checking the length once the commit is merged into `dev`?

TobiTenno:
I would like that, but i don't know that we want to do 2 configs, one for merge, one for publish 🤔 

It's definitely doable

emagnier:
Yeah, it’s not great to duplicate configs just to manage these 2 scenarios. Let’s remove the limit for now. We can revisit this later if it becomes a hassle or we come up with better ideas.

emagnier (approved):

tukusejssirs (approved):
As Etienne approved this PR, I’ll approve it too so that it could get merged. :wink:

(Good to know that pending reviews are _blocking_ auto-merge of a PR—I like it!)

Issue #119, opened 2019-12-14
calendar metadata: checked with official sources


tukusejssirs:
Currently, some calendars are checked with official sources, others rely on non-official sources (such as Wikipedia). I think we should add this kind of metadata to all calendars.

Submetadata could be a year which was checked. That could be useful with the `histcal` (#23).

For example `sk.js` was checked with the official sources, but `hungarian.js` was not (yet). And it was checked only for years 2018 and 2019.

There could be also an option to generate calendar using checked data or not; in this case it could be even better to add this metadata not to _calendar_, but to _celebrations_ (only in particular calendars), but that would create a burden when defining the calender (for this, the calendar definition tool, e.g. a website, would be helpful.

tukusejssirs:
I have just found [GCatholic.org](http://www.gcatholic.org/documents/year/2019.htm#7534) which (besides other stuff) lists (presumably) all the decrees issued by the Congregation for Divine Worship and the Discipline of the Sacraments along with their respective sources (the decrees).

I think we should parse it (by hand at least) to:
- insert the latest additions to the calendars;
- to add `startDate` metadata for historical purposes (see #96).

PR #635, opened 2024-08-15
feat(de): add German translation

btielen:
**Summary**

This pull request introduces German translations (de locale) to our project.

**Details**
- The new `/lib/locales/de.ts` file is derived from the existing `/lib/locales/en.ts` file.
- The names of our Saints and other feast days were initially translated using AI.
- These translations were then carefully reviewed and manually corrected to ensure accuracy.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "romcal",
-  "version": "3.0.0-dev.82",
+  "version": "3.0.0-dev.83",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "romcal",
-      "version": "3.0.0-dev.82",
+      "version": "3.0.0-dev.83",
       "license": "MIT",
       "workspaces": [
         "rites/*",
@@ -3258,7 +3258,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001664",
+      "version": "1.0.30001667",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001667.tgz",
+      "integrity": "sha512-7LTwJjcRkzKFmtqGsibMeuXmvFDfZq/nzIjnmgCGzKKRVzjD72selLDK1oPF/Oxzmt4fNcPvTDvGqSDG4tCALw==",
       "dev": true,
       "funding": [
         {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "romcal",
-  "version": "3.0.0-dev.82",
+  "version": "3.0.0-dev.83",
   "description": "JavaScript library that generates liturgical calendars of the Roman Rite of the Roman Catholic Church.",
   "keywords": [
     "roman",


tukusejssirs (changes requested):
@btielen, thank you very much for creating of this PR, however, I don’t think we should add any AI generated content (calendar/localisation data or celebration metadata / martyrology), more so when the data is quite easily acquirable at least from the secondary sources (ordos/directoriums or official websites of episcopal conferences or their liturgical commissions). I am not totally against the addition of AI generated, manually verified and corrected celebration names for celebrations we don’t have even a third-party source in a particular language, however, I still would prefer a manual translation.

Therefore, I would reject this PR in its current state. WDYT, @emagnier and @TobiTenno?

---

Celebration names in German usually include additional information in comparison with English celebration names (similarly to French celebration names), like a seat of the bishops or date of death for saints/blesseds. For example, you have translated `ansgar_of_hamburg_bishop` as `Hl. Ansgar, Bischof`, while _Die Messfeier – Dokumentensammlung. Auswahl für die Praxis_ (see below) uses the following name: `Ansgar, Bischof von Hamburg-Bremen, Glaubensbote in Skandinavien (865)`.

It would be nice to source the celebrations (both their names and their metadata) from the German translation of the Roman Missal, as I consider it (along with the Holy See decrees) as a primary/first-party source for the data used by romcal. While I have no access to it, I have found [_Die Messfeier – Dokumentensammlung. Auswahl für die Praxis_](https://drive.google.com/drive/u/2/folders/1LBIQ2KZMk26AG5FWU8HQHf7pL_MPPAz2) (The Celebration of the Mass–Collection of Documents. Selection for Practice) published in 2015 which contains on (actual) p131-144 (PDF p128-141). For now, I would follow the celebration names from there, but I would definitely prepend `† ` (a dagger with a space) before each death date.

I have also found a [Missal](https://drive.google.com/drive/u/2/folders/1kORCJb0w-pykhI5r_TJwT7f_c1fe9gIK) (or a snippet only?) used by the [Teutonic Order](https://en.wikipedia.org/wiki/Teutonic_Order) ([Order of Brothers of the German House of Saint Mary in Jerusalem](https://www.deutscher-orden.de/)) which might be also helpful during localisation of the celebration names, later also when we want to add their Proper (calendar) to romcal.

Of course, celebrations not celebrated in German should be sourced from third-party sources or translated into German, but the style should be the same as one of the other, official celebration names.

[Notes for a future revision of the calendars for german-speaking countries/dioceses] Also note that AFAIK Austria, Germany, Switzerland, and German-speaking dioceses use a common, regional calendar ([src](https://en.wikipedia.org/wiki/National_calendars_of_the_Roman_Rite#Austria,_Germany,_Switzerland,_German-speaking_dioceses)). What this means: the German calendars (actually: the celebrations inscribed in them) need to be defined in the correct places so that we can use fully use calendar inheritance:

- General Roman Calendar;
- `europe.ts`;
- the regional calendar (e.g. `austria-germany-switzerland.ts`);
- `austria.ts`, `germany.ts`, `switzerland.ts` for celebrations added on top of the regional calendar, celebrated on the national/country level;
- `${country}.${diocese}.ts` for diocesan celebrations.

And I have also found the following diocesan ordos:

- [Archdiocese of Cologne](https://drive.google.com/drive/u/2/folders/1BHhywaehj6HYxAHPGLnotCGmHQ9IGJuS); 
- [Diocese of Trier](https://drive.google.com/drive/u/2/folders/1Cw3ZRW7Mg_Ket79ppjagkGorQf2PRJSg).

Finally, we now prefer adding a code comment to every localisation string and celebration definition with the source reference.

emagnier:
Thank you so much @btielen for your PR!

In response to @tukusejssirs' comment, I'm inclined to accept this PR. However, the source needs to be clearly indicated. In this case, it should be noted that the translation was generated by AI (which one?) from English.
At least this new localization is in place, making it easier to refine and improve these translations iteratively to raise the quality in Germain. What do you think @tukusejssirs and @TobiTenno?

However, I believe there's a real issue with the ability to rate localizations. We need to improve our tools (and the data model) to quantify the quality of a localization (e.g., whether each translated property comes from an official liturgical book or not). This is of course outside the scope of this PR, but this PR does highlight the issue.

tukusejssirs:
> I'm inclined to accept this PR. However, the source needs to be clearly indicated. In this case, it should be noted that the translation was generated by AI (which one?) from English. At least this new localization is in place, making it easier to refine and improve these translations iteratively to raise the quality in Germain.

While I would prefer having validated data in romcal only (i.e. the validation should take place during the feature review phase), I won’t argue with you about this. :wink:

Therefore, I agree to accept this PR only if each l10n item is accompanied by a comment from where the translation is sourced.

> However, I believe there's a real issue with the ability to rate localizations. We need to improve our tools (and the data model) to quantify the quality of a localization (e.g., whether each translated property comes from an official liturgical book or not). This is of course outside the scope of this PR, but this PR does highlight the issue.

I agree we need to add a way to specify the source reference (incl all the required bibliographical data, as well as the source type [first/second/third party] and other stuff). IMHO we should add this to all the data in romcal, i.e. celebration metadata (`martyrology.ts`), calendars, celebrations inscribed in calendars, localisation strings. (Related: #119, [my list of the GRC with all celebrations accompanied by a source](https://gitlab.com/tukusejssirs/romcal_calendars/-/blob/main/1969/general/general.md).)

TobiTenno:
I think @btielen did specify everything was reviewed and then manually corrected.

@btielen do you have any copies of the German Ordinary calendar to validate against that we could separately verify in the future?

I'm fine with this, but I agree that ideally we'll have a source in the future. I think there should be a "base source" that we use when we don't have an extra explicitly defined, ideally the primary linguistic calendar for the locale writ large

For starting locales (i.e. the German language and not the calendar for Germany), I have no problem relying on "this is what we call it, and I'm German" as the initial translation.

tukusejssirs:
@TobiTenno, the thing is that the AI translated version relies on the format used in `en.ts` which is _not_ wat is used in German. The format is more like the one used in `fr.ts`. Basically, they add some more information to te celebration names (e.g. death dates, bishopric sees).

See [this comment of mine](https://github.com/romcal/romcal/pull/635#pullrequestreview-2244388590) for more info.

emagnier:
Following the merge of https://github.com/romcal/romcal/pull/535, I rebased a few PRs in this repository onto the latest version of the `dev` branch. This branch was also named `dev`, which likely caused an issue during the force push (the branch was closed and emptied). 

So, I reopened a rebased version of this branch here: https://github.com/romcal/romcal/pull/681.

PR #681, opened 2024-10-09
feat(de): add German translation

emagnier:
Originally opened here https://github.com/romcal/romcal/pull/635 by @btielen 

> After a rebase and a force push (since https://github.com/romcal/romcal/pull/535 was merged), the original PR was automatically closed (probably because this branch was also the main branch of the fork). I am reopening it here to continue the work.

---

**Summary**

This pull request introduces German translations (de locale) to our project.

**Details**
- The new `/lib/locales/de.ts` file is derived from the existing `/lib/locales/en.ts` file.
- The names of our Saints and other feast days were initially translated using AI.
- These translations were then carefully reviewed and manually corrected to ensure accuracy.


TobiTenno:
referencing past qualms about merging as-is:
- i'm fine with largely generated layouts that have been reviewed
- the person submitting is a german speaker and indicated that everything seemed correct for the locale
- re: regional/diocesan/europe-wide calendars: these can be created at a later date when/if someone wants to create them, but if a "higher order"  or wider-reaching calendar does not yet sufficiently cover the uses for which this was defined, i think we continue as-is and merge; and if someone creates the wider-reaching ones in the future, we can properly trim from this if needed, or allow it to stand as-is, given that they aren't treated as duplicates, but overrides; so at worst we're representing the same information twice, which works out fine in the final, flattened config.

TobiTenno (approved):

emagnier:
I'm merging the PR. I believe that a contribution like this, which adds support for a new language to Romcal, shouldn't be left pending indefinitely. We can always iterate later to improve quality and validate elements with sources. But in the meantime, Romcal will already be usable in German, which could potentially attract more future contributors in this language.

And thanks again @btielen for this initial work!

PR #686, opened 2024-10-18
feat(calendar): add a romcal option to promote optional memorials

emagnier:
The `elevatedMemorialIds` is a list of optional liturgical memorials that are to be elevated to the rank of mandatory memorials within the liturgical calendar, for a specific calendar.

Each ID in the list corresponds to a specific optional memorial that, according to particular liturgical norms or decisions, must be celebrated as a mandatory memorial.

```ts
const romcal = new Romcal({
  // List of optional memorials to be elevated to mandatory memorials
  elevatedMemorialIds: ['john_paul_ii_pope', 'our_lady_of_fatima'],
});
```

Note:
- If multiple optional memorials occur on the same day, and one of them is elevated, then only the elevated memorial will be retained in the generated calendar. The other optional memorials are removed, as the elevated (mandatory) memorial takes precedence over the others.
- If multiple memorials are elevated for the same day, only the first elevated memorial defined in the calendar will be retained, as there can only be one mandatory memorial on a given liturgical day.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ const romcal = new Romcal({
   epiphanyOnSunday: true | false, // Epiphany always a Sunday (between January 2 - 8), or on January 6
   corpusChristiOnSunday: true | false, // Corpus Christi always a Sunday, or the Thursday after Trinity Sunday
   ascensionOnSunday: true | false, // Ascension always a Sunday, or the 40th day of Easter (a Thursday)
+  elevatedMemorialIds: ['john_paul_ii_pope', 'our_lady_of_fatima'], // List of optional memorials to be elevated to mandatory memorials
 });
 ```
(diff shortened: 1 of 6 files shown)


TobiTenno (approved):
i don't get it, but i like the implementation

emagnier:
This feature can, for example, help address this need (shared in this discussion https://github.com/romcal/romcal/discussions/677 by @tukusejssirs): 

> - add a way to select which celebration will be celebrated on a particular day (feria vs optional memorials):
>   - this would be useful for planning the liturgical celebrations in a particular parish, or even better: define a celebration/formular for a specific Holy Mass (as different Masses on the same date in the same parish might be celebrated Holy Masses with different formular);

Issue #587, opened 2024-06-21
Commons missal metadata


emagnier:
In the missal and the lectionnary, _Commons_ are collections of prayers, readings, and other liturgical texts that are used for various categories of saints or types of celebrations. 

Unless a saint already has all their proper content (readings, prayers, etc.), one or more commons are usually mentioned as references for celebrating that saint's commemoration.

I propose adding the ability to specify this information for each saint or blessed (in the `martyrology.ts` file) through a new property `commons`, which can accept one or more commons. These commons will be listed as an Enum.

Some commons are conditional on the current liturgical season. Therefore, this must be managed dynamically by filtering out the commons that do not correspond to the current liturgical season when building the liturgical calendar.

Some commons are intended for multiple saints or for a single one. Similarly, this can be automatically refined (at the time of building the liturgical calendar) by relying on the `count` property of each entity in the `martyrology.ts` file, where the default value is `1`; or in cases where multiple saints with the same commons are commemorated together on the same liturgical day.

The missal includes the following categories, which can be used to construct two Enums: one simplified (for the source data) and one comprehensive (with refined _common_ values during the build of the liturgical calendar):

- **COMMON OF THE DEDICATION OF A CHURCH (on the anniversary of the dedication)**
  1. In the Church that was Dedicated
  2. Outside the Church that was Dedicated
- **COMMON OF THE BLESSED VIRGIN MARY**
  1. In Ordinary Time
  2. In Advent
  3. In Christmas Time
  4. In Easter Time
- **COMMON OF MARTYRS**
  1. Outside Easter Time
    A. For Several Martyrs
    B. For One Martyr
  2. During Easter Time
    A. For Several Martyrs
    B. For One Martyr
  3. For Missionary Martyrs
    A. For Several Missionary Martyrs
    B. For One Missionary Martyr
  4. For a Virgin Martyr
  5. For a Holy Woman Martyr
- **COMMON OF PASTORS**
  1. For a Pope or for a Bishop
  2. For a Bishop
  3. For Pastors
    A. For Several Pastors
    B. For One Pastor
  4. For Founders of Churches
    A. For One Founder
    B. For Several Founders
  5. For Missionaries
- **COMMON OF DOCTORS OF THE CHURCH**
- **COMMON OF VIRGINS**
  1. For Several Virgins
  2. For One Virgin
- **COMMON OF HOLY MEN AND WOMEN**
  1. For All Categories of Saints
    A. For Several Saints
    B. For One Saint
[...]

PR #689, opened 2024-10-23
feat(calendar): add the initial Commons metadata

emagnier:
Related to https://github.com/romcal/romcal/issues/587.

This PR introduces two new enums to define the liturgical **_Commons_** for each liturgical day (if applicable):
- `Common`: Specifies precisely the liturgical _Common_ to use, taking into account, when applicable, the number of entities (one/several) and the liturgical season. This will be automatically calculated when generating the calendar (PR to follow) based on the `CommonDefinition` enum described below.
- `CommonDefinition`: A simplified version of the enum, to be used only in calendar definitions.

Some liturgical days do not have a defined _Common_ because a specific proper content (prayers, collects, etc.) is already provided in the missal. In such cases, the property `commons: Common.None` is proposed, to explicitly indicate that no _Common_ is to be used for that particular liturgical day.

This PR also adds the `commons` property to each element of the _General Roman Calendar_. The sources have been verified and updated based on the 3rd amended edition of the Roman Missal in Latin (2008), and on the latest versions of the Italian (2020) and French (2021) missals for sanctoral elements added after 2008. When multiple _commons_ are available for the same sanctoral entity, they are defined in Romcal following the exact order written in the Missal.

To-do for future PRs:
- Add the `commons` property to the final `LiturgicalDay` objects calculated and returned by Romcal, considering as mentioned earlier, the number of entities (one/several) and the liturgical season.
- Provide l10n for every value of this new `commons` property.
- Provide CLI tasks to check that commons are correctly defined for a calendar.

I initially considered defining this new property in the martyrology catalog. In fact, I began doing so before deciding to roll it back and define them only within the calendar definitions. In any case, if these _commons_ were defined in the catalog, some would still need to be defined in the calendar, especially for groups of people, since the recommended _commons_ can differ depending on whether a single person or a group is being commemorated. Additionally, by defining commons directly in the calendars, it prevents the issue of sharing _commons_ definitions between different proper calendars through the martyrology catalog, which could lead to cases where a _common_ is suggested in one proper calendar while another proper calendar omits it because a specific proper content is provided. Ultimately, defining the _Commons_ solely within the calendar simplifies the code that handles this metadata, makes the property easier to edit and review, and allows them to be strictly defined as indicated in the respective missals.

### DIFF
--- a/rites/roman1969/__tests__/calendar-builder.test.ts
+++ b/rites/roman1969/__tests__/calendar-builder.test.ts
@@ -15,6 +15,7 @@ import {
   Season,
 } from '@src/rite-roman1969';
 
+import { CommonDefinition } from '../src/constants/commons';
 import { Precedences } from '../src/constants/precedences';
 import { dateDifference } from '../src/utils/dates';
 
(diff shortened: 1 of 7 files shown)


rites/roman1969/__tests__/__snapshots__/calendar-builder.test.ts.snap
TobiTenno, line 1: i'm not sure i understand the point of snapshots on a library, but... cool? 👍

TobiTenno (approved):
i like it so far, but i definitely want to go add all the other metadata for updates now so we can include src there in code instead of our comments

rites/roman1969/__tests__/__snapshots__/calendar-builder.test.ts.snap
emagnier: I would say it's still useful as long as we don't have a truly comprehensive test coverage, especially when modifying the deeper logic of Romcal, to ensure we don't break anything (it's saved me a few times 😅).

PR #690, opened 2024-10-24
feat(calendar): remove dedication_of_consecrated_churches from GRC

### LINKED ISSUE #136: Add `dedicationOfAParticularChurch` to `general.js`

emagnier:
The `dedication_of_consecrated_churches` liturgical day is not declared anywhere in the GRC, so I am removing it.

I am reintegrating it into the proper calendars of:
- Belgium – based on the 2021 French missal
- Canada – based on the 2021 French missal
- France – based on the 2021 French missal
- Poland – I don’t have the sources, but there was only a removal of `dedication_of_consecrated_churches_on_october_25`, not of `dedication_of_consecrated_churches_on_last_sunday_of_october`. So I assumed the latter was correct, but maybe it was an oversight to remove that as well? If in doubt, we can also remove this element from this calendar entirely (we can reintroduce it later if needed).
- Slovakia

For France and Belgium, the Missal also allows transferring the liturgical day to the last Sunday of the October. Since we don’t want to have this same liturgical day twice in the calendar, and I found it unfortunate to lose this information, I suggest adding a new optional property `alternativeTransferDateDefs`.

For now, the dates defined in this property `alternativeTransferDateDefs` are not calculated. I’m reserving this for a future PR.

Closes https://github.com/romcal/romcal/issues/136.

### DIFF
--- a/rites/roman1969/src/particular-calendars/canada.ts
+++ b/rites/roman1969/src/particular-calendars/canada.ts
@@ -196,8 +196,10 @@ export class Canada extends CalendarDef {
     },
 
     // src: mr_fr_2021_ed3
-    dedication_of_consecrated_churches_on_last_sunday_of_october: {
-      drop: true,
+    dedication_of_consecrated_churches: {
+      precedence: Precedences.ProperSolemnity_DedicationOfTheOwnChurch_4b,
+      dateDef: { month: 10, date: 25 },
+      isOptional: true,
     },
   };
 }
--- a/rites/roman1969/__tests__/calendar-builder.test.ts
+++ b/rites/roman1969/__tests__/calendar-builder.test.ts
@@ -1,4 +1,5 @@
 import { England_En } from '@dist/rite-roman1969/bundles/england';
+import { France_Fr } from '@dist/rite-roman1969/bundles/france';
 import { GeneralRoman_En } from '@dist/rite-roman1969/bundles/general-roman';
 import { Germany_En } from '@dist/rite-roman1969/bundles/germany';
 import { Hungary_En } from '@dist/rite-roman1969/bundles/hungary';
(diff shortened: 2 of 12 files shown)


rites/roman1969/src/particular-calendars/canada.ts
```diff
@@ -196,8 +196,10 @@ export class Canada extends CalendarDef {
     },
 
     // src: mr_fr_2021_ed3
-    dedication_of_consecrated_churches_on_last_sunday_of_october: {
-      drop: true,
+    dedication_of_consecrated_churches: {
+      precedence: Precedences.ProperSolemnity_DedicationOfTheOwnChurch_4b,
+      dateDef: { month: 10, date: 25 },
+      isOptional: true,
```
TobiTenno, line 202: the new format seems to differ with the "last sunday" bits

TobiTenno:
```suggestion
      dateDef: { month: 10, date: 25 },
      isOptional: true,
      alternativeTransferDateDefs: [{ dateDef: { month: 10, lastDayOfWeekInMonth: 0 } }],
```

TobiTenno: unless i'm misunderstanding?

TobiTenno (approved):
so is this just common but not universal?

rites/roman1969/src/particular-calendars/canada.ts
emagnier: The proper missal for Canada only specifies October 25 and does not suggest transferring it to the last Sunday of October. That’s why I didn’t add an `alternativeTransferDateDefs` here.

emagnier:
> so is this just common but not universal?

More or less, yes. Not all proper calendars include such a liturgical day, and those that do have some variations, like specifying October 25 with or without the last Sunday of October.

igneus:
>>> The dedication_of_consecrated_churches liturgical day is not declared anywhere in the GRC [...]
>>
>> so is this just common but not universal?
>
> More or less, yes.

There is (AFAIK) no law on the subject, but there is an official interpretation which in fact introduces a universal rule:
[Notitiae 1972, p. 106](https://www.cultodivino.va/content/dam/cultodivino/rivista-notitiae/1970/notitiae-08-(1972)/Notitiae-071-1972.pdf#page=45)

emagnier:
Thank you @igneus, for pointing us to this document, it’s really very interesting!

Looking back, I have to admit I’m not completely satisfied with this PR, and I’m even considering a rollback to give us more time to find a better solution (while keeping the two dates functional in some calendars for now).

Our goal should remain to stay as close as possible to the rules of the missal. So, it’s unfortunate that there isn’t anything very specific on this point directly in the missal. I’m a bit hesitant to reintroduce them in the GRC (as a final solution), and I also have the impression that defining the `dedication_of_consecrated_churches` is left up to each bishops’ conference (meaning it’s decided/defined at the national calendar level).

So, we’d need a solution to:
- Define only one instance per calendar that applies to two/several different dates (for the calendars where two dates are proposed). Perhaps something that could be also useful for integrating recurring items like votive masses in the future.
- And potentially make it easier to define these in each national proper calendar (without having to duplicate the same definitions across many different calendars)—though we’ll have to see if this second point is truly relevant.

PR #693, opened 2024-10-25
feat(canada): add commons on the proper calendar of Canada

emagnier:
- Added commons

### DIFF
--- a/rites/roman1969/src/particular-calendars/canada.ts
+++ b/rites/roman1969/src/particular-calendars/canada.ts
@@ -1,3 +1,4 @@
+import { CommonDefinition as Common } from '../constants/commons';
 import { PatronTitle } from '../constants/martyrology-metadata';
 import { Precedences } from '../constants/precedences';
 import { CalendarDef } from '../models/calendar-def';
@@ -19,6 +20,7 @@ export class Canada extends CalendarDef {
     andre_bessette_religious: {
       precedence: Precedences.ProperMemorial_11b,
       dateDef: { month: 1, date: 7 },
+      commonsDef: Common.None,
     },
 
     // src: mr_fr_2021_ed3
@@ -30,6 +32,7 @@ export class Canada extends CalendarDef {
     marguerite_bourgeoys_virgin: {
       precedence: Precedences.ProperMemorial_11b,
       dateDef: { month: 1, date: 12 },
+      commonsDef: Common.None,
     },
 
     // src: mr_fr_2021_ed3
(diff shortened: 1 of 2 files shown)


rites/roman1969/src/particular-calendars/canada.ts
```diff
@@ -19,6 +20,7 @@ export class Canada extends CalendarDef {
     andre_bessette_religious: {
       precedence: Precedences.ProperMemorial_11b,
       dateDef: { month: 1, date: 7 },
+      commonsDef: Common.None,
```
TobiTenno, line 23: would it be possible to code these so the default is `Common.None` so we don't have boilerplate in case we ever need to change how commons work?

for example, if we expand more to be in the definition of `commonsDef`, like if it needs to be array-like someday?

emagnier: I mainly wanted to make an explicit distinction between: 
- there is no defined common in Romcal: `communsDef: undefined`, which is the default.
- or the missal does not define a common for that day, because specific content is already provided: `commonsDef: Common.None`.

So I think it's good to keep it that way.

emagnier: > for example, if we expand more to be in the definition of `commonsDef`, like if it needs to be array-like someday?

The `commonsDef` property accepts an object of type `CommonDefinition` or a list of `CommonDefinition`. So it’s already possible.

```ts
commonsDef?: CommonDefinition | CommonDefinition[];
```

I defined it this way to make it as easy as possible to set up at the data definition level. However, in an object of type `LiturgicalDayDef`, the `commonsDef` property is always returned as an array (to simplify data consumption).

TobiTenno: `there is no defined common` and `the missal does not define a common`... is the idea that `undefined` should eventually not be possible once all of the calendars have had updates applied to them?

Is the plan, then, to disallow `undefined` once all calendars have been touched?

emagnier: > is the idea that `undefined` should eventually not be possible once all of the calendars have had updates applied to them?

That would be the ideal world, yes! But this would mean that, in addition to updating all current calendars, we would expect every new calendar to have commons defined for each of its entries as well. This raises the bar for contributions, knowing that we don’t always have access to the missals specific to each liturgical calendar (or we would have to infer commons based on the biographies of each saint, which isn’t always straightforward either).

And a proper calendar with no defined commons would imply that none of its entries have a common (suggesting that specific content is available for these liturgical days), which could potentially be misleading information.

I’m not opposed to having `Common.None` as the default value, but perhaps this could be done later, once more calendars are completed? What do you think?

In the meantime, it would remain easier to detect omissions in a proper calendar (I was actually thinking of enhancing the `data:checks` task to detect calendars where some commons for proper liturgical days aren’t defined).

TobiTenno (approved):

PR #744, opened 2025-02-09
fix(general-roman): avoid christmas season psalterWeek reset

### LINKED ISSUE #723: Error with "psalterWeek": "WEEK_2" when its WEEK_1 and
Error from january 2-4, this error: `"psalterWeek": "WEEK_2"` and it should be `"psalterWeek": "WEEK_1"`

And same for this current week (6-11): `"psalterWeek": "WEEK_3"` and should be `"psalterWeek": "WEEK_2"`

TobiTenno:
PR started to fix #723


rites/roman1969/__tests__/christmastide-psalter.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  }, 1000000);
+  test('Psalter weeks after Christmas should match lh.kbs.sk', async () => {
+    let results = [];
+
+    // eslint-disable-next-line no-restricted-syntax
+    for await (const date of Object.keys(romcal)) {
+      const day0 = romcal[date][0];
+
+      const dateParts = date.split('-').map(Number);
+      const psalterWeekRomcal = +(/[0-9]+$/.exec(day0.cycles.psalterWeek)?.[0] ?? 0);
```
tukusejssirs: Why did you add `?? 0`? AFAIK, all celebrationsshould havepsalter week defined.

TobiTenno: I believe this is a side effect of exec's types

TobiTenno: yeah, exec always has a possible ` | undefined` union on its types, which i don't think i can override, thus the `?.[0]` as well

tukusejssirs: Well, I always use `?.[0]`, but never `?? 0` in this case, that’s why I asked about it. From my PoV, it is redundant, as ATM romcal always defines psalter week for every celebration (which I think is not always required, but that is OT in here), therefore, we could remove both `?.[0]` and `?? 0`.

In the end, it does not matter. :man_shrugging:

TobiTenno: removing `?.[0]` would give us an array back, so we'd still have to pull the value out

`?? 0` would make us still need to handle `undefined` as a value. doing it this way keeps the type of the value to just `number`. In javascript without types, it's fine to replace  `?.[0]` with `[0]`; and to remove `?? 0`, given that we think it shouldn't be an issue, would work... but this isn't javascript

tukusejssirs: > removing `?.[0]` would give us an array back, so we'd still have to pull the value out

Oh, my bad, I meant only the question mark and the dot. :wink:

> `?? 0` would make us still need to handle `undefined` as a value.

Yes, it would _if romcal would not provide psalter week number for some dates_, which it does not.

That said, we can keep it. :wink:

> but this isn't javascript

Yeah, I forgot about that. :+1:

TobiTenno: the `?? 0` has nothing to do with romcal providing the psalter week number. the return type of `Regexp.prototype.exec` includes `null` [[docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec)]

tukusejssirs: Yeah, I got that, thanks, @TobiTenno. I you forgot you wrote the test in TS, while I wrote my script in JS. :wink:

TobiTenno: okiedoke. feel free to push to the branch when trying to fix it and update the title and whatnot to whatever you need

tukusejssirs: I will, however, for now, I am trying to find a way to get the data in order to calculate it correctly.

I see two options:

1. either get access to the user-input year in `ProperOfTime`; or
2. get access to the season name (or its ID) in `CyclesMetadata`.

Option 1 seems to be dead, the option 2 seems promissing (and probably better), however, `RomcalCalendarMetadata` does not provide this info, therefore, we might want to add a prop to `RomcalCalendarMetadata`.

TobiTenno: should we add a function prop to just dump out the day before its inserted in `addLiturgicalDay` so fields can be read and modified before it's pushed into the calendar?

tukusejssirs:
@TobiTenno, I have fixed the issue. However, as you created the PR, I cannot request a review via GitHub UI.

rites/roman1969/__tests__/christmastide-psalter.test.ts
```diff
@@ -0,0 +1,43 @@
+import { LiturgicalCalendar, Romcal } from '@src/rite-roman1969';
+
+import { cache } from './fixtures/lh.kbs.sk.fixture';
+
+describe('Psalter weeks after Christmas', () => {
+  let romcal: LiturgicalCalendar;
+
+  beforeAll(async () => {
+    romcal = await new Romcal({ scope: 'liturgical' }).generateCalendar(2025);
```
tukusejssirs: We might want to add a test for the `gregorian` scope.

rites/roman1969/__tests__/fixtures/lh.hbs.sk.fixture.ts
tukusejssirs, line 1: I wanted to rename this file. :wink: This file needs to be removed.

rites/roman1969/__tests__/locales.test.ts
```diff
@@ -42,7 +42,7 @@ describe('Testing localization functionality', () => {
               expect(day?.name).toBe(d.value);
             } else {
               /* eslint-disable-next-line no-console */
-              console.log(
+              console.debug(
```
tukusejssirs: This is an unrelated change.

Anyway, a proper fix would be to test all celebration names regardless whether they are celebrated in a year or not.

rites/roman1969/src/types/liturgical-day.ts
```diff
@@ -221,6 +221,13 @@ export type RomcalCalendarMetadata = {
    * The last day of the current liturgical season.
    */
   endOfSeason: string;
+
+  /**
+   * Whether the Nativity of the Lord occurs on a weekday
+   *
+   * @remarks The property is defined only dates during Christmas Time.
+   */
+  isNativityOfTheLordOnWeekday?: boolean;
```
tukusejssirs: You might want to change the property name.

We might want to remove it from the final romcal object.

rites/roman1969/src/utils/dates.ts
```diff
@@ -293,8 +293,9 @@ export class Dates {
     year = this.#isLiturgicalYear ? this.#year - 1 : this.#year
   ): Date | null => {
     const id = `${year}_${dayOfOctave}`;
-    if (this.#weekdayWithinOctaveOfChristmas[id] !== undefined) {
-      return this.#weekdayWithinOctaveOfChristmas[id];
+    const existing = this.#weekdayWithinOctaveOfChristmas[id];
```
tukusejssirs: I did not change this file. Is this really necessary? :thinking:

tukusejssirs (changes requested):

rites/roman1969/__tests__/fixtures/lh.hbs.sk.fixture.ts
TobiTenno: no, i made this run this way because the test takes over 2 minutes if i don't have it cached

rites/roman1969/__tests__/christmastide-psalter.test.ts
TobiTenno: feel free

rites/roman1969/__tests__/locales.test.ts
TobiTenno: i wasn't changing what was tested, and you're welcome to change the code i added if you like, these were things that made no sense to be plain logs. feel free to remove it.

rites/roman1969/src/utils/dates.ts
TobiTenno: yes, we can all tell from the commits you didn't change it... it's not _necessary_ but it does help understand the point, given that the same reference is made multiple times, and a few of these lines are bad patterns (returning assignments) that reference this, and i didn't want to make logical changes about it....

rites/roman1969/__tests__/fixtures/lh.hbs.sk.fixture.ts
TobiTenno: i see what you're saying now. not sure why you're telling me to remove the file you failed to rename?

rites/roman1969/src/types/liturgical-day.ts
TobiTenno: you could just make it private or available through a getter, in which case it shouldn't be serialized.

rites/roman1969/src/models/calendar.ts
```diff
@@ -100,6 +105,7 @@ export class Calendar implements BaseCalendar {
       endOfSeason: endOfSeason ? endOfSeason.toISOString().substr(0, 10) : '',
       startOfLiturgicalYear: startOfSeasonsDic[Season.Advent].toISOString().substr(0, 10),
       endOfLiturgicalYear: endOfSeasonsDic[Season.OrdinaryTime].toISOString().substr(0, 10),
+      ...(isChristmasTime ? { isNativityOfTheLordOnWeekday: nativityOfTheLordDow !== 0 } : {}),
```
TobiTenno: seems like a decent solution...

rites/roman1969/__tests__/fixtures/lh.hbs.sk.fixture.ts
tukusejssirs: Because now the fixture is duplicated. :wink: I can take care of tomorrow.

Screenshot_20250211_031420_GitHub

rites/roman1969/src/models/calendar.ts
tukusejssirs: Decent as in you could improve it? :thinking:

rites/roman1969/src/types/liturgical-day.ts
TobiTenno: this is pretty much what i was trying to avoid, you can't really not expose it if the type says you do... i don't think we can avoid having it either on the final object if the type dictates it will be defined during Christmas Time.

TobiTenno: imo we should expose the Seasons here again and use those to determine the psalter week rather than determining a boolean in the calendar instance to then be used in the liturgical day. it would make sense to pass metadata for allowing the logic of what the day is precisely so the day can handle all the logic there.

TobiTenno: pushing something shortly.

TobiTenno: pushed

rites/roman1969/src/models/cycles-metadata.ts
```diff
@@ -68,11 +70,19 @@ export class CyclesMetadata implements BaseCyclesMetadata {
...
+    // 1. During the first four days of Lent (Ash Wednesday to the next Saturday),
     // which are in week 4, to start on week 1 after the first sunday of lent.
+    // 2. According to GILH §133, we should not restart the psalter week cycle
+    // for Christmas Time. This is due the fact that the first day of the season
+    // (December 25) does not always start on Sunday, unlike other seasons.
     const weekIndex = (calendar.weekOfSeason % 4) - 1;
-    const psalterWeek = PSALTER_WEEKS[weekIndex > -1 ? weekIndex : 3];
+    const useChristmasPsalterWeek = Boolean(
+      calendar?.seasons.includes(Season.ChristmasTime) ? new Date(`${year}-12-25T00:00:00.000Z`).getUTCDay() : 0
+    );
```
tukusejssirs, line 84: I like this solution using `Boolean()` (it didn’t occurred to me). :100:

tukusejssirs (approved):
Well done, @TobiTenno! Your solution is much cleaner! :pray:

Fernando-Perles:
How can we get the fix?
I mean, how to check in which version is delivered the fix for the users to get it?
Kind regards

tukusejssirs:
@Fernando-Perles, how did you install romcal?

TobiTenno:
@Fernando-Perles version 3.0.0-dev.94 was released from this change. you can use it by updating your local copy of romcal and all associated romcal packages to 3.0.0-dev.94. i'll be updating the dev site shortly to pick up the new version.

Fernando-Perles:
Many thanks.

Fernando-Perles:
@tukusejssirs and @TobiTenno 
Unfortunately we have tested the fix and it does not work.
Checking    https://romcal.js.org/#/calendar/es/general-roman/2026/1
We still see  "psalterWeek":string"WEEK_3"  on January, 8th, as an example.
Are we doing anything wrong? Does that URL point to the latest commit in dev?
Kind regards

tukusejssirs:
@Fernando-Perles, as I wrote in [this comment](https://github.com/romcal/romcal/issues/723#issuecomment-2650869935), that website still needs to be updated. The website app is a separate one from the library, however, it uses romcal as a source.

You can watch romcal-examples#14 to be notified when it is done.

Anyway, at the bottom of the website you can see the romcal version used to build the website. It must be at least `v3.0.0-dev.94` to contain the fix.



(We thought you used the romcal library directly, however, it seems like you just the website.)

Fernando-Perles:
We use the library in our application. The web is used just to check things out.

Thanks for the clarifications. Very useful.

Fernando-Perles:
Hi again.
We have integrated in our application 3.0-dev-95 and it still fails.
Selecting jannary 8th 2026, as an example, it still indicates salterWeek 3

Could you please double check?

Rgds

TobiTenno:
i'll reopen #723 until i get it verified

tukusejssirs:
@Fernando-Perles, I have just tested this with the latest `dev` branch as well as using `romcal@3.0.0-dev.97` and it works as expected:

```bash
# Build from source
# Tested with Node v20.18.3  and NPM v10.8.2
nvm use
pnpm clean
pnpm build

# Install via a package manager
path="$HOME/romcal_test"
mkdir "$path"
cd "$path"
pnpm init
pnpm i romcal@dev
```

```js
const {Romcal} = require('./rites/roman1969/dist/cjs/romcal.js')  // Built from source
const {Romcal} = require('romcal')  // Installed via a package manager

console.log((await new Romcal().generateCalendar(2026))['2026-01-08'])
```

Output:

```js
[
  R {
    id: 'thursday_after_epiphany',
    date: '2026-01-08',
    dateDef: {
      dateFn: 'weekdayAfterEpiphany',
      dateArgs: [Array],
      yearOffset: 0
    },
    dateExceptions: [],
    alternativeTransferDateDefs: [],
    precedence: 'WEEKDAY_13',
    rank: 'WEEKDAY',
    allowSimilarRankItems: false,
    isHolyDayOfObligation: false,
    isOptional: false,
    i18nDef: [ 'seasons:christmas_time.after_epiphany', [Object] ],
    seasons: [ 'CHRISTMAS_TIME' ],
    periods: [ 'DAYS_FROM_EPIPHANY', 'CHRISTMAS_TO_PRESENTATION_OF_THE_LORD' ],
    colors: [ 'WHITE' ],
    commonsDef: [],
    martyrology: [],
    titles: [],
    calendar: {
      weekOfSeason: 3,
      dayOfSeason: 15,
      dayOfWeek: 4,
      nthDayOfWeekInMonth: 2,
      startOfSeason: '2025-12-25',
      endOfSeason: '2026-01-11',
      startOfLiturgicalYear: '2025-11-30',
      endOfLiturgicalYear: '2026-11-28',
      seasons: [Array]
    },
    cycles: G {
      properCycle: 'PROPER_OF_TIME',
      sundayCycle: 'YEAR_A',
      weekdayCycle: 'YEAR_2',
      psalterWeek: 'WEEK_2'
    },
    fromCalendarId: 'ProperOfTime',
    fromExtendedCalendars: []
  }
]
```

Fernando-Perles:
Many thanks.
It is clear it is working fine for you.
We will review our implementation again.
We select the version for Spain. I guess that has nothing to do. Right?

Rgds.

TobiTenno:
That shouldn't matter, as the universal calculations apply to the psalter and non-localized data

Fernando-Perles:
Many thanks for the confirmation.

PR #746, opened 2025-02-11
fix(package): required dependencies not bundled

TobiTenno:
fix an issue where we'd have to install `i18next` as part of any consumers because it's missing from the bundle (encountered this upgrading romcal-examples


tukusejssirs (approved):
LGTM.

PR #762, opened 2025-02-11
ci: switch from directories to versioning strategy

TobiTenno:
ref: https://github.com/dependabot/dependabot-core/issues/5226#issuecomment-1179434437

trying the dependabot part a different way


tukusejssirs (approved):
LGTM.

PR #793, opened 2025-03-25
fix(package): buildable on ts 5.2.3+

### LINKED ISSUE #389: Type issues when using `typescript@4.8`
I’ve just tried to upgrade `typescript` and noticed that:

- [x] the type of `locales` is wrong:
  - we have set the property name to `Lowercase<string>`, however, the property names are in PascalCase;
  - I suggest to change it to `Record<string, Locale>`;

  https://github.com/romcal/romcal/blob/445ca5ad0f390446748a7c9f51dd156a18f68aa3/lib/locales/index.ts#L18-L20

- [x] particular calendar type extends `CalendarDef`, however, the class object is of unique type (e.g. `Americas` class has `Americas` type`:
  - then in `particular-calendars/index.ts`, `particularCalendars` is set to `Record<string, typeof CalendarDef>`, which raises error;
  - another issue in here is the type of `inputs` (`Inputs` in `americas.ts` and `CalendarDefInputs` in `calendar-def.ts`;
  - I have no idea how could we fix this (any suggestion welcome); maybe we should simply add missing properties from particular calendars to `CalendarDef`;
  - TS error:

    ```bash
    TSError: ⨯ Unable to compile TypeScript:
    lib/particular-calendars/americas.ts:7:3 - error TS2416: Property 'inputs' in type 'Americas' is not assignable to the same property in base type 'CalendarDef'.
      Type 'Inputs' is not assignable to type 'CalendarDefInputs'.

    7   inputs: Inputs = {
        ~~~~~~
    lib/particular-calendars/americas.ts:8:5 - error TS2322: Type '{ our_lady_of_guadalupe: { customLocaleKey: string; precedence: "PROPER_FEAST__PRINCIPAL_PATRON_OF_A_REGION_8C"; titles: { append: PatronTitles.PatronessOfTheAmericas[]; }; }; }' is not assignable to type 'Inputs'.
      Object literal may only specify known properties, and 'our_lady_of_guadalupe' does not exist in type 'Inputs'.

      8     our_lady_of_guadalupe: {
            ~~~~~~~~~~~~~~~~~~~~~~~~
      9       customLocaleKey: 'our_lady_of_guadalupe_patroness_of_the_americas',
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ...
     11       titles: { append: [PatronTitles.PatronessOfTheAmericas] },
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     12     },
        ~~~~~
    ```

  https://github.com/romcal/romcal/blob/445ca5ad0f390446748a7c9f51dd156a18f68aa3/lib/particular-calendars/americas.ts#L6

- [ ] TS now prefers using `const = 's' as const` instead of `const: 's' = 's'`, therefore new need to update it in `lib/constants/precedences.ts`;
- [ ] although it is an issue _per se_, however, it makes the code cleaner: we should replace `{En: En}` with `{En}` in `lib/locales/index.ts`, `lib/particular-calendars/index.ts` and maybe in other places;

TobiTenno:
## Description

issue seems to have been our expectation not to emit
which it doesn't seem like it is, but is falsely claiming that it is (see emit status message added to logs
issues noticed previously don't seem to be issues anymore. if they are, the build isn't catching them.
would appreciate a thorough build & test if anyone's got a chance

## Related

closes #389 
closes #775 
closes #803

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -6,7 +6,7 @@ This list recognizes contributors to romcal past and present.
 
 ### Active Contributors
 
-As of September 2021
+As of March 2025:
 
 - [Etienne Magnier](https://github.com/emagnier)
 - [Tukusej’s Sirs](https://github.com/tukusejssirs)
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "rimraf": "^6.0.1",
         "ts-node": "^10.9.2",
         "tslib": "^2.6.2",
-        "typescript": "~5.2.2"
+        "typescript": "~5.8.2"
       },
       "engines": {
         "node": ">=18.0.0"
@@ -11944,7 +11944,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.2.2",
+      "version": "5.8.3",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.8.3.tgz",
+      "integrity": "sha512-p1diW6TqL9L07nNxvRMM7hMMw4c5XOo/1ibL4aAIGmSAt9slTE1Xgw5KWuof2uTOvCg9BY7ZRi+GaF+7sfgPeQ==",
       "devOptional": true,
       "license": "Apache-2.0",
       "bin": {
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "module": "rites/roman1969/dist/esm/romcal.js",
   "typings": "rites/roman1969/dist/index.d.ts",
   "scripts": {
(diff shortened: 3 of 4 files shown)


TobiTenno:
@tukusejssirs you missed this one

emagnier (approved):
I just tested it on my end (build & test), and LGTM 👍

PR #794, opened 2025-03-25
fix(l10n): upgrading i18next to v24

TobiTenno:
upgrade i18next to v24

closes #771 

TODO:
- [x] v22 upgrade
- [x] v23 upgrade
- [x] v24 upgrade

PR #806, opened 2025-04-11
style(eslint): upgrade eslint & lint pr title

TobiTenno:
trying to upgrade eslint to latest version, needed to get some code up and available to attach to the issue i'm logging on eslint/eslint to get feedback

closes #759


.github/workflows/romcal-ci.yml
```diff
@@ -30,10 +30,16 @@ jobs:
         if: github.event_name == 'push'
         run: npx commitlint --last --verbose
 
-      - name: Validate PR commits with commitlint
+      - name: Validate with commitlint (pull_request)
```
tukusejssirs, line 33: :pick: What was wrong with the former version? It was more descriptive IMO.

.github/workflows/romcal-ci.yml
```diff
@@ -30,10 +30,16 @@ jobs:
...
         if: github.event_name == 'pull_request'
         run: npx commitlint --from ${{ github.event.pull_request.head.sha }}~${{ github.event.pull_request.commits }} --to ${{ github.event.pull_request.head.sha }} --verbose
 
+      - name: Lint Title (pull_request)
```
tukusejssirs, line 37: :pick: How about this? We don’t use Title Case in the CI action names, therefore, I think we should stick with this convention (unless you re-case them all). And `title` is quite general term, therefore, I personally miss some level of specificity.

```suggestion
      - name: Lint PR titles (pull_request)
```

.lintstagedrc.yaml
```diff
@@ -8,6 +8,7 @@
 package*.json:
   - npx prettier --write
   - npm dedupe
+  - npx sort-package-json
```
tukusejssirs, line 11: :pick: There is the [`@eslint/json`](https://www.npmjs.com/package/@eslint/json) ESLint plugin which [sorts](https://www.npmjs.com/package/@eslint/json#rules) JSON files. I think we should sort all files, not just `package.json`. Actually, I think we should lint all files committed to the repository.

I personally would prefer using ESLint + Prettier for fixing any stuff related to style and formatting. YMMV.

package.json
```diff
@@ -37,6 +37,11 @@
       "name": "Tukusej’s Sirs",
       "email": "tukusejssirs@protonmail.com",
       "url": "https://github.com/tukusejssirs"
+    },
+    {
+      "name": "Matej Voboril",
+      "email": "tobiah@pm.me",
+      "url": "https://github.com/tobitenno"
```
tukusejssirs: :pick: I think this should be a separate PR (as some other stuff too), as it is unrelated to the PR which talks about upgrading ESLint and linting PR title (which again are too separate features/chores which IMHO should be in two separate PRs).

That said, keep it as is now.

package.json
```diff
@@ -47,15 +52,19 @@
   "main": "rites/roman1969/dist/cjs/romcal.js",
   "module": "rites/roman1969/dist/esm/romcal.js",
   "typings": "rites/roman1969/dist/index.d.ts",
+  "workspaces": [
+    "rites/*",
+    "packages/*"
+  ],
```
tukusejssirs, line 53: :spiral_notepad: This causes PNPM to create separae workspaces. Nothing to do here, as I still wish to migrate to an Nx monorepo where this problem does not exist.

package.json
```diff
@@ -64,22 +73,29 @@
...
     "rimraf": "^6.0.1",
     "ts-node": "^10.9.2",
-    "tslib": "^2.6.2",
-    "typescript": "~5.2.2"
-  },
-  "dependencies": {
-    "i18next": "^21.6.0"
+    "tslib": "^2.8.1",
+    "typescript": "~5.2.2",
+    "typescript-eslint": "^8.29.1"
```
tukusejssirs: :spiral_notepad: I think this is a mess due to having deps defing in multiple `package.json` files. I don’t Jest in here, but I can see it in [`packages/config/package.json`](https://github.com/romcal/romcal/blob/4f509a72f77430a2e748ed2c2fe7de3056cc9aed/packages/config/package.json#L11-L12). Even more deps I see added to [`rites/roman1969/package.json`](https://github.com/romcal/romcal/blob/4f509a72f77430a2e748ed2c2fe7de3056cc9aed/rites/roman1969/package.json#L28-L62). If we continue using this monorepo structure, it will be even less maintainable in the future IMO.

I guess it makes sense now to put deps where they are used, however, you need to look at so many places when looking for them after a while. :thinking:

eslint.config.mjs
```diff
@@ -0,0 +1,213 @@
...
+    },
+    rules: {
+      '@typescript-eslint/no-non-null-assertion': 'off',
+
+      // tests don't export
+      'import/no-unused-modules': 'off',
+
+      // do not complain about importing dist localized calendars
+      'import/no-relative-packages': 'off',
+      'jest/no-conditional-expect': 'off',
```
tukusejssirs, line 129: :pick: I prefer leaving this on, see the [docs](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md). Yes, we should fix the troublesome tests.

rites/roman1969/src/models/liturgical-day-def.ts
```diff
@@ -342,8 +342,7 @@ export class LiturgicalDayDef implements BaseLiturgicalDayDef {
     if (input.titles) {
       martyrology.forEach((m, i) => {
         // TODO: refactor this to avoid non-null assertion
```
tukusejssirs: :pick: Could we call this `TODO` comment done if you added `if (input.titles)`? If yes, we should remove it.

tukusejssirs (commented):
LGTM, I just added some nitpicks (prefixed with :pick:) and notes (prefixed with :spiral_notepad:). If you don’t want to implement them, just tell me and I’ll approve the PR right away, but I wanted to give you some time to check out my comments.

package.json
tukusejssirs: This actually prevents correct usage of the repository while using PNPM. I could not correctly install `i18next@24.2.3` while reviewing #794 which cause build a error that `initAsync` does not exist because I still had `i18next@21.10.0` installed.

The issue is that PNPM does not support `workspaces` property and uses `pnpm-workspace.yaml` instead:

```bash
 WARN  The "workspaces" field in package.json is not supported by pnpm. Create a "pnpm-workspace.yaml" file instead.
 ```

However, removing it only allows installing deps, but building still fails:

```bash
> romcal@3.0.0-dev.97 build $repo_root
> npm run build -w=@internal/easter -w=@internal/rite-roman1969

npm error No workspaces found:
npm error   --workspace=@internal/easter --workspace=@internal/rite-roman1969
npm error A complete log of this run can be found in: $log_path
 ELIFECYCLE  Command failed with exit code 1.
```

---

Actually, this seems to be already broken in `dev` already, as we now specify workspaces using `-w` and also hard-code `npm`. :disappointed:

tukusejssirs (dismissed):

TobiTenno:
@tukusejssirs per your pnpm comments, we've decided previously to not explicitly use pnpm or model how this uses npm... and as noted, that's already the behavior in dev. given that we're using workspaces, it would be nigh-impossible to not break the usage of other package managers' custom usages while doing our best to build up the npm workspaces architecture

.github/workflows/romcal-ci.yml
TobiTenno: i specified it in a parenthetical separation to parallel with the same format in the step before it

.github/workflows/romcal-ci.yml
TobiTenno: there aren't any other Titles and we already specify `pull_request` as the scope. this is a distinction without a difference. i'd rather just remove the parenthetical, but this helps it be clear on a non-PR push why the step was skipped

package.json
TobiTenno: you're right, i'm sorry to document myself as a contributor in something i'm contributing.

package.json
TobiTenno: no, this is because of eslint's new linting format, unfortunately, which does not do shared configs as nicely as before. this was the primary structure recommended for usage by eslint itself when using workspaces.

the same dep can be in multiple package.json files in a workspace and is resolved, typically with the same resolvable version in the root `node_modules`

eslint.config.mjs
TobiTenno: that's not the context of this pr. if you'd like to go implement, that's fine, but it is a defacto standard format until it is replaced.

rites/roman1969/src/models/liturgical-day-def.ts
TobiTenno: sure, can do

.lintstagedrc.yaml
TobiTenno: we were already using it previously, it had been shifted around before. we don't have a pattern of using eslint for everything at present. as you've already mentioned with other changes, that would be outside the scope of this PR.

emagnier (approved):
LGTM

Issue #376, opened 2022-10-02
Update particular calendar names


tukusejssirs:
ATM I am adding calendars for Slovak dioceses, however, I don’t we ever defined the filename and class naming rules.

Based on the current filenames and class names, I suggest to used the following rules:

- particular calendar filenames:
   - use `${continent}.${subcontinent}.${region}.${country}.${dioceseCity}.${parishCity}.${parishName}.${churchName}.ts` format for filenames of particular calendars, where:
      - `continent`: is the continent name:
         - `america` for North/Central/South Americas;
         - `africa`;
         - `eurasia` for Asia and Europe;
         - `australia` for Australia and Oceania;
         - `antarctica` for Antartica;
      - `subcontinent`: is the subcontinent name;
         - `north-america`;
         - `central-america`;
         - `south-america`;
         - `europe`;
         - `asia`;
      - `region`: for regional calendars (like Austria-Germany-Switzerland);
      - `country`: is the country name;
      - `dioceseCity`: is the diocese seat (city) name;
      - `parishCity`: parish city/town/vilage name;
      - `parishName`: parish name (patronage) when there are multiple parishes in a city/town/vilage;
      - `churchName`: a church name when a particular celebration is celebrated in a parcticular church only;
   - each part (one of the six format items of a filename) should be in kebab-case;
- particular calendar class names:
   - same format but with underscores instead of dots and without `.ts` extension should be used for class names in particular calendars (i.e. `${continent}_${subcontinent}_${region}_${country}_${dioceseCity}_${parishCity}_${parishName}_${churchName}`);
   - each part (one of the six format items of a filename) should be in PascalCase/UpperCamelCase;
- additional rules and notes:
   - we should omit any level as needed, based on the calendar inheritance of the particular calendar;
   - I decided to combine Americas, and Asia and Europe to America and Eurasia for simplicity purposes, because:
      - [America] some celebrations are celebrated thoughout Americas, others only in North/Central/South Americas (AFAIK);
      - [Eurasia] there are separate patrons of European and Asian part of Russia, thus Russia should be part of both Europe and Asia (in national calendar; thus `eurasia.russia.ts`), however on regional level (diocese, parish) it should be part of either Asia or Europe only (thus `eurasia.europe.russia.ts` or `eurasia.asia.russia.ts`):
         - this is not ideal; @emagnier, you might come up with a better solution;
         - if we won’t find a better solution, then (either):
            - we should add a way to to exclude a calendar from inheritance;
[...]

PR #834, opened 2025-05-19
refactor(calendar): reorganize calendar definition files

emagnier:
A step in the direction of the ideas shared by @tukusejssirs here: [https://github.com/romcal/romcal/issues/376](https://github.com/romcal/romcal/issues/376).

The goal here is to better organize all these calendars (with better file naming) and prepare for new types of calendars (cities, particular calendars for religious communities, etc.).

Proposed structure:

* `calendars\`
  * `general-roman\`
    * `general-roman.ts`
  * `regions\`
    * contains calendars for large world regions and continents
  * `countries\`
    * `france\` (as an example)
      * `france.ts`
      * `archdiocese-of-paris\`
        * `archdiocese-of-paris.ts`
        * `city-of-paris.ts`
  * `communities\`
    * will contains the particular calendars of religious communities.
    
This PR is also a preparation for two upcoming PRs:

* Allow defining a calendar while selecting multiple parent calendars (instead of just one as currently).
* Allow inserting general calendars specific to religious communities (between the general Roman calendar and particular calendars).

For now, the corresponding NPM packages (and package names) for each of these calendars remain unchanged (so no breaking changes yet). Though this could also use some refinement, in my opinion.


TobiTenno (changes requested):
files with the same name as their parent either:
- should replace their parent if there are no sibling files (as is the case in most of the countries) 
- become `index.ts`, so that it is the default resolution for `rites/roman1969/src/particular-calendars/{country}`.

having to double specify the country is untenable long-term and looks really bad
in short, i don't think those files should be nested, personally, as they are now.
if that is to future-proof for more files, make them `index.ts`

`rites/roman1969/src/particular-calendars/austria/austria` as a path makes no sense. even if we had a structure like...
- `rites/roman1969/src/particular-calendars/austria/index`
- `rites/roman1969/src/particular-calendars/austria/prussia` (it's an example)

this would make far more sense than `austria/austria`

as the imports would, sensibly, be
- `import {blah} from 'rites/roman1969/src/particular-calendars/austria'`
- `import {blah} from 'rites/roman1969/src/particular-calendars/austria/prussia'`

emagnier:
Thanks for your feedback. Yes, your suggestions definitely make sense for cleaner and more maintainable paths. It’s obvious, and I should have done it that way from the start 😄

- I kept a folder for each country and renamed the main files to `index.ts`, so the structure stays visually consistent (whether there are diocesan calendars or not).
- I didn’t create separate folders for each diocese in France (I have less of a strong opinion on that). We’ll see how it goes in practice. In any case, if we ever need to nest a diocesan calendar in a folder (to add a city calendar for exemple), the path won’t need to change.

TobiTenno:
agreed, i think individual sub-folders for diocese or sub-regions should only be created as needed. like if each state in the US had separate calendars, that would make sense, but we largely don't, ours only vary by what is already configuration.

TobiTenno:
I like it now, and I'm going to approve it, but given the change, i'd also like to propose (since we're using exported individual members already, and not default exports on calendars) that we also export all "child" calendars from the parent in each country, i.e. exporting Lyons from France as well, this would make the large page of all the imports much cleaner and allow the logical grouping we're looking for, i.e. `import { France, France_Lyon } from '{...}/france'`

TobiTenno (approved):

emagnier:
Done 👍 . However, we'll need to be careful going forward, whenever a new child calendar is added, it should be exported both in the country's calendar and in the root `index.ts` of all calendars.

PR #835, opened 2025-05-19
feat(calendar): enforce full hierarchy for particular calendar definitions

emagnier:
Follow-up to this PR: [https://github.com/romcal/romcal/pull/834](https://github.com/romcal/romcal/pull/834) (so do not review the first commit).

Previously, only the parent calendar definition was required.
Now, the full hierarchy must be explicitly specified. This is slightly more verbose (since the complete hierarchy must be defined in each calendar), but it should help clarify and better modularize calendar definitions.

💡 Of course, in the `parentCalendars` property, the `GeneralRoman` calendar doesn't need to be specified, as it is the default base for all particular calendars.

This change allows, for example, handling the case of the calendar for Russia, which spans two continents (Europe and Asia). With this, it's now possible to define:

* A calendar for `Russia`
* An `EuropeanRussia` calendar (containing Europe-specific celebrations, included in the Russia calendar)
* An `AsianRussia` calendar (same idea for Asia)


TobiTenno (approved):
this makes sense to me, and i'm approving, but please do #834 first

TobiTenno:
looks great once you rebase, @emagnier

PR #878, opened 2025-06-30
feat(l10n): detect missing name in en.ts

### LINKED ISSUE #877: [improvement] Built error message "Locale ID 'names:${id}' is missing in the calendar '${calendar.name}'"
Build process (npm run build)  show me the message "Error: Locale ID 'names:dominic_of_the_mother_of_god_barberi_priest' is missing in the calendar 'England'."  (this was an intentional error), I look on "England" calendar  for the mistake, and nothing.
Take a look on '/rites/roman1969/build/bundle.ts' to understand what is happening.
This message is  inside a IF statement.

>         if (locale.names && Object.prototype.hasOwnProperty.call(locale.names, id)) {
>             update[id] = locale.names[id];
>           } else if (locales.En.names && Object.prototype.hasOwnProperty.call(locales.En.names, id)) {
>             update[id] = locales.En.names[id];
>           } else {
>             throw new Error(`Locale ID 'names:${id}' is missing in the calendar '${calendar.name}'.`);
>           }

The message tests LOCALES EN, but refers to a locale where is used.
Locale EN should be have ALL 'names' , it is OK, but the message did not show it. 
SUGGESTIONs to line 165: 
1) throw new Error(`Locale ID 'names:${id}' is missing in calendar locale en.ts.`);
2) throw new Error(`Locale ID 'names:${id}' SHOULD BE present in the calendar locale en.ts.`);

Shoud to refers to locale/en.ts file

TobiTenno:

add check for missing name from en.ts (thanks, @RonaldoSCouto)
closes #877

## Concept

Localization

## Example code

```typescript
// Comment anything from rites/roman1969/src/locales/en.ts#names
// ex. comment line 804 (most_holy_body_and_blood_of_christ)
// most_holy_body_and_blood_of_christ: 'The Most Holy Body and Blood of Christ',
```

run `npm run build`

```sh
Error: Locale ID 'names:most_holy_body_and_blood_of_christ' is missing in the locale 'en'. Ensure the value is included in rites/roman1969/src/locales/en.ts.
```

### DIFF
--- a/rites/roman1969/build/bundle.ts
+++ b/rites/roman1969/build/bundle.ts
@@ -157,6 +157,11 @@ export const RomcalBundler = (): void => {
         .map((def) => def.i18nDef[0].substr(6))
         .reduce((obj: LocaleLiturgicalDayNames, id: string) => {
           const update: LocaleLiturgicalDayNames = { ...obj };
+          if (locales.En && !Object.prototype.hasOwnProperty.call(locales.En.names, id)) {
+            throw new Error(
+              `Locale ID 'names:${id}' is missing in the locale 'en'. Ensure the value is included in rites/roman1969/src/locales/en.ts.`
+            );
+          }
           if (locale.names && Object.prototype.hasOwnProperty.call(locale.names, id)) {
             update[id] = locale.names[id];
           } else if (locales.En.names && Object.prototype.hasOwnProperty.call(locales.En.names, id)) {


tukusejssirs (changes requested):
Thanks, @TobiTenno and @RonaldoSCouto.

The fix looks good to me, however, I think that such a check should not be part of the build/bundle script itself. The build/bundle process should be a simple call of a bundler (e.g. Rollup) and it shoud have a dependency (like `prebuild`) where we could check stuff like this. Ideally, such checks should be moved to unit tests and the building/bundling process should depend on the test phase (all tests must be successful).

@TobiTenno, I consider this a `fix` not a `feat`. Why did you chose `feat`? :thinking: Other than that, it can be merged.

TobiTenno:
As this makes an existing check give the correct message, and as you can see we already have a check for the name not existing in non-EN locales or a particular calendar. This is a more precise check for it not existing at a default locale level before checking in calendars. If I were to move all of them, then I would see that as a feature. @tukusejssirs

tukusejssirs:
> As this makes an existing check **_give the correct_** message, and as you can see we already have a check for the name not existing in non-EN locales or a particular calendar. This is a **_more precise check_** for it not existing at a default locale level before checking in calendars.

I’d still call it a `fix`, but I get your PoV now. We can keep it a `feat`.

> If I were to move all of them, then I would see that as a feature.

Now this comments says I am actually right to consider this PR a `fix`? However, in the PR title, you actually say it is a `feat` (`feat(l10n): detect missing name in en.ts`), that is why I suggested to change the scope.

---

All in all, if you want to keep it a `feat`, just let me know and I’ll immedially approve the PR.

TobiTenno:
considering we don't semantically version right now, it doesn't really make much of a difference until after v3 is merged. i'll change it if you feel it necessary, but i don't think it is necessary to change right now.

tukusejssirs (approved):

emagnier (commented):
Thank you for this improvement @TobiTenno 👍

Issue #175, opened 2020-03-01
Proof the strings in `en.ts` and create specific rules how to generate key names


tukusejssirs:
This issue deals with two tasks:
- proof the strings in `en.ts`:
   - this was the original issue;
- define how the key names are generated:
   - for the specific rules, see [this comment](https://github.com/romcal/romcal/issues/175#issuecomment-673048460) below.

---
_Original issue description:_

In liturgy, the Church usually translate the _first name_ (sometimes even the surname) into vernacular, e.g. in English, we use `Mary` instead of `Maria` or `John` instead of `Giovanni`.

Now in romcal in `en.js`, some of the keys (those not in the `general` nor any of the particular calendars of countries with English as official language; in general: those not in any official liturgical book written in English), we use foreign (non-English) names.

I know that some of those names are not translatable or they are used untranslated—those might need to be left untranslated. But I think that most of those names should be localised.

tukusejssirs:
Actually, I think we might create a list of first names of saints/blesseds used in romcal and translate that separately from saints/blesseds. When a person is defined, we would then use a name from the list of names.

@pejulian, @emagnier, what do you think?

tukusejssirs:
@pejulian, @emagnier, I am about to tend to this issue after I finish the Latin localisation.

My plan is to deal with multiple issues with the English locale(s) and the US particular calendar (which with serve as the base for the `en.ts` locale) and all the particular calendars of the British Isles (Enland & Wales, Irish) calendars (which if differ in terms of celebration names from `en.ts`, the differences will be put into either `en-GB.ts` or `en-IE.ts`).

These are the tasks I want to deal with in particular:
- fix 26+ localisation issues which I have found since merging #202 (duplicates, typos, missing titles, etc; I am quite sure I’ll find some more);
- check celebration names (l10n), as well as celebration definitions (in `src/calendars/*.ts`) of the following particular calendars:
   - US;
   - UK (England & Wales, Scotland, Northern Ireland); this should also fix #185);
   - Ireland.

---

Now, @emagnier recently said I he would like to keep the key names constant (i.e. once set, they won’t change) and also that the key names should be stripped down to bare minimum, therefore I’d like to do that to, but I need some feedback on this (i.e. some kind of approval that I may proceed).

My proposal is as follows:
- from the key names, remove canonisation level (saint/blessed; #174);
- from the key names, remove all the titles (like priest or religious) except for the titles necessary to make a distinction between two or more persons:
[...]

PR #917, opened 2025-08-24
feat(general-roman): teresa_of_calcutta_virgin

### LINKED ISSUE #916: New feast day in General Roman Calendar
On 11 February 2022, Pope Francis made a new feast day for St Teresa of Calcutta, Virgin as optional memorial worldwide.

TobiTenno:

closes #916 

## Concept

Calendar Definition (particular entry promoted to general roman calendar, and a change of title)

## Example code

As normal

### DIFF
--- a/rites/roman1969/src/locales/sk.ts
+++ b/rites/roman1969/src/locales/sk.ts
@@ -926,8 +926,8 @@ export const locale: Locale = {
     teresa_benedicta_of_the_cross_stein_virgin: 'Svätej Terézie Benedikty od Kríža, panny a mučenice',
     teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe:
       'Svätej Terézie Benedikty od Kríža, panny a mučenice, patrónky Európy',
-    teresa_of_calcutta_virgin: 'Svätej Terézie z Kalkaty, panny a rehoľníčky',
-    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, ',
+    teresa_of_calcutta_virgin: 'Svätej Terézie z Kalkaty, panny',
+    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, panny',
     teresa_of_jesus_of_avila_virgin: 'Svätej Terézie od Ježiša (z Avily), panny a učiteľky Cirkvi',
     teresa_of_jesus_of_los_andes_virgin: 'Svätej Terézie od Ježiša z Los Andes, panny',
     teresa_of_portugal_religious: 'Blahoslavenej Terezy Portugalskej, rehoľníčky',
--- a/rites/roman1969/src/locales/sk.ts
+++ b/rites/roman1969/src/locales/sk.ts
@@ -926,8 +926,8 @@ export const locale: Locale = {
     teresa_benedicta_of_the_cross_stein_virgin: 'Svätej Terézie Benedikty od Kríža, panny a mučenice',
     teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe:
       'Svätej Terézie Benedikty od Kríža, panny a mučenice, patrónky Európy',
-    teresa_of_calcutta_religious: 'Svätej Terézie z Kalkaty, panny a rehoľníčky',
-    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, panny',
+    teresa_of_calcutta_virgin: 'Svätej Terézie z Kalkaty, panny a rehoľníčky',
+    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, ',
     teresa_of_jesus_of_avila_virgin: 'Svätej Terézie od Ježiša (z Avily), panny a učiteľky Cirkvi',
     teresa_of_jesus_of_los_andes_virgin: 'Svätej Terézie od Ježiša z Los Andes, panny',
     teresa_of_portugal_religious: 'Blahoslavenej Terezy Portugalskej, rehoľníčky',
--- a/rites/roman1969/src/locales/ta.ts
+++ b/rites/roman1969/src/locales/ta.ts
@@ -928,7 +928,7 @@ export const locale: Locale = {
     teresa_benedicta_of_the_cross_stein_virgin: 'சிலுவை ஸ்டெயினின் புனித தெரெசா பெனெடிக்டா - கன்னி, மறைசாட்சி',
     teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe:
       'சிலுவை ஸ்டெயினின் புனித தெரெசா பெனெடிக்டா - கன்னி, மறைசாட்சி, ஐரோப்பாவின் பாதுகாவலி',
-    teresa_of_calcutta_religious: 'கல்கத்தாவின் புனித அன்னை தெரெசம்மாள்',
+    teresa_of_calcutta_virgin: 'கல்கத்தாவின் புனித அன்னை தெரெசம்மாள்',
     teresa_of_jesus_jornet_ibars_virgin: 'இயேசு ஜோர்னட் இபார்சின் புனித தெரெசாள் - கன்னி',
     teresa_of_jesus_of_avila_virgin: 'அவிலா இயேசுவின் புனித தெரெசாள் - கன்னி, மறைவல்லுனர்',
     teresa_of_jesus_of_los_andes_virgin: 'லொஸ் அண்டேஸ் இயேசுவின் புனித தெரெசாள் - கன்னி',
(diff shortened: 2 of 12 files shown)


TobiTenno:
i tried running a snapshot update, but i'm having moduleNameMapper errors lately

same general setup as usual, but i'm running on an immutable base system :thinking: if anyone has a chance to try updating the snapshots for me, i'd appreciate it, @emagnier @tukusejssirs

rites/roman1969/src/locales/ta.ts
```diff
@@ -928,7 +928,7 @@ export const locale: Locale = {
     teresa_benedicta_of_the_cross_stein_virgin: 'சிலுவை ஸ்டெயினின் புனித தெரெசா பெனெடிக்டா - கன்னி, மறைசாட்சி',
     teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe:
       'சிலுவை ஸ்டெயினின் புனித தெரெசா பெனெடிக்டா - கன்னி, மறைசாட்சி, ஐரோப்பாவின் பாதுகாவலி',
-    teresa_of_calcutta_religious: 'கல்கத்தாவின் புனித அன்னை தெரெசம்மாள்',
+    teresa_of_calcutta_virgin: 'கல்கத்தாவின் புனித அன்னை தெரெசம்மாள்',
```
TobiTenno: the reference in Tamil already didn't include a title (at least as far as my rudimentary translation could handle)

TobiTenno:
i'm silly. being a fresh OS, i didn't have the package built yet :man_facepalming:

rites/roman1969/src/locales/sk.ts
```diff
@@ -926,8 +926,8 @@ export const locale: Locale = {
     teresa_benedicta_of_the_cross_stein_virgin: 'Svätej Terézie Benedikty od Kríža, panny a mučenice',
     teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe:
       'Svätej Terézie Benedikty od Kríža, panny a mučenice, patrónky Európy',
-    teresa_of_calcutta_religious: 'Svätej Terézie z Kalkaty, panny a rehoľníčky',
-    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, panny',
+    teresa_of_calcutta_virgin: 'Svätej Terézie z Kalkaty, panny a rehoľníčky',
+    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, ',
```
tukusejssirs: On lc.kbs.sk (semi-official litirgical calendar based on the Slovak ordo) used to have this celebration inscribed (and so did the hard copy of the ordo), however, last time it was seems to be in [2017](https://lc.kbs.sk/#20170905) as _Svätej Terézie z Kalkaty (Matky Terezy), panny a rehoľnice_ [Saint Teresa of Calcutta (Mother Teresa), Virgin and Religious]. That said, let’s follow the official decree in this case too.

```suggestion
    teresa_of_calcutta_virgin: 'Svätej Terézie z Kalkaty, panny',
    teresa_of_jesus_jornet_ibars_virgin: 'Svätej Terézie od Ježiša Jornet Ibars, panny',
```

rites/roman1969/src/locales/ta.ts
tukusejssirs:
```suggestion
    teresa_of_calcutta_virgin: 'கல்கத்தாவின் புனித அன்னை தெரெசம்மாள், கன்னி',
```

tukusejssirs (changes requested):
I would personally rename `teresa_of_calcutta_virgin` to `mary_teresa_of_calcutta_anjeze_gonxhe_bojaxhiu_virgin` to follow the rule to include full religious and civil name.

tukusejssirs:
> i'm running on an immutable base system

I wanted to tell you that this has nothing to do with the issue, but I’m glad you fixed it yourself. Anyway, what OS/distro are you running?

rites/roman1969/src/locales/sk.ts
TobiTenno: the second one doesn't have anything to do with this fix. do you have a source for St. Teresa of Jesus Jornet Ibars?

TobiTenno:
> I would personally rename [...]

I've not seen that rule anywhere and would be interested in reviewing that...

>  Anyway, what OS/distro are you running?

I'm running Bluefin GTS, based on fedora. it's got some decent utils available for easy setups.

rites/roman1969/src/locales/sk.ts
TobiTenno: hmmm, i think the diff here is desceptive, because the original already had virgin for teresa of jesus, so i'll resolve this one.

tukusejssirs:
> > I would personally rename [...]
> 
> I've not seen that rule anywhere and would be interested in reviewing that...

https://github.com/romcal/romcal/issues/175#issuecomment-673048460

Note that we might need to tweak it a bit and definitely document them in the docs.

tukusejssirs:
LGTM, apart from the ID name mentioned above. Do you want to change it in this PR? :thinking:

TobiTenno:
> > > I would personally rename [...]
> > 
> > 
> > I've not seen that rule anywhere and would be interested in reviewing that...
> 
> [#175 (comment)](https://github.com/romcal/romcal/issues/175#issuecomment-673048460)
> 
> Note that we might need to tweak it a bit and definitely document them in the docs.

right, so while that may have been agreed upon, it doesn't seem to be in any of our docs yet. feel free to add it, but for now, i just minimally modified the existing.

TobiTenno:
> LGTM, apart from the ID name mentioned above. Do you want to change it in this PR? 🤔

i'd rather not, given it wasn't already in use and this is as much of an addition as it is an elevation

and i'd disagree with that particular rule, given that's no how saints are titled in the promulgation for their feasts, at least in what we're specifying as our source here. i think, if anything, this is additional martyrology data we should add, not something we need in the id for the localized strings

tukusejssirs:
> right, so while that may have been agreed upon, it doesn't seem to be in any of our docs yet. feel free to add it, but for now, i just minimally modified the existing.

Yes, Etienne with the rules, but we definitely need to add them to the docs + add a check whether all celebration IDs match the specified format which requires adding all its parts to `martyrology.ts`.

> and i'd disagree with that particular rule, given that's no how saints are titled in the promulgation for their feasts, at least in what we're specifying as our source here. i think, if anything, this is additional martyrology data we should add, not something we need in the id for the localized strings

Those rules are about the celebration ID (the key name), not the value.

The rule was defined after I have learnt that there are two saints named `Saint Teresa of Jesus` (cf https://github.com/romcal/romcal/blob/bbacd7aff8218b9e98652982e757b0deaed4e4ac/rites/roman1969/src/locales/en.ts#L1037-L1038), so, as a precaution, I suggested add full names (both religious and civil names).

tukusejssirs (approved):

TobiTenno:
in that case, i think if we need it for disambiguation, that's fine, but making the maximally complex keys ahead of time could only make it harder for people to contribute, in my opinion, while only giving us the benefit of disambiguation when there are multiple that could have the same id.

TobiTenno:
honestly, we probably need a set of rubrics that are as exact as the official roman martyrology rubrics (which i have a copy of now!)

rites/roman1969/src/locales/sk.ts
tukusejssirs: > the second one doesn't have anything to do with this fix. do you have a source for St. Teresa of Jesus Jornet Ibars?
>
> hmmm, i think the diff here is desceptive, because the original already had virgin for teresa of jesus, so i'll resolve this one.

Yeah, you updated the wrong celebration, I just fixed it. I should have described my changes explixitly, sorry.

I have no source for St Teresa of Jesus Jornet Ibars.

tukusejssirs:
> in that case, i think if we need it for disambiguation, that's fine, but making the maximally complex keys ahead of time could only make it harder for people to contribute, in my opinion, while only giving us the benefit of disambiguation when there are multiple that could have the same id.

I think yhe key should be generated, not defined directly. In the end, I’d like to have a UI tool where non-tech users would be able to do CRUD ops on the romcal data (sources, martyrology/catalogue, calendars, celebrations, localisations) which would be then revirwed and a commit would be created (let’s say) once a day with the changes. Then. the ID would not matter at all and we could use even UUIDs.

> honestly, we probably need a set of rubrics that are as exact as the official roman martyrology rubrics (which i have a copy of now!)

What do you mean by these _rubrics_ specifically? I have a copy of Latin _Martyrologium Romanum_ (too) if that is what you mean. :thinking:

Anyway, I still think that celebration ID should not be considered a rubric or tied to anything official—it is just an ID after all.

emagnier:
The rules we agreed on back in 2020 are mainly there to make sure we can clearly identify the person behind an ID, and to keep a consistent ID structure across all people’s IDs.

Most of the time, these rules work fine, but they don’t need to be applied rigidly in every edge case, especially if it goes against common sense (like ending up with an extremely long ID, when a shorter and well-known name would already avoid any confusion, as is the case here).

The main goals should always be:
- IDs that are consistent and stable over time, and not meant to be renamed.
- IDs that clearly identify a person and briefly indicate who they are, without being overly exhaustive (religious, priest, abbot, virgin, etc.).
- These rules are meant as guidance, not strict requirements. The IDs themselves are not official, they’re simply practical identifiers for managing data in our system.

So I don’t think this is something we can fully automate or lint. It’s really on us to make sure IDs stay clear and useful, without getting overly strict about it.

And thanks @tukusejssirs for pointing out that old comment about that. This is definitely something we should document better, instead of it being lost in the GitHub issue tracker.

PR #971, opened 2025-11-12
fix(calendar): change precedence for dedication of Lateran Basilica

kvnang:

Change the rank of feast of the Lateran basilica to Our Lord's feast.

The feast of the Lateran basilica should be considered Our Lord's feast, as indicated in the [Calendarium Romanum (1969)](https://archive.org/details/CalendariumRomanum1969/page/n29/mode/2up) as well.

## Concept

 - Calendar Definition

### DIFF
--- a/rites/roman1969/src/calendars/general-roman/index.ts
+++ b/rites/roman1969/src/calendars/general-roman/index.ts
@@ -1458,7 +1458,7 @@ export class GeneralRoman extends CalendarDef {
 
     // src: mr_la_2008_ed3
     dedication_of_the_lateran_basilica: {
-      precedence: Precedences.GeneralFeast_7,
+      precedence: Precedences.GeneralLordFeast_5,
       dateDef: { month: 11, date: 9 },
       commonsDef: Common.None,
     },
--- a/rites/roman1969/src/calendars/general-roman/index.ts
+++ b/rites/roman1969/src/calendars/general-roman/index.ts
@@ -1456,7 +1456,7 @@ export class GeneralRoman extends CalendarDef {
       commonsDef: Common.Bishops,
     },
 
-    // src: mr_la_2008_ed3
+    // src: Calendarium Romanum 1969 https://archive.org/details/CalendariumRomanum1969/page/n29/mode/2up
     dedication_of_the_lateran_basilica: {
       precedence: Precedences.GeneralLordFeast_5,
       dateDef: { month: 11, date: 9 },
--- a/rites/roman1969/__tests__/__snapshots__/calendar-builder.test.ts.snap
+++ b/rites/roman1969/__tests__/__snapshots__/calendar-builder.test.ts.snap
@@ -31958,7 +31958,7 @@ exports[`Testing calendar snapshots General Roman Calendar in gregorian year 1`]
       "periods": [
         "LATE_ORDINARY_TIME",
       ],
-      "precedence": "GENERAL_FEAST_7",
+      "precedence": "GENERAL_LORD_FEAST_5",
       "rank": "FEAST",
       "seasons": [
         "ORDINARY_TIME",
@@ -68909,7 +68909,7 @@ exports[`Testing calendar snapshots General Roman Calendar in gregorian year 2`]
       "periods": [
         "LATE_ORDINARY_TIME",
       ],
-      "precedence": "GENERAL_FEAST_7",
+      "precedence": "GENERAL_LORD_FEAST_5",
       "rank": "FEAST",
       "seasons": [
         "ORDINARY_TIME",
@@ -106043,7 +106043,7 @@ exports[`Testing calendar snapshots General Roman Calendar in gregorian year 3`]
       "periods": [
         "LATE_ORDINARY_TIME",
       ],
-      "precedence": "GENERAL_FEAST_7",
+      "precedence": "GENERAL_LORD_FEAST_5",
       "rank": "FEAST",
       "seasons": [
         "ORDINARY_TIME",
@@ -146601,7 +146601,7 @@ exports[`Testing calendar snapshots General Roman Calendar in liturgical year 1`
       "periods": [
         "LATE_ORDINARY_TIME",
       ],
-      "precedence": "GENERAL_FEAST_7",
+      "precedence": "GENERAL_LORD_FEAST_5",
       "rank": "FEAST",
       "seasons": [
         "ORDINARY_TIME",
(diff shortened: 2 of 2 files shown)


TobiTenno:
@kvnang thanks for the PR!

 I think we need to update the docs to include the calendarium romanum entry, as we didn't have that on hand as of yet, and in that case, I agree. I was at a tridentine mass that sunday and i think it was the first time i'd ever heard the full title of the ArchBasilica, in which context of the full title _Major Papal, Patriarchal and Roman Archbasilica, Metropolitan and Primatial Cathedral of the Most Holy Savior and Saints John the Baptist and the Evangelist in Lateran, Mother and Head of All Churches in Rome and in the World_; I'd agree that this should be classed as a feast of the Lord. I think the final word would have to come from somewhere where we can find specifically that the title in all caps as a `FESTUM` indicates a feast of the Lord, since that delineation, at least to my remembrance of Latin, wasn't present in the Tabula dierum liturgicorum

TobiTenno:
@kvnang could you force push with an updated commit message so it picks up your title change as well? I'll see if i can get the PR workflow to be more sensitive to changes in the PR like titles as well and properly reflect those in the CI, but until then, just need a new push of your commit.

i.e.
```shell
git commit --amend 
# amend the title, then save with whatever editor you're using to change the message
git push --force-with-lease
```

TobiTenno (approved):

rites/roman1969/src/calendars/general-roman/index.ts
```diff
@@ -1458,7 +1458,7 @@ export class GeneralRoman extends CalendarDef {
 
     // src: mr_la_2008_ed3
```
TobiTenno:
```suggestion
    // src: Calendarium Romanum 1969 https://archive.org/details/CalendariumRomanum1969/page/n29/mode/2up
```

TobiTenno (changes requested):

rites/roman1969/src/calendars/general-roman/index.ts
TobiTenno: @kvnang could you include the new source as a citation in the code comments?

TobiTenno:
@kvnang i think you just need to update the snapshots that are looking for that feast now

kvnang:
Thank you @TobiTenno I have updated the commit title + your amendment.

To your point, I also cannot 100% attribute the fact `FESTUM` with all caps means `Festa Domini`. It doesn't seem to be consistent, and it's not listed here later on in the document:
https://archive.org/details/CalendariumRomanum1969/page/n65/mode/2up

But looking up elsewhere it seems to be commonly understood that this is a Feast of Our Lord, but I'm still not sure what document officially indicated this.

tukusejssirs:
@kvnang, thanks for the PR! :pray:

> Change the rank of feast of the Lateran basilica to Our Lord's feast.

I didn’t know that the Lateran Basilica is dedicated to the Most Holy Saviour (and St John the Baptist and St John the Evangelist; cf [Wikipedia](https://en.wikipedia.org/wiki/Archbasilica_of_Saint_John_Lateran)). As it is the case, it makes sense to change the precedence to Our Lord’s Feast.

> I think we need to update the docs to include the calendarium romanum entry, as we didn't have that on hand as of yet, and in that case, I agree.

I had access to the _Calendarium_ for several years (I have a copy of it from archive.org in my [gDrive](https://drive.google.com/drive/folders/1c21ceESdtfbhzAY-yRsR7F735Nd-Sus8?usp=drive_link)), but we don’t use it yet. We definitely need to use and reference more official sources (starting with the _Calendarium_, _GIRM_, _GILH_, _GNLY_, various translations of the Roman Missal, _Martyrologium Romanum_, and various decress.

> I think the final word would have to come from somewhere where we can find specifically that the title in all caps as a FESTUM indicates a feast of the Lord

Well, it is definitely a feast that has a higher precedence than Sunday (cf [the calendar in Slovak which is based on official Slovak Ordo](lc.kbs.sk/?den=20251109)). I have no proof/source where the celebration is explicitely considered as a feast _of the Lord_, however, the dedication (the principal patron of the church) is Our Lord, therefore, it seems to be correct to consider the feast as a feast of Our Lord.

TobiTenno:
Agreed, I was correlating the title in small capital letters as the highest a feast could be before it moved to a solemnity. (Where they title would be in full large capitals)

rites/roman1969/src/calendars/general-roman/index.ts
tukusejssirs: Agreed with this case, as we should use the earliest source we can find which proves something in a particular GRC revision. In this case the _Calendarium_ was released before _Missale Romanum_.

TobiTenno: the citation is updated, the snapshots just need to be updated so the tests pass now

tukusejssirs:
> I also cannot 100% attribute the fact `FESTUM` with all caps means `Festa Domini`.

I don’t think that _festum_ means _festa Domini_, as there are other feast (_festum_) which are not feast of the Lord. As I understand it, feast of the Lord are those feast which are directly related to Our Lord (e.g. _Corpus Christi_, Transfiguration, and probably also the Dedication of the Lateran Basilica).

TobiTenno:
> I don’t think that festum means festa Domini

Agreed, I was referring to the styling of the title of the feast itself, where the solemnities are large and all capital letters, whereas this feast is small and all capital letters

TobiTenno (approved):

tukusejssirs:
> Agreed, I was referring to the styling of the title of the feast itself, where the solemnities are large and all capital letters, whereas this feast is small and all capital letters

I don’t think the styling of celebration rank matters, ever. It matters if it is used to style the celebration name and no rank is explicitly typed out.

---

What do you think _II cl._ means in `feastum (II cl.)` in [_Calendarium_, p175])(https://archive.org/details/CalendariumRomanum1969/page/n173/mode/2up)? I slightly remember there were rank classes in GRC revisions before the Second Vatican Council, but I don’t remember them to be used after the council. :thinking:

image

TobiTenno:
I'll probably move to the discussion, but those classes correspond to the same I, II, III class system as before Vatican II, defined on pages 19 & 20: https://archive.org/details/CalendariumRomanum1969/page/n19/mode/2up

pvanthony:
Thank you for doing this.
It is working at https://romcal.js.org/#/calendar/en/general-roman/2025/11

PR #455, opened 2023-08-31
fix: make sure all l10n keys are defined in `en.ts`

tukusejssirs:
Changes:

- fix(`la`): rename `mary_of_jesus_in_the_blessed_sacrament_venegas_virgin` to `mary_of_jesus_in_the_blessed_sacrament_venegas_de_la_torre_virgin`;
- test(l10n): add a unit test to check if all `names` l10n keys defined in non-English locales, are also defined in `en.ts`.

PR #992, opened 2025-11-30
fix(l10n): missing locales

TobiTenno:

- add strings missing from en.ts
- tests for names existing in en.ts checking all other locales
- export LOCALES (already present in the bundle, just wasn't exported)
- closes #455  

## Concept

- Localization

## Example code

No new usage, just verification by way of tests
```ts
const English = Romcal.LOCALES.En;
const NonEnglish = Object.values(Romcal.LOCALES).filter((locale) => locale.id !== 'en');
const missing: Array<{ locale: string; key: string }> = [];
NonEnglish.forEach((locale) => {
  Object.keys(locale.names!).forEach((key) => {
    if (!English.names!.hasOwnProperty(key)) {
      missing.push({ locale: locale.id, key });
    }
  });
});
```

### DIFF
--- a/docs/calendar-definitions.md
+++ b/docs/calendar-definitions.md
@@ -56,6 +56,8 @@ In the `england` calendar, we will only define a new entry with its `key` and th
 
 It's generally a good practice to extend an existing liturgical day, with its specifications for a particular calendar, instead of redefining multiple time the same liturgical day in different calendars.
 
+For any extension or redefinition, there must be a corresponding src reference in the calendar to provide context for the definition. See [localization documentation](./localization.md) for more details on how these references should be structured.
+
 ### Overriding a liturgical day by its key
 
 In most countries, All Saints and All Souls are always celebrated on the 1st and 2nd of November respectively. However, in England and Wales, if All Saints (1 November) falls on a Saturday, it is transferred to the Sunday and All Souls is transferred to Monday 3rd November.
--- a/docs/calendar-definitions.md
+++ b/docs/calendar-definitions.md
@@ -56,7 +56,7 @@ In the `england` calendar, we will only define a new entry with its `key` and th
 
 It's generally a good practice to extend an existing liturgical day, with its specifications for a particular calendar, instead of redefining multiple time the same liturgical day in different calendars.
 
-For any extension or redefinition, there must be a corresponding src reference in the calendar to provide context for the definition. See [localization documentation](./localization.md) for more details on how these references should be structured.
+For any extension or redefinition, there must be a corresponding source reference in the calendar to provide context for the definition. See [localization documentation](./localization.md) for more details on how these references should be structured.
 
 ### Overriding a liturgical day by its key
 
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -56,7 +56,7 @@ Each entry in the locale file corresponds to a specific part of the liturgical c
 - `names`: Localized names for specific liturgical days, such as feasts of saints and titles of Our Lord and Our Blessed Mother.
   - Each entry for any name that is added to a specific locale other than `en` must also exist in the `en` locale file, as this is the main source of truth.
   - If a name is not found in the specified locale, romcal will look for it in the fallback locales, ending with `en`.
-  - There are also tests to ensure that all names in other locales exist in the `en` locale file. (pending #992)
+  - There are also tests to ensure that all names in other locales exist in the `en` locale file.
   - Each entry should have a corresponding src comment with a link to the source of the translation, if applicable.
     - If a reference to a missal is used instead, this is the format we've decided to use:
       - `// src: $sourceName`
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -57,7 +57,7 @@ Each entry in the locale file corresponds to a specific part of the liturgical c
   - Each entry for any name that is added to a specific locale other than `en` must also exist in the `en` locale file, as this is the main source of truth.
   - If a name is not found in the specified locale, romcal will look for it in the fallback locales, ending with `en`.
   - There are also tests to ensure that all names in other locales exist in the `en` locale file.
-  - Each entry should have a corresponding src comment with a link to the source of the translation, if applicable.
+  - Each entry should have a corresponding source comment with a link to the source of the translation, if applicable.
     - If a reference to a missal is used instead, this is the format we've decided to use:
       - `// src: $sourceName`
         - when the key value can be found exactly in the referenced document
@@ -71,4 +71,4 @@ Each entry in the locale file corresponds to a specific part of the liturgical c
         - `angers` = optional suffix when variations exist (examples below)
           - English Missals are published e.g. in the UK, the US, Australia
           - French Missals are published per diocese in France, but maybe in Canada too
-      - so this would be the full reference: `// src: mr_fr_2022_ed3_angers`
+      - the parts above are joined with an underscore (`_`): `// src: mr_fr_2022_ed3_angers`
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -1,12 +1,10 @@
-**:warning: The documentation above is written for romcal v2, and has not been updated yet to romcal v3.**
-
 # Localization
 
 ## Localizing liturgical day names
 
 Liturgical day names in romcal can be localized to any language that is already supported by [DayJS i18n](https://day.js.org/docs/en/i18n/i18n).
 
-Locales are stored as `.ts` files in the `src/locales` directory.
+Locales are stored as `.ts` files in the `rites/roman1969/src/locales` directory.
 
 If the given locale contains a region (the second group of letters after a hyphen, e.g. for `fr-ca`, the `ca` represents the region), a base language will be set as a locale fallback (`fr` in our example).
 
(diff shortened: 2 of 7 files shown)


docs/localization.md
```diff
@@ -22,22 +20,53 @@ The structure of the locale file is typically like so:
...
+      - `// src: $sourceName`
+        - when the key value can be found exactly in the referenced document
+      - `// based on: $sourceName`
+        - when the key value is adapted from the referenced document
+      - For a given missal or document, the following short form is used: (values are examples)
+        - `mr` = Missale Romanum;
+        - `fr` = is the locale of the Missal;
+        - `2022` = publication year;
+        - `ed3` = editio typica tertia;
+        - `angers` = optional suffix when variations exist (e.g. English Missals are published e.g. in the UK, the US, Australia; French Missals are published per diocese in France, but maybe in Canada too).
```
emagnier: Just one clarification: the French-language missal is published for several French-speaking countries (including Canada, France, Belgium, Luxembourg, Switzerland, and North Africa). But dioceses also usually publish their own proper missal, which includes the feasts and memorials specific to that diocese. So this is not something unique to dioceses in France, but something I believe all dioceses do.

Other than that LGTM, thank you 👍

emagnier (approved):

docs/localization.md
TobiTenno: sure, that was an example, not so much a rule that only France does this.

TobiTenno: exemplification clarified

TobiTenno:
(disabled automerge so @tukusejssirs has time to review)

docs/localization.md
tukusejssirs: > So this is not something unique to dioceses in France, but something I believe all dioceses do.

I know only about French dioceses which publish diocesan Missals. Maybe there are dioceses in other countries which do so too, however, in Slovakia, the Slovak translation of the Roman Missal (as well the national ordos) includes all the celebrations inscribed to any of the eight Slovak dioceses (probably to save cost, but also there are, I think, up to 20 celebrations inscribed in all of the Slovak dioceses).

TobiTenno: i don't think this changes the validity of the example.

docs/calendar-definitions.md
```diff
@@ -56,6 +56,8 @@ In the `england` calendar, we will only define a new entry with its `key` and th
 
 It's generally a good practice to extend an existing liturgical day, with its specifications for a particular calendar, instead of redefining multiple time the same liturgical day in different calendars.
 
+For any extension or redefinition, there must be a corresponding src reference in the calendar to provide context for the definition. See [localization documentation](./localization.md) for more details on how these references should be structured.
```
tukusejssirs:
```suggestion
For any extension or redefinition, there must be a corresponding source reference in the calendar to provide context for the definition. See [localization documentation](./localization.md) for more details on how these references should be structured.
```

docs/localization.md
```diff
@@ -22,22 +20,55 @@ The structure of the locale file is typically like so:
...
+      - Other fields may be present depending on needs of the particular season.
+- `periods`: Localized names for liturgical periods.
+- `ranks`: Localized names for liturgical ranks.
+- `weekdays`: Localized names for the days of the week.
+- `months`: Localized names for the months of the year.
+- `colors`: Localized names for liturgical colors.
+- `names`: Localized names for specific liturgical days, such as feasts of saints and titles of Our Lord and Our Blessed Mother.
+  - Each entry for any name that is added to a specific locale other than `en` must also exist in the `en` locale file, as this is the main source of truth.
+  - If a name is not found in the specified locale, romcal will look for it in the fallback locales, ending with `en`.
+  - There are also tests to ensure that all names in other locales exist in the `en` locale file. (pending #992)
```
tukusejssirs: The note is unnecessary, as the PR is this one, therefore, when this PR is merged, the remark will be outdated.

```suggestion
  - There are also tests to ensure that all names in other locales exist in the `en` locale file.
```

docs/localization.md
```diff
@@ -22,22 +20,55 @@ The structure of the locale file is typically like so:
...
+  - Each entry should have a corresponding src comment with a link to the source of the translation, if applicable.
```
tukusejssirs:
```suggestion
  - Each entry should have a corresponding source comment with a link to the source of the translation, if applicable.
```

docs/localization.md
```diff
@@ -22,22 +20,55 @@ The structure of the locale file is typically like so:
...
+    - If a reference to a missal is used instead, this is the format we've decided to use:
+      - `// src: $sourceName`
+        - when the key value can be found exactly in the referenced document
+      - `// based on: $sourceName`
+        - when the key value is adapted from the referenced document
+      - For a given missal or document, the following short form is used: (values are examples)
```
tukusejssirs: Right now, we do this for missals only.

```suggestion
      - For missals, the following short form is used: (values are examples)
```

docs/localization.md
```diff
@@ -22,22 +20,55 @@ The structure of the locale file is typically like so:
...
+        - `mr` = Missale Romanum;
+        - `fr` = is the locale of the Missal;
+        - `2022` = publication year;
+        - `ed3` = editio typica tertia;
+        - `angers` = optional suffix when variations exist (examples below)
+          - English Missals are published e.g. in the UK, the US, Australia
+          - French Missals are published per diocese in France, but maybe in Canada too
+      - so this would be the full reference: `// src: mr_fr_2022_ed3_angers`
```
tukusejssirs:
```suggestion
      - the parts above are joined with an underscore (`_`): `// src: mr_fr_2022_ed3_angers`
```

rites/roman1969/src/locales/en.ts
```diff
@@ -806,8 +807,8 @@ export const locale: Locale = {
     michael_kozal_bishop: 'Blessed Michael Kozal, Bishop and Martyr',
     miguel_agustin_pro_priest: 'Blessed Miguel Agustín Pro, Priest and Martyr',
     miguel_febres_cordero_religious: 'Saint Miguel Febres Cordero, Religious',
-    modestus_andlauer_and_andrew_bauer_martyrs: 'Saints Modestus Andlauer and Andrew Bauer, Martyrs',
-    modestus_andlauer_martyr: 'Saint Modestus Andlauer, Martyr',
+    modestus_andlauer_and_andrew_bauer_martyrs: 'Saints Modeste Andlauer and Andrew Bauer, Martyrs',
+    modestus_andlauer_martyr: 'Saint Modeste Andlauer, Martyr', // src: https://en.wikipedia.org/wiki/Martyr_Saints_of_China
```
tukusejssirs, line 811: I am not sure what form of his name we should use in English: `Modeste` (French) or `Modestus` (Latin)? Wikipedia usually prefers original versions, while in celebrations we have the names usually localised. I am not sure what is the English form of the the name or whether this particular saint is well known in English-speaking countries that the original (French) form is used. 🤷‍♂️

We may keep `Modeste` if you want.

tukusejssirs (changes requested):
Just some nitpicks, nothing serious. All in all, I like the changes you’ve made

docs/calendar-definitions.md
TobiTenno: I specifically used `src`  to indicate the precise annotation

tukusejssirs: I don’t think that is necessary, as on L62 a L64 of `docs/localization.md` you explicitly mention the format. Moreover, from grammatical PoV it is incorrect.

rites/roman1969/src/locales/en.ts
TobiTenno: I specified `Modeste` because this is the english locale and Wikipedia uses that spelling, since that's what's in context.

TobiTenno:
@tukusejssirs i've resolved all of your requests

rites/roman1969/src/locales/en.ts
tukusejssirs: Right, but then we should rename the ID to `modeste_andlauer_martyr`.

tukusejssirs: Right, but then we should rename the ID to `modeste_andlauer_martyr`.

tukusejssirs:
Thanms, @TobiTenno. There is one last thing we should fix, then you can merge it. :pray:

tukusejssirs:
Thanms, @TobiTenno. There is one last thing we should fix, then you can merge it. :pray:

TobiTenno:
@tukusejssirs what's the one last thing? you posted it twice and didn't say what needed to be fixed

rites/roman1969/src/locales/en.ts
TobiTenno: i don't think we should change the ids after we've created them, as the ids could be used outside our library and changing them would be a breaking change every time we change them.

TobiTenno: I only changed the name because it didn't have a src, and this was the only english name i could find a source for

tukusejssirs: > i don't think we should change the ids after we've created them, as the ids could be used outside our library and changing them would be a breaking change every time we change them.

Agreed, however, we can change them before major stable versions are released.

In the end, I’d rather decouple the IDs from celebration metadata (e.g. by using UUIDs), however, for that we need a way to automate it so that no one would need to remember the IDs. From my PoE currently, this causes too much troubles (what should be part of an ID? what if the metadata change? and the ID generation ia quite complex). However, I get that we use human-hreadable IDs to simplify our laborous work until we can automate it a bit.

In the end, merge as is if you want, I don’t want fight over this.

TobiTenno: I can't merge without overriding when you have requested changes...

tukusejssirs (approved):

PR #993, opened 2025-12-04
fix(brazil): update Brazilian calendar definitions and add tests

tukusejssirs:
## Summary

Calendar changes:

- Fix `peter_of_alcantara_priest` rank to `OptionalMemorial_12`
- Move `faustina_kowalska_virgin` after `benedict_the_moor_religious`
- Remove `john_de_brebeuf_isaac_jogues_priests_and_companions_martyrs`
  (celebrated alongside `peter_of_alcantara_priest` and `paul_of_the_cross_priest`
  as optional memorials)
- Remove `albertina_berkenbrock_virgin` (not found in Brazilian Ordos
  or cnbb.org.br)
- Add date exceptions for `assumption_of_the_blessed_virgin_mary`
  (transfer to 3rd Sunday of August)
- Add date exceptions for `all_saints` (transfer to 1st Sunday of November)
- Add source references to all celebrations

Locale changes (`pt-br.ts`):

- `andrew_de_soveral_and_ambrose_francis_ferro_priests`
- `anthony_of_saint_anne_galvao_priest`
- `assumption_of_the_blessed_virgin_mary`
- `ephrem_the_syrian_deacon`
- `ignatius_de_azevedo_priest_and_companions_martyrs`
- `john_de_brebeuf_isaac_jogues_priests_and_companions_martyrs`
- `our_lady_of_aparecida_patroness_of_brazil`
- `our_lady_of_mount_carmel`
- `paulina_of_the_agonizing_heart_of_jesus_visintainer_virgin`
- `pontian_i_pope_and_hippolytus_of_rome_priest`
- `roch_gonzalez_alphonsus_rodriguez_and_john_del_castillo_priests`

Test changes:

- Add tests for all Brazilian celebrations in chronological order
- Fix `peter_of_alcantara_priest` test to expect `OptionalMemorial`
- Fix typo in `benedict_the_moor_religious` test name

PR #1001, opened 2025-12-06
refactor(util): improve data checks and update teresa_of_calcutta_virgin translations

tukusejssirs:
## Summary

This PR contains two changes:

### 1. Change unused martyrology items check to informational

- Update `data-checks.ts` to change the "useless martyrology items" check from an error to an informational message
- Valid martyrology entries should not be restricted even if not yet used by any calendar - they may be reserved for future use
- Renumber the check comments to reflect the new order

### 2. Update `teresa_of_calcutta_virgin` translations

Update translations and add source references for `teresa_of_calcutta_virgin`:

- `de`: change `Kalkutta` to `Kolkata`
- `en`: add source reference
- `es`: add source reference  
- `fr`: add translation with source
- `it`: add translation with source
- `pl`: add translation with source

Sources:
- https://press.vatican.va/content/salastampa/it/bollettino/pubblico/2025/02/11/0125/00250.html
- https://liturgia.wiara.pl/doc/9143462.Swieta-Teresa-z-Kalkuty-w-Powszechnym-Kalendarzu-Rzymskim

### DIFF
--- a/rites/roman1969/__tests__/calendar-data.test.ts
+++ b/rites/roman1969/__tests__/calendar-data.test.ts
@@ -1,0 +1,311 @@
+import { calendarDefinitions } from '../src/calendars';
+import { GeneralRoman } from '../src/calendars/general-roman';
+import { Martyrology } from '../src/catalog/martyrology';
+import { locales } from '../src/locales';
+import { RomcalConfig } from '../src/models/config';
+import { CalendarDefInstance, Inputs } from '../src/types/calendar-def';
+import { DateDef } from '../src/types/liturgical-day';
+
+import { findMissingInArray } from './util/findMissingInArray';
+
+const MONTH_NAMES = [
+  'January',
+  'February',
+  'March',
+  'April',
+  'May',
+  'June',
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December',
+];
+
+/**
+ * Map of dateFn names to their fixed date equivalents (month, date).
+ * These functions always return the same date regardless of year.
+ */
+const FIXED_DATE_FN_MAP: Record<string, { month: number; date: number }> = {
+  peterAndPaulApostles: { month: 6, date: 29 },
+  assumption: { month: 8, date: 15 },
+  allSaints: { month: 11, date: 1 },
+  christmas: { month: 12, date: 25 },
+  maryMotherOfGod: { month: 1, date: 1 },
+  epiphany: { month: 1, date: 6 },
+  immaculateConceptionOfMary: { month: 12, date: 8 },
+};
+
+/**
+ * Extract all celebration locale keys (names keys) used across all calendar definitions.
+ * This includes:
+ * - The celebration ID itself if it's in the names object
+ * - The customLocaleId if defined
+ */
+const extractAllCelebrationLocaleKeys = (): Set<string> => {
+  const allKeys = new Set<string>();
+  const devLocale = { id: 'dev' };
+
+  for (const CalendarDefClass of Object.values(calendarDefinitions)) {
+    const isGRC = CalendarDefClass.name === GeneralRoman.name;
+    const config = isGRC
+      ? new RomcalConfig({ scope: 'liturgical' }, Martyrology.catalog, devLocale)
+      : new RomcalConfig({ scope: 'liturgical' }, Martyrology.catalog, devLocale, CalendarDefClass);
+
+    config.calendarsDef.forEach((cal) => cal.buildAllDefinitions());
+
+    for (const def of Object.values(config.liturgicalDayDef)) {
+      // The i18nDef typically starts with 'names:' for celebration names
+      if (def.i18nDef && def.i18nDef[0]?.startsWith('names:')) {
+        const nameKey = def.i18nDef[0].replace('names:', '');
+        allKeys.add(nameKey);
+      }
+
+      // Check for customLocaleId in inputs
+      for (const input of def.input) {
+        if (input.customLocaleId) {
+          allKeys.add(input.customLocaleId);
+        }
+      }
+    }
+  }
+
+  return allKeys;
+};
+
+/**
+ * Get the inputs from a calendar class.
+ */
+const getCalendarInputs = (CalendarClass: CalendarDefInstance): Inputs => {
+  const instance = new (CalendarClass as unknown as new (cfg: unknown) => { inputs?: Inputs })({
+    calendarsDef: [],
+    liturgicalDayDef: {},
+  });
+  return instance.inputs || {};
+};
+
+/**
+ * Get the parent calendars from a calendar class.
+ */
+const getParentCalendars = (CalendarClass: CalendarDefInstance): CalendarDefInstance[] => {
+  const proto = (
+    CalendarClass as unknown as {
+      prototype?: { constructor?: { prototype?: { ParentCalendars?: CalendarDefInstance[] } } };
+    }
+  ).prototype;
+  return proto?.constructor?.prototype?.ParentCalendars || [];
+};
+
+/**
+ * Build the full inheritance chain for a calendar (from most specific to GeneralRoman).
+ * Returns an array of calendar classes, starting with the current calendar.
+ */
+const buildInheritanceChain = (CalendarClass: CalendarDefInstance): CalendarDefInstance[] => {
+  const chain: CalendarDefInstance[] = [CalendarClass];
+
+  if (CalendarClass === GeneralRoman) {
+    return chain;
+  }
+
+  const parents = getParentCalendars(CalendarClass);
+
+  if (parents.length === 0) {
+    // No explicit parents means implicit inheritance from GeneralRoman
+    chain.push(GeneralRoman);
+  } else {
+    // Traverse parent chain
+    for (const parent of parents) {
+      chain.push(...buildInheritanceChain(parent));
+    }
+    // Ensure GeneralRoman is at the end if not already included
+    if (!chain.includes(GeneralRoman)) {
+      chain.push(GeneralRoman);
+    }
+  }
+
+  return chain;
+};
+
+/**
+ * Find the dateDef for a celebration ID by traversing the inheritance chain.
+ * Continues searching up the chain even if an entry exists without dateDef.
+ */
+const findDateDefInChain = (celebrationId: string, chain: CalendarDefInstance[]): DateDef | undefined => {
+  for (const CalendarClass of chain) {
+    const inputs = getCalendarInputs(CalendarClass);
+    const input = inputs[celebrationId];
+
+    if (input) {
+      // Handle both single input and array of inputs
+      const inputObj = Array.isArray(input) ? input[0] : input;
+      if (inputObj.dateDef) {
+        return inputObj.dateDef as DateDef;
+      }
+      // Continue searching - entry exists but has no dateDef (only dateExceptions, etc.)
+    }
+  }
+  return undefined;
+};
+
+/**
+ * Format a DateDef into a human-readable string.
+ */
+const formatDateDef = (dateDef: DateDef | undefined): string => {
+  if (!dateDef) return 'unknown';
+
+  // Fixed date: month and date specified
+  if ('month' in dateDef && 'date' in dateDef) {
+    return `${MONTH_NAMES[dateDef.month - 1]} ${dateDef.date}`;
+  }
+
+  // Fixed date with nthWeekInMonth
+  if ('month' in dateDef && 'nthWeekInMonth' in dateDef) {
+    const ordinal = ['1st', '2nd', '3rd', '4th', '5th'][dateDef.nthWeekInMonth! - 1] || `${dateDef.nthWeekInMonth}th`;
+    return `${ordinal} week of ${MONTH_NAMES[dateDef.month - 1]}`;
+  }
+
+  // Fixed date with lastDayOfWeekInMonth
+  if ('month' in dateDef && 'lastDayOfWeekInMonth' in dateDef) {
+    return `last week of ${MONTH_NAMES[dateDef.month - 1]}`;
+  }
+
+  // Movable feast (dateFn-based)
+  if ('dateFn' in dateDef) {
+    // Check if this dateFn maps to a fixed date
+    const fixedDate = FIXED_DATE_FN_MAP[dateDef.dateFn];
+    if (fixedDate) {
+      return `${MONTH_NAMES[fixedDate.month - 1]} ${fixedDate.date}`;
+    }
+
+    let result = dateDef.dateFn;
+    if ('addDay' in dateDef && dateDef.addDay) {
+      result += ` +${dateDef.addDay} days`;
+    }
+    if ('subtractDay' in dateDef && dateDef.subtractDay) {
+      result += ` -${dateDef.subtractDay} days`;
+    }
+    return result;
+  }
+
+  return 'unknown';
+};
+
+/**
+ * Get a sortable date value from a DateDef.
+ * Fixed dates return month * 100 + date (e.g., Jan 15 = 115, Dec 25 = 1225)
+ * Movable dates (dateFn-based) return a large number so they sort after fixed dates.
+ */
+const getDateSortValue = (dateDef: DateDef | undefined): number => {
+  if (!dateDef) return 99999;
+
+  // Fixed date: month and date specified
+  if ('month' in dateDef && 'date' in dateDef) {
+    return dateDef.month * 100 + dateDef.date;
+  }
+
+  // Fixed date with nthWeekInMonth (e.g., first Monday of September)
+  // Use day 1 of that week range as the sort value
+  if ('month' in dateDef && 'nthWeekInMonth' in dateDef) {
+    // nthWeekInMonth=1 means days 1-7, so use day 1 + (nthWeek-1)*7
+    const approximateDay = 1 + ((dateDef.nthWeekInMonth ?? 1) - 1) * 7;
+    return dateDef.month * 100 + approximateDay;
+  }
+
+  // Fixed date with lastDayOfWeekInMonth
+  if ('month' in dateDef && 'lastDayOfWeekInMonth' in dateDef) {
+    // Approximate: last week of month, so around day 22-28
+    return dateDef.month * 100 + 22;
+  }
+
+  // Movable feast (dateFn-based)
+  if ('dateFn' in dateDef) {
+    // Check if this dateFn maps to a fixed date
+    const fixedDate = FIXED_DATE_FN_MAP[dateDef.dateFn];
+    if (fixedDate) {
+      return fixedDate.month * 100 + fixedDate.date;
+    }
+
+    // True movable feasts (Easter-dependent) sort after all fixed dates
+    const addDay = 'addDay' in dateDef ? (dateDef.addDay ?? 0) : 0;
+    const subtractDay = 'subtractDay' in dateDef ? (dateDef.subtractDay ?? 0) : 0;
+    return 90000 + addDay - subtractDay;
+  }
+
+  return 99999;
+};
+
+type SortingIssue = {
+  key: string;
+  keyDate: string;
+  prevKey: string;
+  prevKeyDate: string;
+};
+
+/**
+ * Check if a calendar's inputs are sorted chronologically.
+ * Uses inheritance chain to resolve dates.
+ */
+const checkCalendarSorting = (CalendarClass: CalendarDefInstance): SortingIssue[] => {
+  const inputs = getCalendarInputs(CalendarClass);
+  const inputKeys = Object.keys(inputs);
+
+  if (inputKeys.length === 0) {
+    return [];
+  }
+
+  const chain = buildInheritanceChain(CalendarClass);
+  const issues: SortingIssue[] = [];
+
+  let prevSortValue = -Infinity;
+  let prevKey = '';
+  let prevDateDef: DateDef | undefined;
+
+  for (const key of inputKeys) {
+    const dateDef = findDateDefInChain(key, chain);
+    const sortValue = getDateSortValue(dateDef);
+
+    // Allow same date (multiple celebrations on same day)
+    if (sortValue < prevSortValue) {
+      issues.push({
+        key,
+        keyDate: formatDateDef(dateDef),
+        prevKey,
+        prevKeyDate: formatDateDef(prevDateDef),
+      });
+    }
+
+    prevSortValue = sortValue;
+    prevKey = key;
+    prevDateDef = dateDef;
+  }
+
+  return issues;
+};
+
+describe('Calendar Data Validation', () => {
+  describe('All celebration keys used in calendars must exist in en.ts names', () => {
+    const allCelebrationKeys = extractAllCelebrationLocaleKeys();
+    const enNamesKeys = Object.keys(locales.En.names ?? {});
+    const missingKeys = findMissingInArray(allCelebrationKeys, enNamesKeys);
+
+    it('should have all celebration names in en.ts', () => {
+      expect(missingKeys).toEqual([]);
+    });
+  });
+
+  describe('Calendar celebrations should be sorted chronologically', () => {
+    for (const [calendarName, CalendarDefClass] of Object.entries(calendarDefinitions)) {
+      const inputs = getCalendarInputs(CalendarDefClass);
+
+      // Skip calendars with no inputs
+      if (Object.keys(inputs).length === 0) {
+        continue;
+      }
+
+      it(`${calendarName} celebrations should be sorted chronologically`, () => {
+        expect(checkCalendarSorting(CalendarDefClass)).toEqual([]);
+      });
+    }
+  });
+});
--- a/rites/roman1969/__tests__/locale-data.test.ts
+++ b/rites/roman1969/__tests__/locale-data.test.ts
@@ -1,0 +1,392 @@
+import { merge } from 'ts-deepmerge';
+
+import { calendarDefinitions } from '../src/calendars';
+import { GeneralRoman } from '../src/calendars/general-roman';
+import { Martyrology } from '../src/catalog/martyrology';
+import { SEASONS } from '../src/constants/seasons';
+import { locales } from '../src/locales';
+import { RomcalConfig } from '../src/models/config';
+import { ProperOfTime } from '../src/proper-of-time/proper-of-time';
+import { Locale } from '../src/types/locale';
+import { toPackageName } from '../src/utils/string';
+
+import { getNestedPropertyNames } from './util/getNestedPropertyNames';
+import { isObjectPropsSortedAlphabetically } from './util/isObjectPropsSortedAlphabetically';
+
+/**
+ * Extract all season keys used by Proper of Time dynamically.
+ * Returns keys like 'advent.sunday', 'lent.weekday', etc.
+ */
+const extractProperOfTimeSeasonKeys = (): string[] => {
+  const seasonKeys = new Set<string>();
+  const devLocale = { id: 'dev' };
+
+  const config = new RomcalConfig({ scope: 'liturgical' }, Martyrology.catalog, devLocale);
+  const properOfTime = new ProperOfTime(config);
+  properOfTime.buildAllDefinitions();
+
+  for (const def of Object.values(config.liturgicalDayDef)) {
+    if (def.i18nDef && def.i18nDef[0]?.startsWith('seasons:')) {
+      // Extract path after 'seasons:', e.g., 'advent.sunday'
+      const path = def.i18nDef[0].replace('seasons:', '');
+      seasonKeys.add(path);
+    }
+  }
+
+  // Also add season names (*.season)
+  SEASONS.forEach((s) => seasonKeys.add(`${s.toLowerCase()}.season`));
+
+  return Array.from(seasonKeys).sort();
+};
+
+/**
+ * Extract ordinal base numbers from en.ts (the reference locale).
+ * Returns ['1', '2', ...] without gender suffixes.
+ */
+const extractOrdinalKeys = (): string[] => {
+  const enOrdinals = locales.En.ordinals;
+  if (!enOrdinals) return [];
+
+  // Get unique base numbers (strip _masculine, _feminine suffixes)
+  const baseNumbers = new Set<string>();
+  for (const key of Object.keys(enOrdinals)) {
+    const baseNum = key.split('_')[0];
+    baseNumbers.add(baseNum);
+  }
+
+  return Array.from(baseNumbers).sort((a, b) => Number(a) - Number(b));
+};
+
+/**
+ * Extract keys from en.ts for a specific namespace.
+ */
+const extractKeysFromEnLocale = (namespace: keyof Locale): string[] => {
+  const obj = locales.En[namespace];
+  if (!obj || typeof obj !== 'object') return [];
+  return Object.keys(obj);
+};
+
+/**
+ * Meta locale keys that must exist in ALL locales.
+ * Derived dynamically from en.ts (reference locale) and Proper of Time.
+ */
+const REQUIRED_META_KEYS = {
+  colors: extractKeysFromEnLocale('colors'),
+  ranks: extractKeysFromEnLocale('ranks'),
+  cycles: extractKeysFromEnLocale('cycles'),
+  weekdays: extractKeysFromEnLocale('weekdays'),
+  months: extractKeysFromEnLocale('months'),
+  ordinals: extractOrdinalKeys(),
+  seasons: extractProperOfTimeSeasonKeys(),
+  periods: extractKeysFromEnLocale('periods'),
+};
+
+/**
+ * Map of locale IDs to their associated calendar names.
+ * A locale must have all keys required by these calendars.
+ */
+const localeToCalendarsMap: Record<string, string[]> = {
+  En: ['GeneralRoman', 'England', 'Scotland', 'Wales', 'Ireland', 'UnitedStates', 'Canada', 'Australia', 'NewZealand'],
+  EnGb: ['England', 'Scotland', 'Wales'],
+  EnIe: ['Ireland'],
+  PtBr: ['Brazil'],
+  Fr: [
+    'France',
+    'France_Angers',
+    'France_Coutances',
+    'France_Lyon',
+    'France_Paris',
+    'France_SaintDenis',
+    'France_Strasbourg',
+    'France_Toulouse',
+  ],
+  De: ['Germany', 'Austria', 'Switzerland'],
+  Es: [
+    'Spain',
+    'Mexico',
+    'Argentina',
+    'Chile',
+    'Bolivia',
+    'CostaRica',
+    'Guatemala',
+    'Panama',
+    'Paraguay',
+    'Peru',
+    'Uruguay',
+    'Venezuela',
+    'PuertoRico',
+  ],
+  It: ['Italy'],
+  Pl: ['Poland'],
+  Cs: ['CzechRepublic'],
+  Sk: ['Slovakia'],
+  La: ['GeneralRoman'],
+  Ta: ['India'],
+};
+
+/**
+ * Meta i18n keys required by all locales (derived from REQUIRED_META_KEYS).
+ * Used for building calendar locale key sets.
+ */
+const metaI18nKeys: string[] = [
+  ...REQUIRED_META_KEYS.colors.map((c) => `colors:${c}`),
+  ...REQUIRED_META_KEYS.ranks.map((r) => `ranks:${r}`),
+  ...REQUIRED_META_KEYS.cycles.map((c) => `cycles:${c}`),
+  ...REQUIRED_META_KEYS.weekdays.map((w) => `weekdays:${w}`),
+  ...REQUIRED_META_KEYS.months.map((m) => `months:${m}`),
+  ...REQUIRED_META_KEYS.seasons.map((s) => `seasons:${s}`),
+  ...REQUIRED_META_KEYS.periods.map((p) => `periods:${p}`),
+];
+
+/**
+ * Extract all locale keys (i18n keys) used across specified calendar definitions.
+ * Also includes customLocaleId values when celebrations define them.
+ */
+const extractLocaleKeysForCalendars = (calendarNames: string[]): Set<string> => {
+  const allLocaleKeys = new Set<string>(metaI18nKeys);
+  const devLocale = { id: 'dev' };
+
+  for (const calendarName of calendarNames) {
+    const CalendarDefClass = calendarDefinitions[calendarName];
+    if (!CalendarDefClass) continue;
+
+    const isGRC = CalendarDefClass.name === GeneralRoman.name;
+    const config = isGRC
+      ? new RomcalConfig({ scope: 'liturgical' }, Martyrology.catalog, devLocale)
+      : new RomcalConfig({ scope: 'liturgical' }, Martyrology.catalog, devLocale, CalendarDefClass);
+
+    config.calendarsDef.forEach((cal) => cal.buildAllDefinitions());
+
+    for (const def of Object.values(config.liturgicalDayDef)) {
+      // Add the i18nDef key
+      if (def.i18nDef && def.i18nDef[0]) {
+        allLocaleKeys.add(def.i18nDef[0]);
+      }
+
+      // Also check for customLocaleId in input definitions
+      for (const input of def.input) {
+        if (input.customLocaleId) {
+          allLocaleKeys.add(`names:${input.customLocaleId}`);
+        }
+      }
+    }
+  }
+
+  return allLocaleKeys;
+};
+
+/**
+ * Extract locale keys used across ALL calendar definitions (without martyrology keys).
+ * Used for checking required keys in en.ts.
+ */
+const extractAllLocaleKeysFromCalendars = (): Set<string> => {
+  const allCalendarNames = Object.keys(calendarDefinitions);
+  return extractLocaleKeysForCalendars(allCalendarNames);
+};
+
+/**
+ * Get namespace from a namespaced key (e.g., 'names:all_saints' -> 'names')
+ */
+const getNamespace = (key: string): string => key.split(':')[0];
+
+/**
+ * Remove namespace from a key (e.g., 'names:all_saints' -> 'all_saints')
+ */
+const removeNamespace = (key: string): string => key.split(':')[1] ?? key;
+
+/**
+ * Group keys by namespace.
+ */
+const groupKeysByNamespace = (keys: string[]): Record<string, Set<string>> => {
+  return keys.reduce<Record<string, Set<string>>>((acc, key) => {
+    const namespace = getNamespace(key);
+    if (!acc[namespace]) acc[namespace] = new Set();
+    acc[namespace].add(removeNamespace(key));
+    return acc;
+  }, {});
+};
+
+/**
+ * Find missing localized items by comparing computed keys against a locale object.
+ */
+const findMissingInLocale = (
+  computedKeys: string[],
+  locale: Locale,
+  options?: { reverse?: boolean; localeKeyNamesOnly?: boolean }
+): Record<string, string[]> => {
+  const groupedKeys = groupKeysByNamespace(computedKeys);
+  const result: Record<string, string[]> = {};
+
+  for (const [namespace, obj] of Object.entries(locale)) {
+    if (namespace === 'id') continue;
+    if (options?.localeKeyNamesOnly && namespace !== 'names') continue;
+    if (typeof obj !== 'object' || obj === null || Array.isArray(obj)) continue;
+
+    const localeProps = getNestedPropertyNames(obj as Record<string, unknown>);
+    const computedProps = groupedKeys[namespace] ?? new Set();
+
+    const missingItems = options?.reverse
+      ? localeProps.filter((item) => !computedProps.has(item))
+      : Array.from(computedProps).filter((item) => !localeProps.includes(item));
+
+    if (missingItems.length > 0) {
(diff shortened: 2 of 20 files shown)


rites/roman1969/build/data-checks.ts
```diff
@@ -255,27 +258,21 @@ for (let i = 0; i < allCalendars.length; i += 1) {
...
 logIf(LogLevel.ERROR, 'Missing martyrology items:', missingMartyrologyKeys);
 
 /**
- * [3] If there are missing localized 'en' items.
+ * [2] If there are missing localized 'en' items.
  */
 const missingEnLocaleKeys = findMissingLocalizedItems('En', allLocalesKeys);
 logIf(LogLevel.ERROR, `Missing localized '${u('en')}' items:`, missingEnLocaleKeys);
 
 Object.keys(Martyrology.catalog).forEach((key) => allLocalesKeys.add(`names:${key}`));
```
TobiTenno, line 272: should we copy the test i wrote into here instead of where it is in the locale tests right now? 🤔

TobiTenno (commented):
looks great to me. one thought on moving a test to a data check, but otherwise 👍

rites/roman1969/build/data-checks.ts
tukusejssirs: I personally think that moving `data:checks` to either an `eslint` rule or a test makes more sense. However, the location of these tests/checks is not that important to me, but I’d prefer a single location which would be used consistently for all simular tests/checks in the future.

I am open to do anything in this PR to use a single location, just tell me what would you prefer. :wink:

TobiTenno: let's move everything to tests, then, i thin that would make them the most clear. a lint rule would be great, but i'll be honest that i'm not as familiar with writing lint rules

tukusejssirs: How about this: I’ll try to create a lint rule (or more of them as necessary) for `data:checks` and for the test you have mentioned above. If I fail, I’ll just move `data:checks` to a test.

I’ll try to find some time for this tomorrow afternoon UTC +01.

TobiTenno: cool!

tukusejssirs: On second thoughts, it turns out that lint rules can’t anything the currwnt file, that is:

| Check | ESLint Rule? | Reason |
|-------|--------------|--------|
| [1] Missing martyrology items | ❌ No | Requires cross-file aggregation |
| [2] Missing 'en' locale items | ❌ No | Requires cross-file comparison |
| [3] Useless localised items | ⚠️ Partial | Could lint against a generated manifest |
| [4] Martyrology alphabetical sort | ✅ Yes | Single-file check |
| [5] Locale names alphabetical sort | ✅ Yes | Single-file check |
| [6] Missing locale items for GRC | ❌ No | Requires cross-file aggregation |
| [7] Unused martyrology items | ❌ No | Requires cross-file aggregation |

Therefore, U’ll create unit tests instead.

tukusejssirs:
@emagnier, actually, do we need to output informational messages about extra martyrology items? I would just remove the test if you and @TobiTenno agree.

I would also remove the check of `No useless localized name items in any locale`: if we keep them in `martyrology.ts`, let’s keep them the locales too.

tukusejssirs:
I have found in [`rites/roman1969/src/calendars/countries/france/diocese-of-coutances.ts`](https://github.com/romcal/romcal/blob/d2b5465b1a56efbe8b8ec4137da1a9150aea4d9d/rites/roman1969/src/calendars/countries/france/diocese-of-coutances.ts#L189-L194) that `all_holy_bishops_of_the_diocese_of_coutances` is used used in `martyrology` prop of the `all_holy_bishops_and_priests_of_the_diocese_of_coutances` celebration. While I don’t why would we need to split such a celebration by the order of the priesthood (I’d split collective celebrations only if there are _known_ individual _persons_ that are part of that collective, e.g. in this case all know French priests and bishops which would be too long a list anyway; imagine that list for `all_saints` :infinity:), there are some other celebrations where this makes. The issue is that `all_holy_bishops_of_the_diocese_of_coutances` is not localised in `en.ts`, but it is in `martyrology.ts` as a separate item.

I suggest that we not only require all celebration IDs used at the top-level props in calendars to be defined in `en.ts`, but also all martyrology items.

Also I think that all celebration IDs in `en.ts` should be also defined in `martyrology.ts` (e.g. `adalbert_of_prague_bishop_patron_of_poland` used in `poland/index.ts` is not defined in ). Below is a list of celebrations defined in `en.ts` which are missing from `martyrology.ts`; some of them have patronage title added, others are composed of multiple persons or with `companions`, however, there are still some explicitly missing celebrations like `canada_day` or `dedication_of_consecrated_churches` or `dedication_of_the_cathedral_basilica_of_saint_denis_france`). We should somehow make sure everything is defined, like we can remove patronage titles using `*patron*` or companions using `*_companions*`. Unfortunately, there is one multi-personal celebrations (where two or more persons are named explicitly) that does not contain `_and_` as it should (it makes it harder to remove them): `alban_of_britain_julius_of_caerleon_aaron_of_caerleon_martyrs` (I suggest to rename it to `alban_of_britain_julius_of_caerleon_and_aaron_of_caerleon_martyrs`).

<details>
  <summary>Missing from <code>martyrology.ts</code>, but defined in <code>en.ts</code></summary>

1. adalbert_of_prague_bishop_patron_of_poland
2. adeline_de_mortain_abbess_and_the_saints_of_savigny
3. aidan_of_lindisfarne_bishop_and_the_saints_of_lindisfarne
4. alban_of_britain_julius_of_caerleon_aaron_of_caerleon_martyrs
5. all_holy_bishops_and_priests_of_the_diocese_of_coutances
6. all_holy_bishops_of_the_archdiocese_of_lyon
7. all_saints_of_the_archdiocese_of_lyon
8. aloysius_versiglia_bishop_and_callistus_caravario_priest_martyrs
9. andrew_apostle_patron_of_russia
10. andrew_apostle_patron_of_scotland
11. andrew_de_soveral_and_ambrose_francis_ferro_priests
12. andrew_dung_lac_priest_and_companions_martyrs
13. andrew_kim_tae_gon_priest_paul_chong_ha_sang_and_companions_martyrs
14. andrew_zorard_of_nitra_and_benedict_of_skalka_hermits
15. anthony_julian_nowowiejski_bishop_and_companions_martyrs
16. arbogast_of_strasbourg_bishop_patron_of_the_archdiocese_of_strasbourg
17. assumption_of_the_blessed_virgin_mary_patroness_of_france
18. augustine_zhao_rong_priest_and_companions_martyrs
19. basil_the_great_and_gregory_nazianzen_bishops
20. benedict_of_nursia_abbot_patron_of_europe
21. boris_of_kiev_and_gleb_of_kiev_martyrs
22. bridget_of_sweden_religious_copatroness_of_europe
23. brigid_of_kildare_virgin_copatroness_of_ireland
24. canada_day
25. canute_iv_of_denmark_eric_ix_of_sweden_and_olaf_ii_of_norway_martyrs
26. catherine_of_siena_virgin_copatroness_of_europe
27. catherine_of_siena_virgin_copatroness_of_italy_and_europe
28. ceslaus_of_poland_and_hyacinth_of_poland_priests
29. chad_of_mercia_and_cedd_of_lastingham_bishops
30. charles_lwanga_and_companions_martyrs
31. charles_spinola_and_jerome_de_angelis_priests
32. christopher_magallanes_priest_and_companions_martyrs
33. clement_of_ohrid_and_gorazd_of_moravia_bishops_and_companions
34. columba_of_iona_abbot_copatron_of_ireland
35. conrad_of_constance_and_gebhard_of_constance_bishops
36. cornelius_i_pope_and_cyprian_of_carthage_bishop_martyrs
37. cosmas_of_cilicia_and_damian_of_cilicia_martyrs
38. cyril_constantine_the_philosopher_monk_and_methodius_michael_of_thessaloniki_bishop
39. cyril_constantine_the_philosopher_monk_and_methodius_michael_of_thessaloniki_bishop_copatrons_of_europe
40. cyril_constantine_the_philosopher_monk_and_methodius_michael_of_thessaloniki_bishop_slavic_missionaries_copatrons_of_europe
41. david_of_wales_bishop_patron_of_wales
42. dedication_of_consecrated_churches
43. dedication_of_the_cathedral_basilica_of_saint_denis_france
44. dedication_of_the_cathedral_of_notre_dame_de_strasbourg_france
45. dedication_of_the_cathedral_of_saint_john_the_baptist_lyon_france
46. dedication_of_the_notre_dame_de_paris_cathedral_paris_france
47. denis_of_paris_bishop_and_companions_martyrs
48. denis_of_paris_bishop_patron_of_the_archdiocese_of_paris
49. denis_of_paris_bishop_patron_of_the_city_and_of_the_diocese_of_saint_denis
50. epipodius_of_lyon_and_alexander_of_lyon_martyrs
51. eugenia_of_alsace_and_attala_of_alsace_virgins
52. florian_of_lorch_and_companions_martyrs
53. francis_de_sales_bishop_patron_of_the_clergy_of_the_archdiocese_of_lyon
54. francis_diaz_del_rincon_priest_and_companions_martyrs
55. francis_of_assisi_patron_of_italy
56. fructuosus_of_braga_martin_of_braga_and_gerald_of_braga_bishops
57. fructuosus_of_tarragona_bishop_and_augurius_of_tarragona_and_eulogius_of_tarragona_deacons_martyrs
58. george_of_lydda_martyr_patron_of_england
59. gorazd_of_moravia_and_companions
60. gregory_grassi_francis_fogolla_and_anthony_fantosati_bishops_and_companions_martyrs
61. henry_ii_emperor_and_cunigunde_of_luxembourg
62. hilarius_of_toulouse_bishop_and_sylvius_of_toulouse_bishops
63. ignatius_de_azevedo_priest_and_companions_martyrs
64. immaculate_conception_of_the_blessed_virgin_mary_patroness_of_the_philippines
65. immaculate_conception_of_the_blessed_virgin_mary_patroness_of_the_usa
66. irenaeus_of_lyon_bishop_patron_of_the_archdiocese_of_lyon_and_companions_martyrs
67. jacinta_marto_and_francisco_marto
68. jacques_francois_lefranc_martyr_and_martyrs_of_the_revolution
69. jacques_jules_bonnaud_priest_and_companions_martyrs
70. james_apostle_patron_of_spain
71. jean_chevillard_and_companions_martyrs
72. jean_robert_queneau_and_companions_martyrs
73. joachim_and_anne_parents_of_mary
74. joachim_and_anne_patroness_of_the_province_of_quebec_parents_of_mary
75. joan_of_arc_virgin_copatroness_of_france
76. john_de_brebeuf_isaac_jogues_priests_and_companions_martyrs
77. john_de_brebeuf_isaac_jogues_priests_and_companions_martyrs_copatrons_of_canada
78. john_de_brebeuf_priest_and_companions_martyrs
79. john_fisher_bishop_and_thomas_more_martyrs
80. joseph_spouse_of_mary_patron_of_canada
81. kilian_of_wurzburg_bishop_and_companions_martyrs
82. labor_day
83. laud_of_coutances_bishop_patron_of_the_diocese_of_coutances
84. laurence_wang_bing_and_companions_martyrs
85. lawrence_ruiz_and_companions_martyrs
86. leo_ignatius_mangin_priest_and_companions_martyrs
87. marcellinus_of_rome_and_peter_the_exorcist_martyrs
88. margaret_clitherow_anne_line_and_margaret_ward_virgin_martyrs
89. marko_krizin_melchior_grodziecki_and_stephen_pongracz_priests
90. martha_of_bethany_mary_of_bethany_and_lazarus_of_bethany
91. martin_wu_xuesheng_and_companions_martyrs
92. mary_stella_of_the_blessed_sacrament_mardosewicz_and_companions_virgins
93. maternus_of_cologne_valerius_of_trier_and_eucharius_of_trier_bishops
94. maurice_of_agaunum_and_companions_martyrs
95. maurice_of_agaunum_and_companions_martyrs_patron_of_the_diocese_of_angers
96. maurilius_of_angers_bishop_second_patron_of_the_diocese_of_angers
97. mederic_of_autun_and_droctoveus_of_autun_abbots
98. michael_gabriel_and_raphael_archangels
99. modestus_andlauer_and_andrew_bauer_martyrs
100. nereus_of_terracina_and_achilleus_of_terracina_martyrs
101. nykyta_budka_and_vasyl_velychkovsky_bishops
102. odile_of_alsace_abbess_patroness_of_alsace
103. our_lady_mother_of_divine_providence_patroness_of_puerto_rico
104. our_lady_of_angels_patroness_of_costa_rica
105. our_lady_of_aparecida_patroness_of_brazil
106. our_lady_of_guadalupe_patroness_of_the_americas
107. our_lady_of_guadalupe_patroness_of_the_philippines
108. our_lady_of_hungary_patroness_of_hungary
109. our_lady_of_la_salette
110. our_lady_of_lujan_patroness_of_argentina
111. our_lady_of_mount_carmel_mother_and_queen_of_chile
112. our_lady_of_sorrows_patroness_of_slovakia
113. paternus_of_avranches_bishop_and_scubilion_abbot
114. patrick_of_ireland_bishop_patron_of_ireland
115. paul_chen_changpin_and_companions_martyrs
116. paul_miki_and_companions_martyrs
117. perpetua_of_carthage_and_felicity_of_carthage_martyrs
118. peter_and_paul_apostles
119. peter_baptist_blasquez_paul_miki_and_companions_martyrs
120. peter_chanel_priest_patron_of_oceania
121. peter_de_zuniga_and_louis_flores_priests
122. peter_kibe_priest_and_companions_martyrs
123. peter_poveda_and_innocent_of_mary_immaculate_canoura_priests_and_companions_martyrs
124. philip_and_james_apostles
125. philip_evans_and_john_lloyd_priests
126. philip_of_jesus_de_las_casas_paul_miki_and_companions_martyrs
127. pontian_i_pope_and_hippolytus_of_rome_priest
128. pothinus_of_lyon_bishop_blandina_of_lyon_virgin_and_companions_martyrs
129. pothinus_of_lyon_bishop_patron_the_city_of_lyon_blandina_of_lyon_virgin_and_companions_martyrs
130. raymond_costeran_and_companions_martyrs
131. roch_gonzalez_alphonsus_rodriguez_and_john_del_castillo_priests
132. rose_of_lima_virgin_copatroness_of_the_philippines
133. rupert_of_salzburg_and_virgilius_of_salzburg_bishops
134. sancha_of_portugal_and_mafalda_of_portugal_virgins
135. simon_and_jude_apostles
136. sixtus_ii_pope_and_companions_martyrs
137. six_welsh_martyrs_and_companions
138. stanislaus_of_szczepanow_bishop_patron_of_poland
139. stephen_the_first_martyr_and_principal_patron_of_the_archdiocese_of_toulouse
140. teresa_benedicta_of_the_cross_stein_virgin_copatroness_of_europe
141. thanksgiving_day
142. therese_of_the_child_jesus_and_the_holy_face_of_lisieux_virgin_copatroness_of_france
143. thomas_hioji_rokuzayemon_nishi_priest_and_companions_martyrs
144. thursday_of_the_lords_supper
145. timothy_of_ephesus_and_titus_of_crete_bishops
146. translation_of_the_relics_of_odile_of_alsace_abbess
147. ursula_of_cologne_and_companions_virgins
148. vincent_lewoniuk_and_companions_martyrs
149. wenceslaus_i_of_bohemia_martyr_patron_of_the_czech_nation
150. wladyslaw_bladzinski_priest_and_companions_martyrs

</details>

After removing all celebrations that match `*(_and_|_companions|patron)*`, only 14 celebrations remain (incl `alban_of_britain_julius_of_caerleon_aaron_of_caerleon_martyrs` that should be renamed):

<details>
  <summary>Missing celebrations</summary>

  1. [RENAME] alban_of_britain_julius_of_caerleon_aaron_of_caerleon_martyrs
  2. all_holy_bishops_of_the_archdiocese_of_lyon
  3. all_saints_of_the_archdiocese_of_lyon
  4. canada_day
  5. dedication_of_consecrated_churches
  6. dedication_of_the_cathedral_basilica_of_saint_denis_france
  7. dedication_of_the_cathedral_of_notre_dame_de_strasbourg_france
  8. dedication_of_the_cathedral_of_saint_john_the_baptist_lyon_france
  9. dedication_of_the_notre_dame_de_paris_cathedral_paris_france
  10. labor_day
  11. our_lady_of_la_salette
  12. thanksgiving_day
  13. thursday_of_the_lords_supper
  14. translation_of_the_relics_of_odile_of_alsace_abbess

</details>

WDYT, @emagnier and @TobiTenno?

emagnier:
Some thoughts on the points discussed here:

1. The `data:check` task is mainly meant to run in CI to ensure that our data is clean and complete when a PR is submitted. So I agree that an informational message can be removed. Maybe in the future we’ll have a dedicated task to generate statistics, and such informational messages would make more sense there.

2. I also agree that any martyrology entry referenced in a `LiturgicalDay` should indeed be translated.

3. Regarding `all_holy_bishops_and_priests_of_the_diocese_of_coutances`, I created two separate martyrology entries to distinguish the titles of these two groups of people.

4. I’m not opposed to using Jest tests for some checks, but I think `data:check` should remain a separate task for now. Tests should stay focused on code behavior, not on validating large datasets. A future ESLint plugin could centralize these data checks, provide CI validation, and offer immediate IDE feedback, which seems like a better long-term solution.

tukusejssirs:
> 1. The `data:check` task is mainly meant to run in CI to ensure that our data is clean and complete when a PR is submitted. So I agree that an informational message can be removed. Maybe in the future we’ll have a dedicated task to generate statistics, and such informational messages would make more sense there.

I agree with the statistics generation, I have talked some stuff like that before (incl generation of calendar propers, localisation coverage). :+1:

> 2. I also agree that any martyrology entry referenced in a `LiturgicalDay` should indeed be translated.

:+1: Perfect, I’ll a test for that in this PR.

> 3. Regarding `all_holy_bishops_and_priests_of_the_diocese_of_coutances`, I created two separate martyrology entries to distinguish the titles of these two groups of people.

Do you have use case for that? What is the advantage of distinguishing between bishops and priests in the celebration? :thinking:

> 4. I’m not opposed to using Jest tests for some checks, but I think `data:check` should remain a separate task for now.

Why should it remain as a separate task for now? :thinking: If we move the checks to unit tests, the checks/tests would be run in the CI anyway.

> Tests should stay focused on code behavior, not on validating large datasets.

I disagree with this statement.

IMHO unit tests should test _anything_ related to _our code_ (i.e. romcal in this case). And the data must be validated, as otherwise we can so much bad data and bad data results in inferior quality of romcal which results in hardly noticable bugs.

> A future ESLint plugin could centralize these data checks, provide CI validation, and offer immediate IDE feedback, which seems like a better long-term solution.

As I have stated in [this comment](https://github.com/romcal/romcal/pull/1001#discussion_r2596139588), some checks from `data:check` cannot be ported to ESLint because ESLint can process one file at a time, but some checks must use various/many files. I don’t say that some checks cannot be done at all, but they can be done either in a limited or very complex or resource-intensive way.

Either way, previously we have approved that the data (sources, `martyrology.ts`/catalogue, calendars with celebrations, localisations) would be moved to a separate package, then we could move these changes out there and in the core package and calendar revision packages would test the code part only.

emagnier:
The two separate martyrology entries for `all_holy_bishops_and_priests_of_the_diocese_of_coutances` allow flexibility for future features, localized displays, and better data organization/searchability, which could add long-term value.

I fully agree that data must be validated in CI, that’s not in question. My point is about **separation of concerns**: unit tests are meant to verify a single functionality or small scope of code, while heavy multi-file data validations are better handled by `data:check`. This keeps tests fast, readable, and maintainable, while still ensuring data integrity.

To clarify my previous point regarding ESLint: it was suggested as a possible way to complement `data:check` and provide IDE feedback for simpler checks. I agree that multi-file validations cannot be fully moved there, but it remains an interesting option for future improvements. `data:check` continues to be the main mechanism for complex validations.

tukusejssirs:
> The two separate martyrology entries for `all_holy_bishops_and_priests_of_the_diocese_of_coutances` allow flexibility for future features, localized displays, and better data organization/searchability, which could add long-term value.

I still don’t see much value in having two separate martyrology entries. IMHO the localisation of this particular celebration will always be done at the `all_holy_bishops_and_priests_of_the_diocese_of_coutances` level, but okay, you want what you’t talking about and it is out of the scope of this PR. :man_shrugging:

> My point is about **separation of concerns**: unit tests are meant to verify a single functionality or small scope of code, while heavy multi-file data validations are better handled by `data:check`. This keeps tests fast, readable, and maintainable, while still ensuring data integrity.

As for the fast tests: all tests (incl my new tests and coverage) take about ~22 seconds to run and there 3 failing suites and 23 failing tests (the more output, the longer it takes). Based on the CI logs (where it look ~20 seconds, it add a second for many tests. And when we move to Vitest (which is my desire), it could be even faster. YMMV

As for readability and maintainability: that all depends on the tests author and preferences.

> To clarify my previous point regarding ESLint: it was suggested as a possible way to complement `data:check` and provide IDE feedback for simpler checks.

My idea is to use a single, consistent way to test our data and code.

As for IDE feedback, yes, I get that lint errors/warnings could be useful. How important that is, that is another question.

> I agree that multi-file validations cannot be fully moved there, but it remains an interesting option for future improvements.

I consider that a crucial thing to test them now, therefore, I am creating the unit tests.

> `data:check` continues to be the main mechanism for complex validations.

I frankly disagree with that.

---

Another possible issue: why don’t we localise all these periods, but only `epiphany` and `holy_week` in the locales?

- ChristmasOctave
- DaysBeforeEpiphany
- DaysFromEpiphany
- ChristmasToPresentationOfTheLord
- PresentationOfTheLordToHolyThursday
- HolyWeek
- EasterOctave
- EarlyOrdinaryTime
- LateOrdinaryTime

TobiTenno:
> data:check continues to be the main mechanism for complex validations.

i agree that these should become tests, and we should just run tests to validate data before we publish, it keeps the step as a standard lifecycle hook instead of a completely custom one

For all the rest of that, let's do those other checks that we 'd want to move (or other modifications) in a separate PR, please. this one was fairly straightforward, just touching this part of locales. if we want more enhancements, let's do it in a separate PR.

tukusejssirs:
To let you guys review the WIP changes (if you want), I have just pushed a commit with all the tests I have created.

Pending tasks:

- [ ] review the tests for false positives/negatives;
- [ ] fix the found issues (I think of skipping the addition of Brazilian celebrations and localisations which I have already mostly fixed in #993);
- [ ] remove `data:checks` NPM script along with the JS script file and its call in the CI.

OFC we need to have the consensus on some of the parts, that is the main reason I want to share with you the tests I have created before I consider them done.

---

Issue the tests have found:

1. Martyrology Data (2 failures):
  - Missing translations in `en.ts` for some used martyrology keys;
  - Martyrology catalog keys not sorted alphabetically.

2. Locale Meta Keys (7 failures):
  - `cs.ts`: Missing cycles and ordinals
  - `es.ts`: Missing cycles
  - `la.ts`: Missing cycles
  - `pl.ts`: Missing cycles and ordinals

3. Locale Calendar Keys (3 failures):
  - `pt-br.ts`: Missing GRC keys and Brazil calendar keys
  - `es.ts`: Missing keys for Spanish-speaking country calendars

4. Calendar Chronological Sorting (25 failures):
  - Celebrations not sorted chronologically in many calendars like Bolivia, BosniaHerzegovina, Brazil, CostaRica, England, Finland.

---

> i agree that these should become tests, and we should just run tests to validate data before we publish, it keeps the step as a standard lifecycle hook instead of a completely custom one

I agree with this. :+1:

> For all the rest of that, let's do those other checks that we 'd want to move (or other modifications) in a separate PR, please. this one was fairly straightforward, just touching this part of locales. if we want more enhancements, let's do it in a separate PR.

To be honest, I wanted to expand the scope of this PR, as a check in `data:checks` blocked me, thus I wanted to fix it by refactoring it to unit tests, but along the line, I have found a few issues (bugs in the `data:checks` script, missing stuff, etc), therefore, I tried to do it in a more robust way, while still focusing on the same checks for `martyrology.ts`, calendars and locales.

Could you make a quick (not thorough) review of the changes I have made in the latest commit please? Does it seem to much of the scope broadening? If yes, what would you move in a separate PR?

I personally would keep all changes from the latest commit and update the PR title and description instead. However, I am open for merging the changes in any number of PRs. :man_shrugging:

rites/roman1969/__tests__/util/getNestedPropertyNames.ts
```diff
@@ -0,0 +1,29 @@
...
+  typeof obj === 'object' && obj !== null && !Array.isArray(obj);
+
+/**
+ * Get all nested property names from an object using dot notation.
+ * e.g., { a: { b: 1, c: 2 } } -> ['a.b', 'a.c']
+ * @param obj - The object to extract property names from
+ * @param parentKey - The parent key for recursion (internal use)
+ * @returns Array of nested property names in dot notation
+ */
+export const getNestedPropertyNames = (obj: Record<string, unknown>, parentKey?: string): string[] => {
```
TobiTenno: we shouldn't reinvent the wheel. use an existing library like [`jsonata`](https://jsonata.org/)

TobiTenno: this would eliminate the need for extra utility tests

rites/roman1969/__tests__/util/isObjectPropsSortedAlphabetically.test.ts
```diff
@@ -0,0 +1,65 @@
+import { isObjectPropsSortedAlphabetically, findUnsortedKeys } from './isObjectPropsSortedAlphabetically';
+
+describe('isObjectPropsSortedAlphabetically', () => {
```
TobiTenno, line 3: i disagree with this in general, objects in JS are not guaranteed to remain "sorted" in memory

TobiTenno (commented):
overall seems good, i think some things should use reliable libraries if we need the utility directly

rites/roman1969/__tests__/util/isObjectPropsSortedAlphabetically.test.ts
tukusejssirs: This was true pre-ES2015, but the spec has changed.
Since ES2015 ([ECMA-262 §9.1.12 `[[OwnPropertyKeys]]`](https://262.ecma-international.org/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys)), property order is guaranteed for ordinary objects:

1. integer indices: ascending numeric order;
2. string keys: creation/insertion order;
3. symbol keys: creation order.

Now, in GitHub Actions (both [dev](https://github.com/romcal/romcal/blob/d2b5465b1a56efbe8b8ec4137da1a9150aea4d9d/.github/workflows/dev-publishing.yml#L20-L23) and [romcal CI](https://github.com/romcal/romcal/blob/d2b5465b1a56efbe8b8ec4137da1a9150aea4d9d/.github/workflows/romcal-ci.yml#L21-L24) workflows), use [`.nvmrc`](https://github.com/romcal/romcal/blob/d2b5465b1a56efbe8b8ec4137da1a9150aea4d9d/.nvmrc) which is set to `lts/iron` (`v20`/`ES2023`).

For object literals, ‘creation order’ equals source order. Since all romcal keys (at least those we test) are non-integer strings, `Object.keys()` will return them in the exact order they appear in the source.

This function isn’t trying to keep objects sorted—it’s checking whether the keys are already in alphabetical order when enumerated. For our `snake_case` string keys, that enumeration order reliably reflects source order.

The only caveat: if we ever had integer-like keys (e.g. `10`, `2`), V8 would re-order them numerically. But that doesn’t apply to the test data.

That said, I’ll make some changes to fix some minor issues / edge cases which I have found while researching sources for this comment:

1. integer-like keys would break the check (not a problem for romcal (all the keys are `snake_case` identifiers), but it is worth documenting):

```ts
const obj = {'10': 'x', '2': 'y', 'foo': 'z'}
Object.keys(obj) // ['2', '10', 'foo'] — V8 reorders integers!
```

2. `localeCompare()` is locale-sensitive:

```ts
'ä'.localeCompare('z') // -1 in 'de', +1 in 'sv'

// For ASCII lowercase keys this won't matter, but for robustness we could use this instead
a.localeCompare(b, 'en', {sensitivity: 'base'})
```

rites/roman1969/__tests__/util/getNestedPropertyNames.ts
tukusejssirs: Okay, I’ll use `jsonata`.

rites/roman1969/build/sort-data.ts
```diff
@@ -0,0 +1,211 @@
+/**
+ * Utility script to check and fix alphabetical sorting of data files.
+ *
+ * Usage:
+ *   npx ts-node build/sort-data.ts check   - Check for sorting issues
+ *   npx ts-node build/sort-data.ts fix     - Fix sorting issues in place
+ */
```
emagnier, line 7: **Suggestion**: I’m not a big fan of this kind of utility that parses the codebase using regular expressions. It adds an extra layer of complexity to data validation and makes future refactoring much harder, which is inevitable in the life of a project. This is typically the kind of work that should be handled by a **linter** instead.

rites/roman1969/__tests__/calendar-data.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    prevKey = key;
+    prevDateDef = dateDef;
+  }
+
+  return issues;
+};
+
+describe('Calendar Data Validation', () => {
+  describe('All celebration keys used in calendars must exist in en.ts names', () => {
+    const allCelebrationKeys = extractAllCelebrationLocaleKeys();
```
emagnier, line 288: **Issue**: `extractAllCelebrationLocaleKeys()` is called at the describe level, so it is executed even when the test is skipped. This can be costly.
**Recommendation**: move the extraction inside the test or use `beforeAll()` with caching.

rites/roman1969/__tests__/calendar-data.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // True movable feasts (Easter-dependent) sort after all fixed dates
+    const addDay = 'addDay' in dateDef ? (dateDef.addDay ?? 0) : 0;
+    const subtractDay = 'subtractDay' in dateDef ? (dateDef.subtractDay ?? 0) : 0;
+    return 90000 + addDay - subtractDay;
+  }
+
+  return 99999;
+};
```
emagnier, line 236: **Question**: Where does this ordering rule come from? I’m not against proposing a standard order for the definitions, but the logic here feels a bit complex and, in the end, not really natural chronologically. Why not base it on a typical year instead, and propose a sorting based on actual dates of this selected year?

In any case, it’s better to avoid using magic numbers directly and define named constants instead, for example:

```ts
const SORT_VALUE_UNKNOWN = 99999;
const SORT_VALUE_MOVABLE_BASE = 90000;
const SORT_VALUE_MONTH_MULTIPLIER = 100;
const SORT_VALUE_LAST_WEEK_APPROX = 22;
```

rites/roman1969/__tests__/calendar-data.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+describe('Calendar Data Validation', () => {
+  describe('All celebration keys used in calendars must exist in en.ts names', () => {
+    const allCelebrationKeys = extractAllCelebrationLocaleKeys();
+    const enNamesKeys = Object.keys(locales.En.names ?? {});
+    const missingKeys = findMissingInArray(allCelebrationKeys, enNamesKeys);
+
+    it('should have all celebration names in en.ts', () => {
+      expect(missingKeys).toEqual([]);
+    });
```
emagnier, line 294: **Issue**: if the test fails, the error message does not list the missing keys.

rites/roman1969/__tests__/proper-of-time.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    it('should have no missing i18n keys overall', () => {
+      expect(missingKeys).toEqual([]);
+    });
+  });
+
+  // Note: Season keys for all locales are checked in locale-data.test.ts
+  // under "All locales must have required meta keys" -> seasons
+
+  describe('Expected season keys are present in en.ts', () => {
+    const expectedSeasonKeys = [
```
emagnier, line 106: **Suggestion**: I feel like this test is redundant with the previous one.

rites/roman1969/__tests__/calendar-data.test.ts
```diff
@@ -0,0 +1,311 @@
...
+ * Get the parent calendars from a calendar class.
+ */
+const getParentCalendars = (CalendarClass: CalendarDefInstance): CalendarDefInstance[] => {
+  const proto = (
+    CalendarClass as unknown as {
+      prototype?: { constructor?: { prototype?: { ParentCalendars?: CalendarDefInstance[] } } };
+    }
+  ).prototype;
+  return proto?.constructor?.prototype?.ParentCalendars || [];
+};
```
emagnier, line 98: **Issue**: fragile access to prototypes using multiple `as unknown as` casts.
**Recommendation**: rely instead on the calendar definitions, which already provide this information.

rites/roman1969/__tests__/proper-of-time.test.ts
```diff
@@ -0,0 +1,163 @@
...
+import { locales } from '../src/locales';
+import { RomcalConfig } from '../src/models/config';
+import { ProperOfTime } from '../src/proper-of-time/proper-of-time';
+import { Locale } from '../src/types/locale';
+
+/**
+ * Extract all i18n keys used by the Proper of Time.
+ * These come from the i18nDef property of each liturgical day definition.
+ */
+const extractProperOfTimeI18nKeys = (): Set<string> => {
```
emagnier, line 11: **Suggestion**: I think there’s some refactoring to be done here with `extractAllCelebrationLocaleKeys`, to better avoid code repetition.

rites/roman1969/__tests__/locale-data.test.ts
```diff
@@ -0,0 +1,392 @@
...
+    'Venezuela',
+    'PuertoRico',
+  ],
+  It: ['Italy'],
+  Pl: ['Poland'],
+  Cs: ['CzechRepublic'],
+  Sk: ['Slovakia'],
+  La: ['GeneralRoman'],
+  Ta: ['India'],
+};
```
emagnier, line 125: **Todo:** I would advise against doing this mapping from the tests, and instead recommend adding a property to the calendar definitions indicating which locales a particular calendar is generally used in. This keeps the source of truth in the data (not the tests) and ensures that the tests stay up to date, even if new calendars are added.

emagnier (changes requested):
Thanks for adding these validation checks. I appreciate the effort, but I'd like to clarify the **separation of responsibilities**.

Following good software architecture practices, unit tests should focus on small-scope, single-functionality checks, while large-scale, multi-file data integrity validations should remain separate. Jest isn't well-suited for these large-scale data integrity checks because it adds unnecessary overhead, slows development, produces less clear error messages, and cannot distinguish between errors, warnings, and informational messages.

Writing custom checks in `data:checks` is the appropriate approach here: it provides structured output with severity levels (ERROR/WARN/INFO), targeted execution only when needed (CI/CD or pre-build), and proper exit codes. As the author and main developer of this version of Romcal, I want to ensure we keep `data:checks` (or an equivalent mechanism) for large-scale data validations, while unit tests cover smaller, targeted checks.

rites/roman1969/__tests__/util/isObjectPropsSortedAlphabetically.test.ts
emagnier: In any case, I think it’s more the role of a linter to handle this, rather than a custom check function & test.

TobiTenno: @emagnier agreed, i think this kind of check should stay in the linter, since those should also be able to auto-fix things like this, which i'd assume wouldn't be part of our tests

tukusejssirs:
@emagnier, thanks for the thoughtful feedback on my changes. I’ll fix you findings them ASAP (last week I’ve been busy both at work and at home). However, note that I shared the latest commit just to be able to discuss the refartoring of `data:checks` to unit tests; I know there are still some rough edges and I personally didn’t actually review my own code yet (that’s why I have converted the PR to a draft for now). :wink:

I want to address [your point #4](https://github.com/romcal/romcal/pull/1001#issuecomment-3623552248) in more detail since we seem to have different perspectives on where data checks should live. Let me share some research I did (using Claude Code / AI) on this topic.

## TL;DR

I believe consolidating data checks into Jest is the right approach for romcal, and I think we already have established precedent for this in our own codebase.

---

## 1. We Already Use Jest for Data Validation

Before looking at external examples, let’s look at what **we already do in romcal**:

### Snapshot Testing of Calendar Data

In `calendar-builder.test.ts:521-541`, we have:

```typescript
describe('Testing calendar snapshots', () => {
  test('General Roman Calendar in gregorian year', async () => {
    const gregorianCalendar2020 = await new Romcal().generateCalendar(2020);
    const gregorianCalendar2021 = await new Romcal().generateCalendar(2021);
    const gregorianCalendar2022 = await new Romcal().generateCalendar(2022);

    expect(gregorianCalendar2020).toMatchSnapshot();
    expect(gregorianCalendar2021).toMatchSnapshot();
    expect(gregorianCalendar2022).toMatchSnapshot();
  });
  // ...
});
```

According to `git blame`, this code was written by _you_ on 12 Nov 2022. :wink:

This is **literally data validation using Jest**. We’re validating that the generated calendar data matches expected snapshots across multiple years. If ‘[t]ests should stay focused on code behavior, not on validating large datasets,’ then this existing code contradicts that principle.

### The New Data Tests Follow the Same Pattern

The new tests in `martyrology-data.test.ts`, `locale-data.test.ts`, and `calendar-data.test.ts` are doing exactly the same thing: validating that our data is correct, complete, and consistent. They’re a natural extension of what we already do.

---

## 2. External Examples: This Is Common Practice

I did some research, and using test frameworks for data validation is actually very common.

### Major Data Quality Frameworks

- **[Deequ](https://github.com/awslabs/deequ)** - AWS’s library is explicitly designed for defining **‘unit tests for data’**. It’s used for large-scale data quality on Apache Spark.

- **[Great Expectations](https://greatexpectations.io/)** - Another major framework where tests serve as both executable validation and documentation ([‘tests are docs, and docs are tests’](https://legacy.017.docs.greatexpectations.io/docs/0.14.13/reference/core_concepts/#:~:text=tests%20are%20docs%2C%20and%20docs%20are%20tests)).

Both frameworks explicitly embrace the concept of _unit tests for data_. This isn’t a fringe idea—it’s a recognised best practice in data engineering.

### Jest-Specific Examples

- **[jest-json-schema](https://github.com/americanexpress/jest-json-schema)** (American Express) - Provides `toMatchSchema` matcher for validating JSON data against schemas in Jest tests.

- **[i18n.test.js Gist](https://gist.github.com/Michaelvilleneuve/8808ba2775536665d95b7577c9d8d5a1)** - Jest tests specifically for:
  - Checking missing translation keys
  - Finding unused translation keys
  - Validating translation completeness across languages

  This is almost exactly what our locale validation tests do!

- **[babel-plugin-tester](https://github.com/babel-utils/babel-plugin-tester)** - Uses Jest to validate fixture files and transformations. The author explicitly prefers fixture-based testing to inline code because it’s easier to read with syntax highlighting.

### i18next Community Practices

In [i18next community discussions](https://github.com/i18next/i18next/discussions/1848), developers recommend building automated testing frameworks that compare translation keys across languages to ensure completeness. Many teams implement this directly in Jest.

---

## 3. The ESLint Plugin Argument

You mentioned:

> A future ESLint plugin could centralize these data checks, provide CI validation, and offer immediate IDE feedback

I see the appeal, but I have concerns:

### ESLint Processes One File at a Time

Some of our data checks need to cross-reference multiple files:

- **Martyrology validation** - needs to check that keys used in calendars exist in the martyrology catalogue
- **Locale validation** - needs to compare calendar definitions against locale files
- **Calendar sorting** - needs to traverse inheritance chains across multiple calendar files

ESLint rules receive one file at a time. To do cross-file validation, you’d need to either:
1. Build complex caching mechanisms that persist data between file analyses
2. Load external files during rule execution (resource-intensive)
3. Use ESLint’s newer `meta.docs.url` pattern with custom validators (still complex)

In other words, it would require significant effort to achieve what Jest already does out of the box.

### It’s Speculative Future Work

Building a custom ESLint plugin is a significant engineering effort. We’d need to:
- Design the plugin architecture
- Implement custom rules
- Handle the cross-file challenges
- Maintain two systems during transition

Meanwhile, Jest is already set up, running in CI, and familiar to contributors.

### IDE Feedback Is Nice-to-Have

Yes, real-time feedback would be great. But our data doesn’t change frequently — it’s not like code where you’re constantly making changes. Running `npm test` after modifying data files is perfectly acceptable.

---

## 4. Devil’s Advocate: Arguments for Separation

To be fair, here are the legitimate arguments for keeping data checks separate:

| Argument                                                           | Counter-argument                                                                  |
|--------------------------------------------------------------------|-----------------------------------------------------------------------------------|
| **Conceptual separation**: unit tests = code, data checks = data   | We already mix these with snapshot tests. The boundary is fuzzy in practice.      |
| **Different failure modes**: a test failure vs. data quality issue | Jest’s output clearly shows which test failed. We can use descriptive test names. |
| **Slower tests**: data checks might be slower                      | Our data checks run in milliseconds. Not a real concern.                          |
| **Different audiences**: devs vs. data maintainers                 | Everyone runs `npm test`. Having one command is simpler.                          |

The strongest argument is conceptual purity. But I think practical benefits outweigh theoretical elegance here.

---

## 5. Practical Benefits of Jest Integration

| Benefit                    | Explanation                                    |
|----------------------------|------------------------------------------------|
| **Single test runner**     | One `npm test` command runs everything         |
| **Unified CI integration** | No need to manage multiple NPM scripts in CI   |
| **Consistent reporting**   | Same output format for all validations         |
| **Watch mode**             | `npm test -- --watch` works for data files too |
| **Filtering**              | Can run specific data tests with `-t` flag     |
| **Coverage reports**       | Data validation is tracked in coverage         |

---

## 6. Proposed Compromise

If the separation concern is really important to you, we could:

1. Keep all data validation in Jest
2. Place data tests in a clearly separated location: `__tests__/data-validation/` or files named `*.data.test.ts`
3. Add a separate npm script: `npm run test:data` that filters to only data tests

This gives us:
- ✅ Single test runner and CI integration
- ✅ Clear logical separation in the file structure
- ✅ Ability to run data checks separately if needed
- ✅ All the tooling benefits of Jest

---

## Conclusion

I think we should consolidate data checks into Jest because:

1. **We already do it** — snapshot tests validate calendar data
2. **It’s common practice** — Deequ, Great Expectations, jest-json-schema, i18n validation gists all show this pattern
3. **ESLint is impractical** — cross-file validation is complex, and it’s speculative future work
4. **Practical benefits** — one command, one CI step, familiar tooling

What do you think? Happy to discuss further or implement a compromise approach!

TobiTenno:
i generally agree that unit tests are a good clean way of doing what we're wanting. trying to make lint rules work (a large lift of its own) for this for live feedback in a definition that will most certainly change in the near future as we try to get more information into locale and definitions seems like folly to me.

rites/roman1969/__tests__/util/isObjectPropsSortedAlphabetically.test.ts
TobiTenno: this is one of the few i think should be a lint rule, as this is already part of https://github.com/haenah/eslint-plugin-sort-properties, we could use something out of the box to do this instead of building ourselves

TobiTenno:
@tukusejssirs we need to move forward with this, we're getting progressively more stale, and it's time to finally get a plan for v3 stable release to go out

emagnier:
Thanks for the detailed research @tukusejssirs! I could respond point-by-point (with Claude's help too 😄), but I think we'd end up in an endless back-and-forth that wouldn't move us forward.

A quick clarification on snapshot tests: I wrote those specifically during the v2→v3 rewrite to catch regressions between versions. They're not really data quality checks, they verify that output stays consistent across code changes. And honestly, they're less useful now that the rewrite is done.

Here's my perspective: I still believe we should separate concerns and avoid putting data quality/integrity validation directly in Jest. I see `data:check` evolving into an **audit tool**, something less binary that can report errors, warnings, and suggestions for future improvements. (Quick note: ESLint was just an example that could consume parts of this audit tool later. Let's close that topic.)

That said, Jest can absolutely *consume* this audit tool to assert on specific critical aspects. Something like:

```typescript
const results = runLocalizationAudit();
const errors = results.filter(r => r.severity === 'ERROR');
expect(errors).toEqual([]);
```

This way we get the best of both worlds: a rich audit system with proper severity levels, and Jest validating what truly needs to block CI.

@TobiTenno agreed, we need to move forward. @tukusejssirs, your work on these tests is valuable and will be useful later. I'd like to go with this approach (audit tool + Jest as consumer). For now, let's keep this PR focused on the original scope (Teresa translations + the minor data-check adjustment). Could you move the refactoring work to a separate branch/PR?

TobiTenno:
@emagnier which specific parts should be split? I can go split out the PRs so we can continue discussing, since you and @tukusejssirs are vastly busier than I am.

emagnier:
Thanks @TobiTenno for offering to help move this forward! 🙏
I'd suggest splitting this into two PRs:

- **A new one** with the two original commits (data-check adjustment + Teresa translations), so we can merge it quickly.
- **This one** to continue exploring how to improve our data validation.

@TobiTenno feel free to go ahead if you have the bandwidth!

TobiTenno:
Will do when I do. Life got more chaotic since I offered,  but I will when I can.

TobiTenno:
part 1: https://github.com/romcal/romcal/pull/1081
part 2: https://github.com/romcal/romcal/pull/1082 (sort-keys impl)

TobiTenno:
closing, as this has been supplanted by #1081 & #1082

PR #1082, opened 2026-03-01
style(eslint): enforce object sort order

TobiTenno:

Use eslint-plugin-perfectionist to enforce object key sort order, similar to #1001's sort-keys implementation, but enforcing w/ eslint where possible

## Concept

- Localization
- Documentation (keeping keys sorted)


TobiTenno:
@tukusejssirs @emagnier seems like sorting everything breaks precedences, from what i'm seeing and in the tests. i fixed, as far as i could tell, the moduleNameMapper for resolution purposes, but other sorts, i'm assuming in the calendars, causes issues for precendences.

emagnier:
Really like the idea of enforcing sort order with ESLint, that's a great addition for code quality. A few thoughts:

**Pure data: 👍**
For locales, translations, and similar data-only objects, this is a clear win. It removes unnecessary diff noise and makes contributions more predictable. Love it.

**Semantic ordering**
For objects where the key order carries meaning for readability, like `{ start, end }` vs `{ end, start }`, I'm not so sure. Code should stay readable and intuitive, and sometimes that matters more than alphabetical consistency.

**Potential hidden impacts**
You already fixed the `Precedences` and `RanksFromPrecedence` cases, but there might be other places in the codebase where definition order has a functional impact that's less obvious. With 111 files touched, it's hard to be sure without a thorough audit, and I don't have everything fresh in mind right now.

**Suggestion: opt-in rather than opt-out**
Rather than sorting everything and then excluding problematic objects as we find them, what about flipping the approach? We could enable the rule specifically on the file patterns where we know it's safe and beneficial (locales, types, some configs...), and leave the rest untouched. It reduces the scope of the PR and protects us against the unknowns. I'd prefer we go that route before merging. What do you think?

As a follow-up (separate PR would be easier to review): I had a quick look at the plugin docs, and `sort-objects` seems to support custom groups with `elementNamePattern` and `useConfigurationIf`, which could let us enforce a semantic sort order where it matters (enums like Precedences, calendar definitions, etc.). Could add quite a bit of config though, so worth evaluating case by case.

TobiTenno:
Given the failures, it seems more likely that the error is from sorting the calendar itself rather than just a bundle of translation strings. I'll work on making the rule only apply to files that are the translation files first and then we can expand from there. How does that sound, @emagnier?

emagnier:
Sounds good, that's exactly the approach I'd go for 👍

TobiTenno:
@emagnier everything rebased & formatted locales & utils

emagnier (approved):

PR #1127, opened 2026-04-17
RFC: Roman 1962 rite (Tridentine Mass) support - opening for discussion, not ready to merge

AlexBocken:
### Summary

Hi romcal maintainers — first, thank you for the library. This PR represents roughly 2–3 weeks of work adding **1962 Roman Rite (Tridentine / Missale Romanum 1962, Rubricae 1960) Mass calendar support** to romcal. It is large (~47 commits, ~250 files).

> [!IMPORTANT]
> **This is not a merge request  but a discussion opener.** I don't think this is ready to land as-is, and I wouldn't want you to treat it that way. What I'd like is your feedback on the architecture, the scope, and whether the overall direction is one you'd be willing to adopt. If yes, I'm happy to reshape, split, or rewrite whatever is needed to get to something mergeable. If no, I'd still find the review valuable.

The **guiding principle throughout: stay as close to the existing 1969 rite as possible, and only deviate where the rubrics genuinely require it.** Where the two rites share structure, they now share code; where they don't, the divergence is documented and deliberate.

---

### What you get

A new `Romcal1962` class with an API intentionally shaped like `Romcal`:

```ts
import { Romcal1962 } from 'romcal/1962';

const r = new Romcal1962({
  localeId: 'en',            // 14 locales available, see below
  includePropers: true,      // attach Mass proper texts (various levels of support, latin and english most extensive)
  propersLocales: ['la', 'en'],
  commemorationLimit: 'solemn', // Rubricae 1960 §111–113
});

const year = await r.generateCalendar(1962);
// → Record<isoDate, LiturgicalDay1962[]>
//   index 0 = primary Mass after occurrence+transfer,
//   index ≥ 1 = surviving commemorations in precedence order.

const easter = await r.resolveDate('1962-04-22');
easter[0].key;                  // 'easter_sunday'
easter[0].name;                 // 'Easter Sunday'
easter[0].propers.introit.la;   // 'Resurréxi, et adhuc tecum sum…'
easter[0].propers.introit.en;   // 'I arose, and am still with Thee…'

// Parity with 1969's id-based lookup:
const stJoseph = await r.getOneLiturgicalDay(
  'saint_joseph_spouse_of_the_blessed_virgin_mary',
  { year: 1962, computeInWholeYear: true },
);
```

Coverage today: Tempora cycle (Advent→Pentecost), full 1960 sanctoral, occurrence + commemoration + forward-transfer rubrics, Class II Lord precedence, vigil suppression, commemoration cap, Mass propers in Latin + 13 vernaculars. The Divine Office (Breviary), Christmas multi-Mass, and the special Good Friday / Holy Saturday / Pentecost Vigil liturgies are documented gaps.

---

### Repo reshape: `rites/` + `packages/`

To integrate 1962 cleanly I had to pull the 1969 code out of the repo root and share its abstractions. Net structural change:

```
After
────────
rites/
  roman1969/   ← existing romcal rite-specific code
  roman1962/   ← new rite
packages/
  romcal-core/      ← shared structural contracts
  calendars/        ← shared CalendarDef<Entry>
  proper-of-time/   ← shared Easter/Advent anchors
  i18n/             ← shared i18next bootstrap
  constants/        ← Months/Weekdays/DayOfWeek
  easter/           ← already shared
  lunar-new-year/   ← already shared (not required for roman1962)
  config/
```

Public npm import surface is preserved:

```ts
import { Romcal } from 'romcal';          // 1969, unchanged
import { Romcal1962 } from 'romcal/1962'; // new
```

This reshape is one of the bigger things I want your opinion on — see "Open questions" below.

---

### Extraction sequence (the alignment phases)

To avoid a single monolithic refactor, I landed the shared extraction as an explicit, logged plan at [`docs/alignment/00-plan.md`](docs/alignment/00-plan.md). Each phase was a commit-sized change that left both rites building green.

| Phase | What moved out to shared | Effect on 1969 |
|------|--------------------------|----------------|
| **A** | Read-only audit of every public symbol | None |
| **B** | `@internal/proper-of-time` — `computeAnchors`, `YearAnchors`, date helpers, `ProperOfTimeYear<Entry>` generic | Delegation only, no API change |
| **C1–C3** | `@internal/calendars` — abstract `CalendarDef<Entry>` + `flattenCalendarChain`; 1969's `CalendarDef` re-parented onto it | Zero public-API change; 57 country classes untouched |
| **D1–D4** | `@internal/romcal-core` — rite-neutral structural contracts (`BaseLiturgicalDay`, config shape, year cache). `Romcal1962` refactored to mirror 1969's class layout (`LiturgicalDay1962`, `LiturgicalDayDef1962`, `LiturgicalDayConfig1962`, `Calendar1962`) | Minor: now implements shared contracts |
| **E1–E4** | `@internal/i18n` — i18next factory + name-translator. Both rites consume it. Shared `IRomcalConfigRoot` base config contract | Minor: i18n plumbing now from shared pkg |

Hard divergences I deliberately did *not* try to unify (documented in `00-plan.md`):

- **Rank system** — 1969 uses `Precedence` (1–17); 1962 uses `Rank1962` (Class I–IV) + `numericRank`. Different ordinals, different semantics.
- **Commemoration caps** — 1962 honours Rubricae 1960 §111–113; 1969 has none.
- **Octaves** — 1962 has Class I/II/III octaves with explicit octave days; 1969 has only Easter + Christmas.
- **Vigils & forward-transfer** — 1962-only machinery.
- **Martyrology metadata, cycles (Sunday/Weekday/Psalter)** — 1969-only concepts.

The 1969 public API is therefore untouched. The rites discriminate by `rite: 'roman1969' | 'roman1962'` on the base day type.

---

### Where the data comes from

Mass propers, rubrics, vernacular text and the 1960 kalendarium are sourced from the excellent [**divinum-officium**](https://github.com/divinumofficium/divinum-officium) project (MIT-licensed, so license-compatible).

The submodule lives at `divinum-officium/` and is processed by a one-shot importer at `rites/roman1962/build/import-divinum-officium/` which emits normalized JSON into `rites/roman1962/data/`:

```
rites/roman1962/
├── build/import-divinum-officium/   ← parser, rank map, emit pipeline
│   ├── parser.ts                       DO's `.txt` → AST
│   ├── rank-map.ts                     DO rank strings → Class I–IV
│   ├── key-mapping/                    DO short keys (e.g. `Pasc0-0`)
│   │                                   → readable slugs (`easter_sunday`)
│   ├── emit-propers.ts                 per-section Mass text
│   ├── emit-locales.ts                 per-locale name bundles
│   └── *-name-overrides.ts             hand-audited corrections per-lang
└── data/
    ├── calendar-1960.json           ← 1960 sanctoral with Class I–IV ranks
    ├── tempora.json, sancti.json, commune.json
    ├── source.json                  ← IMPORTER_VERSION, provenance
    └── propers/{la,en,de,fr,…}/…    ← per-locale proper text
```

**Key shape migration (M10):** DO's short keys (`Adv1-0`, `Quadp3-3`, `Pasc0-0`, `C5`, `07-22`) are rewritten at the importer's emit boundary to readable snake_case slugs aligned with 1969's conventions (`advent_1_sunday`, `quinquagesima_wednesday`, `easter_sunday`, `common_of_confessor_not_pontiff`, `mary_magdalene_penitent`). **10 celebrations that coincide with 1969 reuse the exact same key** — so anything downstream that switches rites gets free parity on those days. A `key-shape.test.ts` guards the invariant.

Supported locales (auto-generated, hand-audited where needed):

```
cs, cs-schaller, da, de, en, es, fr, hu, it, la, nl, pl, pt, uk
```

---

### Regional overlays: Switzerland as the reference implementation

As I live in Switzerland I created a reference implementation for diocese/country level additions for the 1962 rite. without diocesan propers is only half-useful: local patrons, translations, and Class I diocesan patronal feasts are what actually appear in a parish Ordo. So the 1962 rite ships with a working Swiss overlay tree as both a feature and a reference implementation for future countries.

The overlay system **deliberately mirrors 1969's overlay architecture**, which is the main reason Phase C of the alignment extracted `@internal/calendars` as shared infrastructure. Both rites now consume the same abstract `CalendarDef<Entry>` base class and the same `flattenCalendarChain` parent-walk helper. The differences are narrow and confined to the entry payload.

1969-side, Swiss territory today is a single national file (`countries/switzerland/index.ts`). 1962-side, the overlay tree is deeper because the 1962 rite has more pronounced diocesan proper calendars:

```
rites/roman1962/src/calendars/
├── regions/
│   └── europe.ts                     ← regional anchor
└── countries/
    └── switzerland/
        ├── index.ts                  ← national: Switzerland extends Europe
        ├── diocese-of-basel.ts       ← extends Switzerland
        ├── diocese-of-chur.ts        ← extends Switzerland
        ├── diocese-of-lausanne-geneva-fribourg.ts
        ├── diocese-of-lugano.ts
        ├── diocese-of-sankt-gallen.ts
        ├── diocese-of-sion.ts
        └── abbey-of-saint-maurice.ts ← territorial abbey
```

Each overlay is a class extending `CalendarDef<CalendarOverlayEntry>` with a `parents` chain and an `entries` array — the same shape 1969 uses. Example:

```ts
// rites/roman1962/src/calendars/countries/switzerland/diocese-of-chur.ts
import { CalendarDef } from '@internal/calendars';
import { Switzerland } from '.';
import type { CalendarOverlayEntry } from '../../types';

export class Switzerland_Chur extends CalendarDef<CalendarOverlayEntry> {
  readonly parents = [new Switzerland()];

  readonly entries: readonly CalendarOverlayEntry[] = [
    {
      mmdd: '12-03',
      fileKey: 'saint_lucius_of_chur_bishop_and_martyr_patron',
      class1962: 1,
      rank1962: 'ClassI',
      numericRank: 6,
      mode: 'add',                              // 'add' | 'replace' | 'upgrade'
      mass: {
        colors: ['Red'],
        references: {                           // points at COMMONS_1962
          Introitus: 'Commune/common_of_one_martyr_pontiff',
          // …
        },
        rubrics: { gloria: true, credo: true },
      },
      names: { la: '…', de: '…', fr: '…', en: '…', it: '…' },
    },
  ];
}
```

The chain walks parents-first, id-deduped, with the nearest override winning — identical traversal semantics to 1969. Consumer code is also identical in shape:

```ts
import { Romcal1962 } from 'romcal/1962';
import { Switzerland_Chur } from '@romcal/rite-roman1962/calendars';

const r = new Romcal1962({ calendar: Switzerland_Chur, localeId: 'de' });
// St. Lucius of Chur (Dec 3, ClassI) outranks St. Francis Xavier in the diocese of Chur;
// St. Francis Xavier drops to commemoration per Rubricae 1960 §107–111.
```

So: if you already know how 1969's `@romcal/calendar.france` works, you already know how the 1962 overlay system works. The only 1962-specific thing on the entry payload is the rubric-bearing fields (`class1962`, `rank1962`, `numericRank`, `mass.rubrics`, `mass.references`) — everything else is shared.

The Swiss tree is not trying to be canonical Swiss liturgy for every diocese — **it's meant as a working template** that covers every overlay mechanic (regional anchor, national defaults, diocesan proper patrons, territorial abbey) so future contributors can copy the structure for their own country. Sources are cited inline in each overlay file (Roman Martyrology, diocesan proper missals, Catholic Encyclopedia 1913 where modern sources are sparse). Where no official diocesan propers have been digitised, the `references` field points at the appropriate `Commune/*` so the Mass is at least valid.

If you're uncomfortable shipping a specific country's overlay tree in-core, I'm equally happy to extract it as `@romcal/calendar1962.switzerland` mirroring the 1969 calendar-plugin convention — the overlays use the public API, so it's a rehousing, not a rewrite.

---

### Milestones (commit-traceable)

For reviewers who want to go milestone-by-milestone, the work is documented in [`docs/1962/`](docs/1962/):

| M | What |
|---|------|
| M1 | Scaffold `rites/roman1962/` |
| M2 | Divinum-Officium importer |
| M3 | Proper of Time resolver |
| M4 | Proper of Saints resolver (1960 kalendarium) |
| M5 | Rubrics engine: occurrence + commemoration + forward-transfer |
| M6 | Mass-propers resolver (inline, ref, entry-level references, Sunday fallback) |
| M7 | `Romcal1962` public class |
| M8 | Rubrics polish: Class II Lord precedence, vigil suppression, commemoration cap |
| M9 | Localization alignment with 1969 (i18next + per-locale bundles) |
| M10 | DO → readable slug key migration |

Each milestone has its own design doc. The alignment phases (A–E) that shaped the shared `packages/` came *after* M10 — the rite was initially standalone, and alignment was a deliberate second pass once the 1962-specific logic had stabilized.

---

### Open questions I'd love your input on

1. **Is the `rites/ + packages/` reshape a direction you'd accept?** If not, I can keep 1962 as a plugin-style external package, though sharing abstractions is much harder that way. I'd rather have your preference than guess.
2. **Shared `@internal/romcal-core` contracts.** I tried not to constrain future rites (e.g. Ambrosian, Mozarabic) but you know the library better. Are the abstractions at the right level? Too shallow, too deep, wrong seams?
3. **`romcal/1962` as a subpath export.** Alternatives: a separate npm package (`@romcal/rite-roman1962`), or a full merge into the default export with a `rite` config option. Happy to reshape.
5. **Data provenance.** I pull from divinum-officium with attribution. If you'd prefer a different source, a different vendoring strategy (vendor vs. submodule vs. no sources in-repo), or stronger license-notice machinery, please say.
7. **PR shape.** This PR is huge. I'm very willing to split it along the phase boundaries (Phase B, then C, then D, then E, then M1–M10 in sequence) if that makes review realistic. I kept it as one PR here so the full picture is visible — it's meant as a conversation starter, not a merge candidate.

Thanks for reading this far and thanks again for romcal.

Heads-up, I did use Claude Code for a large part of this PR but of course aim to keep the permanent code (not build-scripts) human-reviewed.


TobiTenno:
hi Alex, we already have a 1962 folder ready to use. the current shape already leverages packages and rites, if you want to separate out utilities from the 1969 package, that would be acceptable, but i don't think the whole reshape is welcome.

TobiTenno:
given that `Romcal1962` needs to conform to the general type of `Romcal` anyway, `Romcal1962` should be treated as a base new calendar, not an entirely calendar _system_.

TobiTenno:
You requested comments, but you're not responding to them...

AlexBocken:
> You requested comments, but you're not responding to them...

Apologies, first wanted to draft something up according to your comments.

> given that `Romcal1962` needs to conform to the general type of `Romcal` anyway, `Romcal1962` should be treated as a base new calendar, not an entirely calendar _system_.

Totally valid argument. There is a large overlap here and we can definitely incorporate some logic we are using in Romcal. Currently trying via inheritance to get something in that direction (detailed write-up to follow).

A rough outline of my current approach:

I add three virtual methods on `Calendar`, one factory on `Romcal`, one discriminator field on `LiturgicalDay`, and one declarative `octave` input — all with default implementations that reproduce today’s behavior byte-for-byte. There is no consumer change; 100% of the existing test suite remains green. This surface enables any rite variant (1962, Ambrosian, Mozarabic, Dominican, …) to subclass and plug in rite-specific precedence / transfer / factory behavior without forking the engine.

> "given that Romcal1962 needs to conform to the general type of  
> Romcal anyway, Romcal1962 should be treated as a base new calendar,  
> not an entirely calendar system."

The problem I had was that 1962's rubrics (Rubricae 1960 §111–113), forward transfer + vigil suppression, and its own `LiturgicalDay1962` shape couldn’t be expressed via the existing extension seams. `CalendarDef` subclassing only covers inputs + proper-of-time (which is already cleanly subclassable — `ProperOfTime extends CalendarDef`). But the `Calendar` reducer and the `new LiturgicalDay(…)` factory were hardcoded inside `Calendar`/`Romcal`.

Three narrow virtuals unlock all three 1962 needs. I introduce no new types and no new concepts. The file count and class count remain the same.

This does change the signature of some of the classes 1969 uses (as we use templates now) but the API itself does not change in any way.


Does this sound reasonable to you in approach?


>hi Alex, we already have a 1962 folder ready to use. the current shape already leverages packages and rites, if you want to separate out utilities from the 1969 package, that would be acceptable, but i don't think the whole reshape is welcome.

Noted, reverting reshape. Fair point.

rites/roman1962/data/calendar-1960.json
TobiTenno, line 1: this should use the typescript definition, and not try to create a new class or numeric rank system. if you want to integrate equivalences between Class & Precedences, that would be more appropriate of an equivalence, as you're doing with `numericRank`.

TobiTenno:
I think this is too much all in one place with how many commits are here, i can't feasibly review all of the changes. i'm going to work on doing some additional scaffolding work on the project soon (pulling out types into packages so there's not so much purely internal-to-1969 stuff)

AlexBocken:
Totally understandable. I was thinking of restructuring my commits more
cleanly, now that I have gotten it to a working point for the full PR. The
commits have quite some noise which would be reduced this way.What would be
your estimated timeline for the restructuring? Any way I can support you
such that also the 1962 rite could benefit best from the restructuring?

TobiTenno:
i can't give a very good timeline sadly, i've been really really busy with work, so i'm hoping that I have time to get it started this weekend, planning the following: (several very similar to your restructure)
```diff
rites/
   roman1969/
   roman1962/
packages/
+  build/					# Move build scripts to consume any/all rites & packages
+  docs/					# Separate Documentation from just appearing to apply (or being nested in) roman1969
   calendars/
#   proper-of-time/	# Haven't planned what this _should_ be, going to review your changes more here
+  constants/			# Months/Weekdays/DayOfWeek
+	 types/					# Move shared types from between packages into here 
   easter/
   lunar-new-year/
   config/
+	martyrology/		# Add a separation for Martyrology catalogs to allow collating several (such as Roman martyrology & Melkite synaxaria)
```

TobiTenno:
i'm really sorry for the super long delay, i'm working on other projects getting them updated, too, and should be back around to this shortly (hopefully this month)

TobiTenno:
sorry, almost caught up. should still be august on some weekend

emagnier:
Hi @AlexBocken,

First of all, thanks a lot for all the work you’ve put into this. There’s a huge amount here, and it’s honestly really nice to see this kind of contribution to romcal. Support for the 1962 calendar has also been requested a few times already, so I’m especially glad to see this moving forward.

Also, sorry for only getting back to you now. I haven’t yet had enough time to look through everything properly. I’ll try to spend more time on it and look at it more closely. I probably won’t be able to do a full review straight away, so I don’t want to promise a timeline I can’t keep, but I do want to keep the discussion moving and make sure this work doesn’t get lost.

After thinking about it a bit more, my current feeling is that the 1962 and 1969 calendars should have two separate generators, as well as separate base/temporal calendars. The differences between the two sets of calendar rules are simply too significant for the 1962 calendar to be implemented by extending and adapting the 1969 generator. They can still live in the same project, but I think they should remain independent at that level.

I would apply the same idea to calendar definitions: the 1962 and 1969 definitions should be separate. However, we should keep a shared canonical identifier underneath whenever they genuinely refer to the same saint or mystery. This would let us link and compare the two calendars without pretending that their dates, ranks, names or rules are necessarily the same.

The general infrastructure should of course remain shared wherever it makes sense: basic date utilities, localization infrastructure, common identifiers, and other rite-neutral parts.

For the Roman Martyrology, I think a shared catalogue would make sense. Each entry could include some informational metadata listing the official Martyrology editions in which it appears, for example:

- the preconciliar _Martyrologium Romanum, editio quarta post typicam_ (1956), together with the 1960 _Variationes in Martyrologio Romano_ used with the 1962 liturgical books;
- the postconciliar _Martyrologium Romanum, editio typica_ (2001), later revised as the _editio typica altera_ (2004).

I see this edition membership mainly as provenance and reference information, rather than something that should directly drive calendar generation. An entry could belong to one or several editions.

This could also be a good opportunity to store optional beatification and canonization dates. These would have a practical use later: when generating a historical calendar, they could help prevent a Blessed or Saint from appearing before the relevant beatification or canonization had taken place. Of course, these dates would remain optional, especially for ancient Saints for whom no formal date exists.

I think these dates should remain separate from a calendar entry’s `effectiveFrom` date, though. Canonization is a historical fact about the person, while `effectiveFrom` tells us when a particular celebration actually became part of a specific universal, national, diocesan or religious calendar. They may sometimes be the same date, but not necessarily. We may eventually also need an `effectiveTo` date when a celebration is removed, replaced or changed.

For translations, I think we can have a shared base for simple and universal names, for example the ordinary name of a Saint. Beyond that, I would keep the translations separate. Liturgical titles and feast names are not always identical between the two calendars, even when they refer to the same person or mystery.

More generally, I would strongly prefer correctness and auditability over trying to eliminate every bit of duplication. Some explicit duplication is better than a theoretically elegant shared model that becomes liturgically ambiguous.

This is only my current thinking before looking more deeply into all the details, but if you continue working on it in the meantime, I think keeping the 1962 calendar logic independent from 1969, and only sharing the stable and genuinely common parts, would be the safest direction.

Thanks again for all this work. I’d really like us to find a way to bring it into romcal progressively, in a form that remains easy to understand, verify and maintain.

TobiTenno:
part 1 of my reorg is up at #1219, i've got several more parts planned:
- packages/proper-of-time
- packages/generator
- packages/builder
- roman1962 rules + temporale
- getting to the meat of publishing roman1962 general roman

PR #1219, opened 2026-08-27
docs: move contributor docs into docs/

TobiTenno:

Housekeeping that clears the ground for [RFC 1127](https://github.com/romcal/romcal/pull/1127), plus npm trusted publishing. Nothing here changes the public API or the generated calendar data.

It is four separate commits, reviewable in order:

**1. `docs:` move contributor docs into `docs/`**

`CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, `AUTHORS.md` and `CHANGELOG.md` move out of the repository root, which GitHub still recognises under `docs/`. References are repointed, and the link checker now ignores `mailto:` and `tel:` URLs so contact links no longer read as broken. `.npmignore` keeps excluding `CODE_OF_CONDUCT.md` from the published package via its new path.

**2. `ci:` publish with npm trusted publishing**

Adds `trust:check` and `trust:sync`, which report and configure trusted publishers for `romcal` and every `@romcal/calendar.*` bundle. Results are cached so that checking and configuring do not each spend a 2FA window on 100+ lookups. New packages cannot start with trusted publishing, since npm requires a package to exist before a publisher can be attached, so `publish-new-packages.yml` covers that case on demand. The trust commands need npm >= 11.15.0, hence `lts/krypton` in `.nvmrc`.

**3. `build:` move tooling into `.config/` and drop Prettier for ESLint**

Tooling configuration moves into `.config/`. ESLint 9 does not search that folder, so every invocation names the config explicitly: npm scripts, lint-staged, the husky hooks, CI and the editor. `typescript-eslint` resolves its `project` paths relative to the config file, so it gets `tsconfigRootDir`.

Prettier is removed entirely. `@stylistic` covers JS/TS, and `jsonc` / `yml` / `markdownlint` cover data and documentation files, because `@eslint/json` and `@eslint/markdown` are lint-only by design. `build/bundle-doc.ts` calls the ESLint Node API in place of `prettier.format`.

This also fixes tooling that was already broken: `lint-staged` and `sort-package-json` were invoked by the pre-commit hook but declared in no `package.json`, and `commitlint` lived in the `roman1969` workspace while its config and CI are at the root. Deletes `packages/config`, which was Jest-only despite its name, and the dead `tsconfig.release.json`.

**4. `style(stylistic):` reformat for the new rules**

Three ternaries carried comments between the operator and its operand, which no `operator-linebreak` setting accepts. The comments move above the operator, the double ternary picking the liturgical year is flattened, and a member chain inside a ternary branch in `calendar.ts` becomes a plain conditional so the `indent` rule can express it.

### Notes for review

A naive port of the Prettier options reformatted roughly 700 lines, most of it worse rather than merely different, so a few rules deliberately diverge and are commented as such in `.config/eslint.config.mjs`:

| Rule | Setting | Why |
| --- | --- | --- |
| `operator-linebreak` | `after`, with `?` `:` `\|` before | Prettier breaks before union `\|` but after intersection `&`; the default rewrote leading-pipe unions to trailing pipes |
| `quote-props` | off | Prettier left numeric keys as authored, so nothing reproduces both `0:` in the locale tables and `'2001':` in the Easter fixtures |
| `indent-binary-ops` | off | Wanted the operands of a multi-line intersection type flush left, where Prettier indented them |
| `indent` | `TSConditionalType` ignored | Would flatten Prettier's indentation of nested conditional types such as `CamelToUpperSnakeCase` |
| `quotes` | `avoidEscape` | Keeps double quotes around strings that contain single quotes |
| `spaced-comment` | `markers: ['=', '/']` | Preserves the `//= ===` section banners in `dates.ts` |
| `max-len` | error | unbreakable calendar/locale ids get `eslint-disable-next-line`; strings/templates/urls are ignored |

`engines` stays at `>=18` even though `.nvmrc` is now `lts/krypton`: `.nvmrc` pins the development toolchain while `engines` constrains consumers, so raising it would be a breaking change that does not belong here.

Commits 2 and 3 are not individually lint-clean; only the branch tip is. That is inherent to keeping the reformat in its own commit.

## Concept

- Infrastructure
- Documentation

## Example code

No consumer-facing change. The public API, the `exports` map and the published bundle layout are all untouched, so existing usage is unaffected:

```typescript
import Romcal from 'romcal';
import { GeneralRoman_En } from '@romcal/calendar.general-roman';

const romcal = new Romcal({ localizedCalendar: GeneralRoman_En });
const calendar = await romcal.generateCalendar(2026);
// identical output to before this pull request
```

Verified locally:

```bash
npm run lint    # 0 errors
npm run build   # 104 calendars in 13 locales, ~18s
npm test        # 495 tests, 6 snapshots, all passing
npm run doc     # regenerates docs/calendar-plugins.md byte for byte
```

The doc check is the important one for the Prettier removal: the ESLint-based generator reproduces the committed `docs/calendar-plugins.md` exactly, so dropping Prettier changed no generated output.

## Follow-ups

Not in this pull request, tracked for the RFC 1127 groundwork:

- Extract `packages/proper-of-time` (date primitives and year anchors)
- Extract `packages/generator` (core engine, `CalendarDef`, i18n, cycle framework)
- Extract `packages/builder`, turning the rite-local `build/` directory into an argument-driven CLI that decides which rite, calendars, locales and formats to build
- The root `CONTRIBUTING.md` content is stale (Travis CI, "docs live in README") next to the real v3 instructions in `docs/contribute-to-romcal.md`; merging the two is a separate change


TobiTenno:
@emagnier when you have time, i'm refactoring and reorganizing the repo a little bit more to move more of the commonly usable logic out to its own package, there's a few more PRs to come, i just want to do this stable step-by-step and not introduce too much at once.

TobiTenno:
@emagnier i'm going to merge this saturday if i don't hear from you

.github/workflows/dev-publishing.yml
```diff
@@ -86,19 +90,24 @@ jobs:
...
         run: npm run build
 
       - name: Run tests
         run: npm run data:checks && npm run test:without-coverage
 
+      - name: Check trusted publishing configuration
+        # Fails with the list of packages missing a trusted publisher, instead of
+        # a 403 in the middle of publishing. Run `npm run trust:sync` to fix.
+        working-directory: rites/roman1969
+        run: tsx build/trusted-publishing.ts
```
emagnier: This step can't work here: the job no longer has any npm credentials (the `NPM_TOKEN` env is removed from the workflow, and there's no `registry-url`), and OIDC only kicks in for `npm publish`, not for the trust API. I tried `npm trust list romcal` anonymously and got `E401 Bearer token authorization is required`. So every package ends up as error, the script exits 1, and nothing gets published or tagged.

I'd drop this step from the workflow (keep `trust:check` as a local command), or at most make it `continue-on-error: true` so it only writes the step summary.

rites/roman1969/build/trusted-publishing.ts
```diff
@@ -0,0 +1,593 @@
...
+    const done = await sync(entries);
+    if (!done && !options.json) log(chalk.yellow('\nSome packages still need a trust configuration.'));
+  }
+
+  report(entries);
+  writeStepSummary(entries);
+  saveCache(entries);
+
+  const actionable = entries.filter((e) => e.status !== 'ok').length;
+  process.exit(actionable > 0 ? 1 : 0);
```
emagnier, line 592: `unpublished` counts as actionable, so once a PR adds a calendar (like #1209) the next push to dev fails at this gate and the whole release stops, romcal included. Today a new bundle just gets published on the next run.

New packages will always need a manual first publish anyway. Can we treat `unpublished` as a warning: skip those packages in `publish.ts`, list them in the summary, and still publish everything else?

.github/workflows/publish-new-packages.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.ref }}
+
+      - name: Install Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: '.nvmrc'
```
emagnier, line 58: `registry-url` is missing here. Without it, setup-node doesn't write the `.npmrc` that makes npm pick up `NODE_AUTH_TOKEN`, and npm doesn't read `NPM_TOKEN` at all. So `npm publish` and `npm trust` in this job run unauthenticated and fail with `ENEEDAUTH`, even though the secret is there.

```yaml
with:
   node-version-file: '.nvmrc'
   registry-url: 'https://registry.npmjs.org'
```

Same thing applies to the comment in `publish.ts` about falling back to `NPM_TOKEN`: that fallback doesn't exist as written.

rites/roman1969/build/publish.ts
```diff
@@ -1,59 +1,90 @@
...
  */
-const afterPublish = (data: Results): void => {
-  if (data.oldVersion !== data.version) {
-    log(` ✓ Package "${data.name}" published: ${data.oldVersion} → ${data.version} (${data.tag})\n`);
-  } else if (data.oldVersion === data.version) {
-    log(` ✓ Package "${data.name}" is already published: ${data.version} (${data.tag})\n`);
-  } else {
-    log(data, '\n');
-  }
+const publishedVersions = async (name: string): Promise<string[] | null> => {
```
emagnier, line 26: Two things from the old `@jsdevtools/npm-publish` behaviour that got lost:

 - `publishedVersions` throws on any non-404 error (429/5xx on 100+ sequential GETs is not unlikely) and the IIFE has no catch, so a single flaky response kills the loop halfway with no summary. A try/catch per package that counts into `failures` would keep the per-package accounting you already have.

rites/roman1969/build/publish.ts
```diff
@@ -1,59 +1,90 @@
...
-    token,
-    tag,
-    dryRun,
-  });
-  afterPublish(mainData);
+/**
+ * Publish with OIDC (trusted publishing) when the workflow provides an ID token,
+ * otherwise fall back to the credentials in the environment (NPM_TOKEN).
+ */
+const publish = (target: Target): boolean => {
```
emagnier, line 40: Same comment as above about the old `@jsdevtools/npm-publish` behaviour that got lost:

 - It ran `npm publish --ignore-scripts` by default. Without it, the root `prepare: husky` now runs on every publish. Probably fine on the runner, but I'd rather keep `--ignore-scripts` and not find out.

rites/roman1969/build/publish.ts
```diff
@@ -1,59 +1,90 @@
...
+    log(` - Publishing ${target.name}@${target.version}${dryRun ? chalk.gray(' (dry-run)') : ''}${isNew ? chalk.gray(' (new package)') : ''}`);
+    if (publish(target)) {
+      log(` ✓ Package "${target.name}" published: ${target.version} (${tag})\n`);
+    } else {
+      failures += 1;
+      log(`${chalk.red(' ✗')} Package "${target.name}" failed to publish\n`);
+    }
+  }
+
+  if (onlyNew && !targets.length) log(chalk.dim('No packages found to publish.'));
```
emagnier: `if (onlyNew && !targets.length)` can't be hit since `targets` always contains the root package.

rites/roman1969/build/trusted-publishing.ts
```diff
@@ -0,0 +1,593 @@
...
+        }
+        allDone = false;
+        continue;
+      }
+    }
+
+    const created = createTrust(entry.name);
+    if (created.ok) {
+      log(`${chalk.green('✓')} ${entry.name}${options.dryRun ? chalk.dim(' (dry-run)') : ''}`);
+      entry.status = 'ok';
```
emagnier: With `--dry-run, npm trust … --dry-run` exits 0 without creating anything, but the entry still gets marked `ok` and written to the cache. A `trust:check` in the next 60 minutes then says everything is configured, and the next dev publish 403s. Only mark `ok` when not in dry-run (or don't save the cache after a dry run).

.config/eslint.config.mjs
```diff
@@ -0,0 +1,360 @@
...
+        'warn',
+        {
+          code: 120,
+          ignoreComments: true,
+          ignoreRegExpLiterals: true,
+          ignoreStrings: true,
+          ignoreTemplateLiterals: true,
+          ignoreUrls: true,
+        },
+      ],
```
emagnier, line 102: Prettier used to fail CI at 120 columns. Now `max-len` is a warning that `npm run lint` doesn't fail on (currently 11 warnings, exit 0 at the branch tip), and nothing auto-fixes it.

If we drop Prettier, the ESLint rules should be as strict as it was, otherwise line length is no longer enforced and formatting could drift with each merge.

emagnier (changes requested):
Thanks for this. I used AI (Claude Fable) to help with this review and dig deeper into the security side, and double-checked the findings myself.

Commits 1 and 4 are fine as far as I'm concerned, and commit 3 too (the `.config/` cleanup, the Prettier → stylistic table, the lint-staged/commitlint fixes) with one thing to fix on `max-len`, see inline 6. If you want to go faster, feel free to open a dedicated PR with those three and merge it once that's done.

Commit 2 is why I can't approve the PR as is: the new check step fails on every push to `dev` (`npm trust list` needs auth, I get a 401 without credentials), a new calendar would block the whole release, and the `publish-new-packages` workflow can't authenticate. Details inline.

I ran `trust:check` locally: 107 of the 108 packages already have their trusted publisher, nice. Only `@romcal/calendar.france.carcassonne-narbonne` is missing, so worth a `trust:sync` before this lands, otherwise the first dev publish after merge will fail on that one. Apart from that, the three points above are really the only thing between this PR and a working OIDC flow. The token-based path for new packages is short-lived anyway, with npm dropping direct publish for those tokens in January 2027.

Nice groundwork overall, the tooling part in particular.

rites/roman1969/build/trusted-publishing.ts
TobiTenno: that's intended, as tokens are not going to be able to publish new packages soon. we're going to have to do the publish and trusted setups for those manually now.

.github/workflows/publish-new-packages.yml
TobiTenno: that is the default registry, that's a no-op.

rites/roman1969/build/trusted-publishing.ts
TobiTenno: addressed

rites/roman1969/build/publish.ts
TobiTenno: addressed

.github/workflows/publish-new-packages.yml
emagnier: so we're good

.config/eslint.config.mjs
TobiTenno: the problem is all of the excessively long l10n keys, which is why i switched these to warn, so that the fixable ones could be fixed as needed.

otherwise addressed

rites/roman1969/build/publish.ts
TobiTenno: addressed

.github/workflows/publish-new-packages.yml
TobiTenno: added it anyway.

rites/roman1969/build/trusted-publishing.ts
TobiTenno: addressed, regardless

rites/roman1969/build/publish.ts
TobiTenno: addressed

.github/workflows/dev-publishing.yml
TobiTenno: removed

TobiTenno:
i also got carcassonne set up

emagnier (approved):
Thanks for the quick fixes, all good for me 👍

emagnier:
Side note, not for this PR: now that we're moving to OIDC publishing, workflow linters like zizmor (security) and actionlint (syntax) would be worth adding to CI. zizmor would flag `phips28/gh-action-bump-version@master` for instance, which runs unpinned in the job that can publish romcal.

TobiTenno:
sure, i'll do a pass on those once i get the general re-layout done to add more syntax/security protection. that was part of why i added pinact a while back

PR #1223, opened 2026-08-29
feat(calendar): support adjacent-Sunday transfer rules

emagnier:
## What is the purpose of this pull request?

Add support for the transfer rules used by the England and Wales calendars:

- transfer Epiphany to the adjacent Sunday when January 6 falls on Saturday or Monday;
- transfer All Saints to the adjacent Sunday under the same conditions;
- preserve the behavior of explicit `epiphanyOnSunday` overrides.

This also introduces configurable temporal anchor exceptions that can be inherited by child calendars.

Closes #185.

## Concept

- Calendar Definition
- Date calculation
- Configuration
- Documentation

Rather than introducing another calendar-specific boolean, this PR models the behavior as declarative temporal anchor exceptions. This makes the current rule explicit while providing an extension point for similar transfer rules in other jurisdictions, without expanding the public configuration with additional one-off flags.

## Liturgical sources

The implemented transfer rules follow the norms published for England and Wales:

- [Decree of the Congregation for Divine Worship and the Discipline of the Sacraments](https://www.liturgyoffice.org.uk/Calendar/Info/Holydays-Decree-CDW.pdf): Epiphany is observed on January 6, except when it falls on Saturday or Monday, in which case it is transferred to the following or preceding Sunday respectively.
- [Liturgy Office - Holydays of Obligation](https://www.liturgyoffice.org.uk/Calendar/Holydays.shtml): documents the Saturday/Monday transfer rule for the relevant holydays in England and Wales, including All Saints.
- [Official Liturgical Calendar 2024](https://www.liturgyoffice.org.uk/Calendar/2024/Calendar-2024.pdf): confirms the concrete 2024 and 2025 dates covered by the tests.

## Example code

England and Wales bundles apply their transfer rules automatically:

```typescript
import Romcal from 'romcal';
import { England_En } from '@romcal/calendar.england';

const calendar = await new Romcal({
  localizedCalendar: England_En,
}).generateCalendar(2024);
```

Temporal anchor exceptions can also be configured explicitly:

```typescript
const calendar = await new Romcal({
  temporalOverrides: {
    anchorExceptions: {
      epiphany: [
        {
          when: { dayOfWeek: 'saturday' },
          then: { transferTo: 'sunday' },
        },
      ],
    },
  },
}).generateCalendar(2024);
```

---

- [x] By submitting this pull request, I confirm that I have read and agree to the [Code of Conduct](../docs/CODE_OF_CONDUCT.md).

PR #1224, opened 2026-08-29
feat(package): separate utilities, build, generator from rite

### LINKED ISSUE #1226: [Bug]: i18next 26 discards the custom interpolation formatters, so romanize/capitalize/uppercase are inert
### What happened?

The custom interpolation formatters registered in `RomcalConfig` — `romanize`, `capitalize` and `uppercase` — never run. Every locale string that relies on one renders its raw value instead, in all 10 locales that use them.

Italian, where an unformatted value is unmistakable:

```
martedì della 1 settimana di Avvento
```

English, where the damage is subtler because most source strings happen to already be capitalised:

```
first Sunday of Advent
```

The cause is upstream. romcal passes the formatter as `interpolation.format` to `i18next.createInstance()`, and i18next 26 reads that option (`Interpolator` sets `this.format` from it) but then overwrites it during init:

```js
// node_modules/i18next/dist/cjs/i18next.js:1851
this.options.interpolation.format = s.formatter.format.bind(s.formatter);
```

The user-supplied function is replaced by the formatter service before any translation runs, so it is discarded. i18next also removed `format` from the `InterpolationOptions` type in v26 — the typing was upstream signalling that the option is no longer the way to do this.

The fix is to register the three formatters through the formatter service instead:

```ts
i18n.services.formatter.add('romanize', (value) => toRomanNumber(parseInt(value, 10)));
```

### What didn't happen?

The formatters should be applied, giving `martedì della I settimana di Avvento` and `First Sunday of Advent`.

### Installation Method

CommonJS Module

### Installation Version

3.0.0-dev.139

### Usage Method

Node.js

### Runtime Version

Node.js v24 and v26

### Relevant log output

```shell
Locale strings affected, by count of `romanize` / `capitalize` / `uppercase` uses:

  en 16   de 16   it 13   fr 13   sk 9
  pt-br 9  pl 9   la 9    es 9    cs 9
```

### Code Sample

```typescript
const { Romcal } = require('@romcal/calendar.general-roman');
const { GeneralRoman_It } = require('@romcal/calendar.general-roman/cjs/it');

const cal = await new Romcal({ localizedCalendar: GeneralRoman_It }).generateCalendar(2026);
const day = Object.values(cal).flat().find((d) => /settimana di Avvento/.test(d.name));

console.log(day.name);
// actual:   martedì della 1 settimana di Avvento
// expected: martedì della I settimana di Avvento
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

---

Found while splitting the shared engine out of `roman1969` (#1224). It is pre-existing and unrelated to that PR: the same check against a build from before the split produces byte-identical output. That PR only makes the *type* error visible, because the new `@internal/generator` runs `tsc` directly whereas the rite emits its declarations through the builder, which ignores the exit code. It works around the typing with a cast and a comment pointing here, leaving the runtime behaviour exactly as it is today.

TobiTenno:

Scaffolds `@internal/proper-of-time`, the package that will hold the date
primitives and liturgical year anchors shared by every Roman Rite calendar.

This is groundwork for RFC #1127. Adding a 1962 rite means either duplicating the
date arithmetic that `rites/roman1969/src/utils/dates.ts` owns today, or having one
rite import from a sibling. This package is where that logic goes instead.

Nothing moves yet. This PR adds the manifest, tsconfigs, jest config, a documented
entry point and a README, and wires them into the root `build` and `clean` scripts,
the `tsconfig.base.json` path alias and both ESLint project arrays. The extraction
then lands as a reviewable move, against unchanged snapshots, rather than as a move
mixed with the plumbing needed to receive it.

The package answers "when does this fall?", never "what does this mean?". Easter
already lives in `@internal/easter` and the lunar calculation in
`@internal/lunar-new-year`; this package builds the shared anchors on top of them.
The primitives and the fixed anchors move here. Anything shaped by a rite's rubrics
stays with the rite: Ordinary Time, `epiphany` and `baptismOfTheLord` with their
movable-Sunday options, the `ascension` and `corpusChristi` transfers, and the
season boundaries keyed by a rite's own `Season` enum. Anchors will be pure
`(year: number) => Date` functions, so they carry no rite configuration.

The adjacent-Sunday transfer work in #1223 reworked `dates.ts` while this was in
progress, and the split survives it. `temporalOverrides` is applied after an anchor
is computed — `epiphany` works out its base date, then passes it through
`#applyAnchorExceptions` — so the base computation moves here while the exception
layer stays with the rite, next to the config describing it. `ShiftableAnchor` is
`'epiphany'` today but is written to grow, so any anchor extracted here should be
assumed to acquire a rite-side override layer eventually. The README records this
so the extraction does not have to rediscover it.

Lint, `tsc --noEmit` and the build are clean, the full suite passes at 502 tests
and 6 snapshots, and the release build emits `dist/src/index.js` to match the
manifest's `main`. `packages/builder` is the same groundwork for the build tooling,
in a sibling PR.

## Concept

- Infrastructure

## Example code

The package exports nothing yet and the public API is untouched. After the
extraction, a rite gets its anchors from the package instead of computing them:

```typescript
import { easterSunday, firstSundayOfAdvent } from '@internal/proper-of-time';

const easter = easterSunday(2026);
const advent = firstSundayOfAdvent(2026);
```


TobiTenno:
@emagnier sorry for so many file touches. i tried to break it up as much as possible, but these 3 sections (splitting out the new packages) has a lot of moves and new types to touch to contain generator and builder better

packages/proper-of-time/src/anchors.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return date;
+};
+
+/** December 8, moved to the following Monday when it falls on a Sunday. */
+export const immaculateConceptionOfMary = (year: number): Date => {
+  const date = getUtcDate(year, 12, 8);
+  return date.getUTCDay() === 0 ? addDays(date, 1) : date;
+};
```
emagnier, line 130: This Monday transfer is a 1969 rule. In 1962 the feast is kept on the Sunday (it outranks the Advent Sunday). Same idea for `annunciation` above, even if both rites agree there: I'd have the package return the raw date (Dec 8, Mar 25) and let each rite apply its own transfer rules, otherwise the "when, not what" boundary won't hold.

emagnier (changes requested):
The PR grew well beyond what the description says: it's not a scaffold anymore, it moves the whole engine and the build tooling. The work is clean (byte-identical dist, good READMEs), but I'd like to split it before merging.

**Naming.** `@internal/proper-of-time` only holds date primitives and anchors, while the real `ProperOfTime` now lives in `@internal/generator`. Confusing. Something like `@internal/calendar-dates` fits better.

**Generator.** My main concern. `RomcalConfig` does [`new ProperOfTime(this)`](https://github.com/romcal/romcal/blob/dd8c73a189165dc3b71113f7f0c8eab7eb051dbd/packages/generator/src/models/config.ts#L192), and the package hardcodes Ordinary Time, the 1969 seasons, the precedence table, etc. So it's really the 1969 engine renamed, not a shared one. And 1962 doesn't fit that shape: it ranks by class (I to IV) instead of the 1969 precedence table, has no Ordinary Time but Septuagesima and Sundays after Pentecost, keeps octaves and vigils, and resolves occurrence through commemorations rather than by dropping the lower celebration. That's a different temporal calendar and a different occurrence engine, not a variant of the current one. As I wrote in https://github.com/romcal/romcal/pull/1127#issuecomment-5400731916, I'd rather have separate generators per rite and only share what's genuinely rite-neutral (dates, i18n infra, identifiers), even if that means some duplication. So I'd leave the generator extraction out of this PR for now.

**i18next fix.** The fix for #1226 is a behaviour change inside a refactor meant to be byte-identical. I would open it as its own PR: easier to review and to land on its own.

Suggested split:
1. calendar-dates (primitives + neutral anchors)
2. builder
3. i18next fix
4. generator: not for now, see above

On the overall direction, splitting into more packages is the right move. `rites/roman1969` currently mixes date maths, engine, data and build scripts, and separating those domain layers gives each part a small explicit API, makes the dependency direction visible (data > engine > dates), and shows clearly what a second rite really needs versus what is 1969-specific. I'd just cut along those layers rather than along a "shared vs rite" line.

TobiTenno:
i disagree wholeheartedly with the generator sitting per-rite. it's not calendar data, it's package generation.

Additionally, the generator is almost entirely 1969 for now because there hasn't been any difference to split them yet. i'm still working on that for the next part, but i'll split out out more fully here instead of later.

TobiTenno:
Split into the stack above, per review: #1226 fix, CI checks, calendar-dates, generator, builder.

PR #1225, opened 2026-08-29
feat(package): scaffold @internal/builder

TobiTenno:

Scaffolds `@internal/builder`, the package that will own the build, bundle, check,
document and publish tooling for the romcal rites.

That logic lives in `rites/roman1969/build/` today and only knows how to build one
rite: the paths back to the repository root are hardcoded, and the calendar and
locale globs assume that rite's layout. A second rite would leave two bad options,
copying the scripts into `rites/roman1962` or having one rite reach into a
sibling's build directory. Moving it out before `roman1962` exists avoids both.

Nothing moves yet. This PR adds the manifest, tsconfig, jest config, the
`RiteBuildManifest` contract, the CLI argument surface, a logger, and a README
with the file-by-file move table, wired into the root `clean` script, the
`tsconfig.base.json` path alias and both ESLint project arrays. The commands then
move one per commit, with byte parity of `dist/` as the acceptance test.

The builder contains no rite name. Everything rite-specific comes from a
`romcal.build.ts` manifest at the rite root, and CLI flags override what it
declares. That override is the point: rebuilding one calendar in one locale during
development currently means rebuilding all 104 across 13 locales.

The scripts being moved here each grew their own logging conventions: two symbol
vocabularies, three ways of indenting a detail line, and a `hasWarnings` flag in
`data-checks.ts` set as a side effect of formatting a label. A call site now picks
a severity and consola handles the presentation, choosing a reporter to suit the
output. The wrapper adds only what consola has no opinion about: a `--dry-run` tag,
nesting, and collapsible sections on GitHub Actions, which matters when 104 bundles
each log a line. Warnings and errors go to stderr, so redirecting stdout to a file
keeps problems visible. Worth flagging for review: `chalk` is imported by all six
build scripts while declared in no manifest at all, so it has been resolving
transitively. It is now a real dependency of this package.

None of this reaches a published artifact. The builder runs in Node under `tsx`,
while the `cjs`, `esm` and `iife` bundles are esbuild's output from a rite's `src`
— grepping a built `dist/` for `chalk`, `esbuild` or `glob` finds nothing,
including in `dist/iife/romcal.js`. A new ESLint rule keeps it that way:
`rites/*/src` cannot import `@internal/builder`, because that would pull the
toolchain into what consumers download. The rule was verified to fire against a
throwaway file rather than assumed to.

Lint, `tsc --noEmit` and the build are clean, and the full suite passes at 502
tests and 6 snapshots, including 7 new logger tests. `packages/proper-of-time` is
the same groundwork for the shared date logic, in a sibling PR.

## Concept

- Infrastructure

## Example code

No consumer-visible change. Internally, the CLI narrows a build and the commands
share one logger:

```typescript
// tsx packages/builder/src/cli.ts build --rite roman1969 --calendars france --locales fr

const log = createLogger({ verbose, dryRun });

log.group('bundles', (inner) => {
  inner.success('france');
});
```

### DIFF
--- a/.config/eslint.config.mjs
+++ b/.config/eslint.config.mjs
@@ -195,6 +195,7 @@ export default [
         parser: '@typescript-eslint/parser',
         project: [
           './tsconfig.base.json',
+          './packages/builder/tsconfig.json',
           './packages/easter/tsconfig.json',
           './packages/lunar-new-year/tsconfig.json',
           './rites/roman1962/tsconfig.json',
@@ -226,6 +227,7 @@ export default [
         parser: '@typescript-eslint/parser',
         project: [
           './tsconfig.base.json',
+          './packages/builder/tsconfig.json',
           './packages/easter/tsconfig.json',
           './packages/lunar-new-year/tsconfig.json',
           './rites/roman1962/tsconfig.json',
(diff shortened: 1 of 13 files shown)


TobiTenno:
this is moving into #1224

PR #1229, opened 2026-08-30
fix(l10n): register the interpolation formats with i18next's formatter

### LINKED ISSUE #1226: [Bug]: i18next 26 discards the custom interpolation formatters, so romanize/capitalize/uppercase are inert

TobiTenno:

Custom i18next interpolation formats have been silently ignored, so every localised
name that relies on one renders its raw value.

i18next 26 reads `interpolation.format` and then overwrites it with its own formatter
service during init. romcal declares `romanize`, `uppercase` and `capitalize` there,
so all three were dropped, and nothing failed loudly: the interpolation just passed
the value through.

Registering each format with the formatter service is how a custom format is declared
now. It is also addressable rather than positional, so an unknown `{{value, whatever}}`
warns instead of falling through a chain of comparisons.

Split out of #1224 at review request, since it is a behaviour change and that PR is a
refactor. Closes #1226.

## Concept

- Localization

## Example code

```typescript
const calendar = await new Romcal({ localizedCalendar: Italy_It }).generateCalendar(2024);
calendar['2024-12-02'][0].name;

// before: 'lunedì della 1 settimana di Avvento'
// after:  'Lunedì della I settimana di Avvento'
```

Two tests cover it, one for `romanize` with `capitalize` on an Italian weekday and one
for `capitalize` on an English ordinal. Both fail without the fix.


TobiTenno:
@emagnier here's the new root of the stack. i still disagree with the generator part, but we can talk about it on #1232

rites/roman1969/__tests__/locales.test.ts
```diff
@@ -47,6 +49,21 @@ describe('Testing localization functionality', () => {
...
+  test('`romanize` and `capitalize` are applied to an Italian weekday', async () => {
+    const calendar = await new Romcal({ localizedCalendar: Italy_It }).generateCalendar(2024);
+    expect(calendar['2024-12-02'][0].name).toBe('Lunedì della I settimana di Avvento');
+  });
+
+  test('`capitalize` is applied to an English ordinal', async () => {
+    const calendar = await new Romcal({ localizedCalendar: GeneralRoman_En }).generateCalendar(2024);
+    expect(calendar['2024-12-01'][0].name).toBe('First Sunday of Advent');
+  });
+});
```
emagnier, line 77: can we add a small case for uppercase too?

emagnier (approved):
Just a nitpick, otherwise it's good to me.

> @emagnier here's the new root of the stack. i still disagree with the generator part, but we can talk about it on #1232

For the disagreement, I answered here: https://github.com/romcal/romcal/pull/1232#issuecomment-5477385543

rites/roman1969/__tests__/locales.test.ts
TobiTenno: Added — no locale interpolates `uppercase` into a celebration name, so this hits the nested weekday lookup that format is for (`lunedì` → `LUNEDÌ`).

PR #1231, opened 2026-08-30
refactor(package): move the date primitives into @internal/calendar-dates

TobiTenno:

First of the four PRs #1224 was split into. It moves the date arithmetic and the
liturgical year anchors into `@internal/calendar-dates`.

These are the one part of a Roman Rite calendar that no rubric touches: a Sunday is
seven days after a Sunday under any missal, and Easter is computed the same way for
1962 as for 1969. A second rite should reuse them rather than copy them.

Named `calendar-dates` rather than `proper-of-time` per review. The package holds date
primitives and neutral anchors; the `ProperOfTime` pipeline is a different thing and
is not in it.

What stays with the rite is everything that depends on an option or on a rubric:

| Moved | Stays with the rite |
| --- | --- |
| `addDays`, `subtractsDays`, `startOfWeek`, `rangeOfDays` and the other primitives | `epiphany` and `baptismOfTheLord`, which depend on movable-Sunday options |
| `easterSunday`, `firstSundayOfAdvent`, `christmas`, `ashWednesday`, `palmSunday`, `pentecostSunday` | the `ascension` and `corpusChristi` transfer options |
| the fixed solemnities (`annunciation`, `assumption`, `allSaints`, …) | `holyFamily`, `divineMercySunday`, `christTheKingSunday` |
| `lunarNewYear`, `sundayOnOrAfterLunarNewYear` | the `temporalOverrides` layer added by #1223 |
| | `startOfSeasons` / `endOfSeasons`, keyed by the rite's own seasons |

Anchors are pure `(year: number) => Date` functions and take no configuration, because
anything configurable is by definition rite-shaped. The one argument any of them takes
beyond the year is the Easter calculation type, which is a choice of computation rather
than of rubric, and which both rites need.

`temporalOverrides` survives the split because it is applied *after* an anchor is
computed: `Dates#epiphany` works out the base date and then passes it through
`#applyAnchorExceptions`. The base computation moved; the exception layer stayed.

## Concept

- Infrastructure

## Example code

```typescript
import { easterSunday, addDays } from '@internal/calendar-dates';

// A pure function of the year, with no config and no rite.
easterSunday(2024);                    // 2024-03-31
easterSunday(2024, 'julian');          // 2024-05-05
addDays(easterSunday(2024), 39);       // Ascension Thursday, if the rite fixes it there
```

The rite's `utils/dates.ts` re-exports the primitives, since they are part of romcal's
public surface and are imported from that path across the codebase. The `Dates` class
keeps its per-year memoisation and its year defaulting, and delegates the computation.

Verified: `npm ci`, lint, build and the full suite green, and `rites/roman1969/dist` is
byte-identical to dev's — all 107 calendar bundles and every emitted `.js`. The 13
anchor tests in the new package now run in `test:without-coverage`, which was pinned to
the rite workspace and would otherwise have skipped them.


packages/calendar-dates/src/anchors.ts
```diff
@@ -0,0 +1,149 @@
...
+export const easterSunday = (year: number, easterCalculationType: EasterCalculationType = 'gregorian'): Date => {
+  const id = `${year}_${easterCalculationType}`;
+  if (easterCache[id]) return easterCache[id];
+
+  const { day, month } =
+    easterCalculationType === 'gregorian'
+      ? calculateGregorianEasterDate(year)
+      : calculateJulianEasterDateToGregorianDate(year);
+
+  return (easterCache[id] = getUtcDate(year, month, day));
```
emagnier: This returns the cached `Date` object itself. The cache is now module-global, so a consumer mutating a returned Easter date corrupts every later Romcal instance in the process. Before, the blast radius was one `Dates` instance. Cache the epoch millis and return `new Date(cached)`.

packages/generator/src/utils/dates.ts
```diff
@@ -662,11 +578,7 @@ export class Dates {
     easterCalculationType: EasterCalculationType = this.#config.easterCalculationType
   ): Date => {
     if (this.#easter[year]) return this.#easter[year];
-    const { day, month } =
-      easterCalculationType === 'gregorian'
-        ? calculateGregorianEasterDate(year)
-        : calculateJulianEasterDateToGregorianDate(year);
-    return (this.#easter[year] = getUtcDate(year, month, day));
+    return (this.#easter[year] = anchors.easterSunday(year, easterCalculationType));
```
emagnier: `#easter` is keyed only by year even though this method accepts an explicit calculation type. The refactor now mixes methods that use this cache with methods that call `anchors.*` using `#config.easterCalculationType`. On a Gregorian instance, calling `easterSunday(2024, 'julian')` first makes `ascension()` use Julian Easter while `pentecostSunday()` uses Gregorian Easter, so Pentecost falls before Ascension. Remove this local cache and delegate directly to it.

packages/calendar-dates/src/anchors.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return date;
+};
+
+/** December 8, moved to the following Monday when it falls on a Sunday. */
+export const immaculateConceptionOfMary = (year: number): Date => {
+  const date = getUtcDate(year, 12, 8);
+  return date.getUTCDay() === 0 ? addDays(date, 1) : date;
+};
```
emagnier: These shared anchors should only return March 25 and December 8. We should then use `dateExceptions` in the 1969 Roman calendar definitions (following the existing Saint Joseph pattern). Keeping the transfers here would give a Roman 1962 calendar the wrong date for the Immaculate Conception.

emagnier: Actually, these should be defined entirely in the GRC (`rites/roman1969/src/calendars/general-roman/index.ts`), with transfers handled by `dateExceptions`. No shared anchors are needed here. `anchors.ts` can stay focused on reusable dates that other calculations depend on, or calculations complex enough to share.

packages/calendar-dates/src/primitives.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * For a given date, get the ISO week number
+ *
+ * @remarks This function was sourced from [StackOverflow](https://stackoverflow.com/a/6117889/3408342).
+ *
+ * @param    date - Date which should be considered
+ * @returns  ISO week number of the specified date
+ */
+export const getWeekNumber = (date: Date): number => {
+  const d = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
```
emagnier, line 73: Hmm, an existing bug that’s worth fixing while we're at it.

This uses local-time getters even though the file requires timezone-invariant calculations. West of UTC, a date around New Year can be assigned to the wrong ISO week. Use `getUTCFullYear()`, `getUTCMonth()` and `getUTCDate()` instead.

emagnier (changes requested):
My review below, plus two minor follow-ups:
- `EasterCalculationType` is now duplicated between `calendar-dates` and `roman1969`. It would be better to import and re-export the shared type.
- `packages/calendar-dates/README.md` says anchors only take a year, except for the Easter calculation type, but `lunarNewYear` also takes a UTC offset. `That` on line 31 is also unclear.

packages/calendar-dates/src/primitives.ts
TobiTenno: ok, i'll do that in a distinct PR that'll be slotted after the draft

TobiTenno: added as #1238

packages/calendar-dates/src/anchors.ts
TobiTenno: fixed on [ab94f7f](https://github.com/romcal/romcal/pull/1231/commits/ab94f7f0d4b5d8ab010abdb51aa34bb297db7824)

packages/generator/src/utils/dates.ts
TobiTenno: fixed on [ab94f7f](https://github.com/romcal/romcal/pull/1231/commits/ab94f7f0d4b5d8ab010abdb51aa34bb297db7824)

packages/calendar-dates/src/anchors.ts
TobiTenno: fixed on [ab94f7f](https://github.com/romcal/romcal/pull/1231/commits/ab94f7f0d4b5d8ab010abdb51aa34bb297db7824)

emagnier:
Hello @TobiTenno, sorry for the delay. I can’t be as responsive here as I’d like. I’ll re-review again this PR and the others ASAP.

rites/roman1969/src/models/liturgical-day-config.ts
```diff
@@ -103,12 +103,16 @@ export class LiturgicalDayConfig implements BaseLiturgicalDayConfig {
     let updatedDate: Date | null = date;
 
     const setDate = (dateDefExtended: DateDefExtended): void => {
-      if (isInteger(dateDefExtended.addDay)) {
+      // Prefer a full date lookup so `{ dateFn, addDay }` resolves against the
+      // named date, not the original. Bare `{ addDay }` / `{ subtractDay }`
+      // still offset the originally computed date.
+      const lookedUp = this.#dateLookup(dateDefExtended, yearOffset);
```
emagnier: This is a real bug fix, not just a refactor detail 👍 
I haven’t checked whether this bug affected existing outputs, but if it did, just note that the refactor won’t produce bitwise-identical results.

rites/roman1969/src/models/liturgical-day-config.ts
```diff
@@ -103,12 +103,16 @@ export class LiturgicalDayConfig implements BaseLiturgicalDayConfig {
...
+      if (lookedUp) {
```
emagnier: Related: when `dateFn` does not match a method on `Dates`, `#dateLookup` returns `null`. With the new fallback, a typo can silently keep the original date or apply the offset to it.
Could `#dateLookup` throw when `dateFn` is a string that does not exist on `Dates`? TypeScript catches this in typed definitions, but untyped input can still reach it.

rites/roman1969/__tests__/dates.test.ts
```diff
@@ -217,6 +217,12 @@ describe('Testing specific liturgical date functions', () => {
         expect(rangeContainsDate(range, easter)).toBeTruthy();
       }
     });
+
+    test('An explicit Julian lookup does not leak into other dates on a Gregorian instance', () => {
+      const dates = new Romcal().dates(2024);
+      dates.easterSunday(2024, 'julian');
+      expect(dates.pentecostSunday().getTime()).toBeGreaterThan(dates.ascension().getTime());
+    });
```
emagnier, line 227: This only checks the relative ordering, so it would still pass if both dates drifted to Julian together.
Please assert the absolute Gregorian dates after the explicit Julian lookup: `easterSunday()` should be 2024-03-31 and `pentecostSunday()` should be 2024-05-19.

rites/roman1969/__tests__/dates.test.ts
```diff
@@ -764,96 +770,35 @@ describe('Testing specific liturgical date functions', () => {
...
-          // all of the above condition are not met... that is tested in another use case
-          expect(true).toBeTruthy();
-        }
+    test('The Dates helper is always 25 March; transfers live on the GRC', () => {
+      for (let i = 2020; i <= 2030; i += 1) {
+        const date = new Romcal().dates().annunciation(i);
+        expect(date.getUTCMonth()).toEqual(2);
+        expect(date.getUTCDate()).toEqual(25);
       }
     });
```
emagnier: The 2024 test covers a date inside the transfer window, but not the inclusive boundaries.
Please also cover 2018, when March 25 was Palm Sunday, and 2035, when March 25 will be Easter Sunday. These two cases verify both ends of the `inclusive: true` rule.

rites/roman1969/src/utils/dates.ts
```diff
@@ -932,33 +839,22 @@ export class Dates {
...
    * Get the date of the celebration of the Annunciation
    *
-   * *Occurs on March 25th, moved to Monday after Divine Mercy Sunday
-   * if it is within Holy Week or Easter Octave*
+   * Occurs on March 25. The 1969 rite transfers it out of Holy Week and the
+   * Octave of Easter via `dateExceptions` on the General Roman Calendar.
    *
    * @param year Gregorian year
    */
   annunciation = (year = this.#year): Date => {
```
emagnier: `dates().annunciation()` and `dates().immaculateConceptionOfMary()` used to return the transferred celebration date. They now return only March 25 and December 8, even when the calendar celebrates them on another day.
Nothing in `src` references these methods any more now that the GRC owns the fixed dates and transfers through `{ month, date }` and `dateExceptions`. Since romcal v3 has not had a final release yet, I think it's the right time to remove both helpers rather than keep an ambiguous public API.

rites/roman1969/src/calendars/general-roman/index.ts
```diff
@@ -319,8 +319,17 @@ export class GeneralRoman extends CalendarDef {
     // src: mr_la_2008_ed3
     annunciation_of_the_lord: {
       precedence: Precedences.GeneralSolemnity_3,
-      // 03-25
-      dateDef: { dateFn: 'annunciation' },
```
emagnier, line 323: As we remove those two helpers, the same cleanup seems applicable to the other fixed-date helpers that only proxy `getUtcDate`:
`presentationOfTheLord`, `nativityOfJohnTheBaptist`, `peterAndPaulApostles`, `transfiguration`, `assumption`, `exaltationOfTheHolyCross`, and `allSaints`.
Each currently goes through `GRC dateFn -> Dates wrapper -> calendar-dates anchor -> getUtcDate`, without adding any calculation. Could we define them directly with `{ month, date }` and remove the corresponding wrappers and anchors?
I would keep structural anchors such as `christmas`, `maryMotherOfGod`, and `divineMercySunday`, since other calculations and date exceptions depend on them.

rites/roman1969/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@internal/easter": "file:../../packages/easter",
     "@internal/lunar-new-year": "file:../../packages/lunar-new-year",
+    "@internal/calendar-dates": "file:../../packages/calendar-dates",
```
emagnier: The dependencies are not sorted. `npx sort-package-json --check rites/roman1969/package.json` fails because `@internal/calendar-dates` should come before `@internal/easter`.

packages/calendar-dates/src/anchors.ts
```diff
@@ -0,0 +1,129 @@
...
+import { addDays, getUtcDate, subtractsDays } from './primitives';
+
+/**
+ * The points in the year that every Roman Rite calendar is measured from.
+ *
+ * Each anchor is a pure `(year: number) => Date`. Nothing here reads configuration,
+ * because anything configurable is by definition rite-shaped: Epiphany's
+ * movable-Sunday rule, the Ascension and Corpus Christi transfers, and the
+ * `temporalOverrides` exceptions all stay with the rite, applied on top of the
+ * dates computed here.
```
emagnier: **Nitpick**: This still says that every anchor is `(year: number) => Date`.
`easterSunday` and its derived anchors also take a calculation type, while `lunarNewYear` takes a UTC offset. It would be more accurate to say that anchors are pure and configuration-free, although some take an explicit calculation parameter.

rites/roman1969/src/models/liturgical-day-config.ts
TobiTenno: got this cleaned up, throwing an error when there's a bad dateFn now.

rites/roman1969/__tests__/dates.test.ts
TobiTenno: done, pinned.

rites/roman1969/package.json
TobiTenno: it's sorted now.

packages/calendar-dates/src/anchors.ts
TobiTenno: sure, docs updated.

rites/roman1969/src/calendars/general-roman/index.ts
TobiTenno: yup, done.

TobiTenno:
@emagnier looking forward to further thoughts

emagnier:
yes, I’ve been busy these past few days. I will pick it up again starting tomorrow.

rites/roman1969/src/models/liturgical-day-config.ts
emagnier: That fixes the unknown `dateFn` case. There’s still a separate case: a valid function can return `null` when the requested day does not exist. For example, with Ascension on Thursday, `setDate: { dateFn: 'weekdayOrSundayOfEasterTime', dateArgs: [4, 6] }` returns `null` because that Thursday is Ascension. The new fallback keeps the original date instead.

Can you handle bare `addDay`/`subtractDay` separately from a full date lookup, keep `null` when the lookup returns it, and add a test for that case?

emagnier (changes requested):

rites/roman1969/src/models/liturgical-day-config.ts
TobiTenno: done — bare `addDay`/`subtractDay` handled separately; lookup `null` kept; covered in `config.test.ts`.

emagnier (approved):

Issue #59, opened 2019-02-12
Create `CODE_OF_CONDUCT.MD`


tukusejssirs:
I suggest to create `CODE_OF_CONDUCT.MD` and copy there the [Contributor Covenant v1.4](https://www.contributor-covenant.org/version/1/4/code-of-conduct.html). I read it and I think is good.

tukusejssirs:
Note that in the file, we need to place a repo email. I believe @pejulian used his email, though we might create separate one.

TobiTenno:
I don't have access to it, but there is the CI email, that could forward to each of us, esp if we're able to add it as a send alias for ourselves. I'll make a COC and use the CI email for now, and we can discuss on the PR

PR #1232, opened 2026-08-30
refactor(package): move the calendar engine into @internal/generator and open its rite seams

TobiTenno:

Second of the four PRs #1224 was split into. It moves the calendar engine into
`@internal/generator` — the models, types, utils and the Proper of Time pipeline — and
opens the seams that make it an engine for the Roman Rite rather than the 1969 engine
under a new name.

`rites/roman1969` mixed four things: date maths, engine, data and build scripts. The
dates left in the previous PR, the scripts leave in the next one, and this is the
engine. The dependency direction becomes visible and one-way: data → engine → dates.

The engine no longer names a rite. `RomcalConfig` imported `GeneralRoman` directly; it
now asks a registry, which the rite fills as it loads.

### The seams

The review on #1224 was that `@internal/generator` was the 1969 engine renamed, and that
1962 needs its own. That reading was correct of the state it saw: the package hardcoded
the 1969 precedence table, the 1969 seasons and Ordinary Time. Rather than duplicate the
pipeline, those rules now come from the rite:

| Seam | 1969 | 1962 |
| --- | --- | --- |
| precedence and ranks | the 27 categories of UNLY #59 | four classes, `Rubricae` §91 |
| seasons and their numbering | Advent → Ordinary Time, with the Lenten shift and the late-OT wrap | Advent → Time after Pentecost, counted straight through |
| dates | romcal's `Dates` | its own, with Septuagesima, Passiontide, Ember and Rogation days |

Each is a contract on the config that defaults to the 1969 norms, so a rite with no
opinion behaves exactly as romcal always has. The special cases moved with them: the
Lenten and late-Ordinary-Time week numbering left `Calendar` for the 1969 rubrics, where
they belong, and the two period rules that read a date by name left `LiturgicalDay`.

### The vocabulary

Opening those seams first widened `Precedence` and `Rank` to `string`, which was the
wrong answer: it loosened the public types for everyone in order to let one rite name
its own classes.

Instead, the vocabulary a rite celebrates by — its precedences, ranks, seasons and
periods — is now a type parameter threaded through every engine type that carries one.
`Vocabulary` is an ordered utility type, so a rite overrides only what it names and the
engine passes a single argument down. The rite pins it at its public methods, and
`Precedence` and `Rank` are closed unions again.

The result is that a rite's rules are checked in the rite's own vocabulary: a 1962 rank
is not assignable where a 1969 one is wanted, which is how the mistyped 1962 rubrics on
the branch above this one were caught.

Still to come, in the branch above: a rite-supplied occurrence policy for 1962's
commemorations, forward transfer, vigil suppression and octaves, and a seam for a rite's
own Proper of Time. Those two are where the "different occurrence engine" objection
really bites, and they can be judged there rather than assumed here. Discussion of the
direction belongs in #1127.

## Concept

- Infrastructure
- Discussion

## Example code

```typescript
// A rite supplies its rules and its dates together, and the vocabulary they are
// written in. Nothing here is exported by the engine.
const Rubricae1960Rite: Rite<Roman1962Vocabulary> = {
  dates: Dates1962,
  rubrics: Rubricae1960Rubrics,
};

// Consumers are unaffected. A generated day still reports the closed 1969 unions:
const day = (await new Romcal().generateCalendar(2024))['2024-12-25'][0];
const rank: Rank = day.rank; // 'SOLEMNITY' | 'SUNDAY' | ... , not string
```

Verified: lint, build, typecheck and the full suite green, 526 tests across the
workspaces. A type probe against the built `dist` confirms consumers still resolve the
closed unions, and that a rank, precedence or season from outside the 1969 vocabulary is
rejected.

On output parity: 5,246 of the 5,253 files in `rites/roman1969/dist` are byte-identical
to dev's, including all 107 calendar bundles. The three top-level bundles differ by 786
bytes, from the one runtime module the rite's pinned `CalendarDef` adds, and by the
minifier's identifier allocation shifting with it. Generating every calendar from both
builds — 14 calendars, 5,105 days — gives identical output, apart from `fromCalendarId`,
which is derived from `constructor.name` and so differs between any two minified builds.


emagnier:
Only skimmed so far, will do a proper review as soon as I can (and the `Vocabulary` approach is interesting!).

> i still disagree with the generator part, but we can talk about it on [#1232](https://github.com/romcal/romcal/pull/1232)

On the generator, I think I was unclear: by "separate generators" I meant the occurrence logic (which celebration wins a date, what gets commemorated or transferred, octaves, vigils...), not the orchestration (walking the year, joining temporal and sanctoral, config, i18n, bundling...). Sharing the orchestration is fine, and this PR starts moving the 1969 rules out of the engine (precedence, seasons, dates), which was my real objection on #1224. If your disagreement was about something else, tell me what I missed.

The real test for 1 engine vs 2 will be the occurrence policy work you mention (in the generated output, 1969's per-date array lists alternatives, 1962's would be the Mass plus its commemorations), so let's settle it there.

TobiTenno:
My disagreement was in the mistaken belief that you meant everything you just said you didn't mean, so as long as you're ok with the pattern i'm going with here (where as much as possible for rules lives in the rite itself, and the tooling stays distinct), I completely agree.

emagnier:
Yes, that's exactly the pattern I had in mind: rules in the rite, tooling shared.

TobiTenno:
i screwed up the rebase here, @emagnier, but i got it cleaned up.

PR #1233, opened 2026-08-30
refactor(package): move the build scripts into @internal/builder

TobiTenno:

Last of the four PRs #1224 was split into. The seven scripts under
`rites/roman1969/build` become commands behind a `romcal-build` CLI in
`@internal/builder`.

It sits after the generator extraction because the scripts use engine APIs
(`CalendarDef`, `Locale`, `MartyrologyCatalog`, `toPackageName`, `toPascalCase`). Landing
it first would mean importing those from `@internal/rite-roman1969`, which would cost
the builder the property that makes it worth extracting: it names no rite.

Everything rite-specific is read from a `romcal.build.ts` manifest the rite owns —
paths, output formats, and the calendars, locales and martyrology themselves.

`--calendars` and `--locales` narrow a build to one calendar rather than all 107, which
takes a France-only rebuild from minutes to a couple of seconds. That is the part that
changes day-to-day work on a single calendar.

`chalk` gives way to a `consola`-based logger: severity-based, problems to stderr,
dry-run tagging, and collapsible sections on GitHub Actions. An ESLint rule keeps
`rites/*/src` from importing the tooling, since a rite's source is bundled into cjs, esm
and iife for consumers and would drag the whole toolchain in with it. The build
toolchain moves to the package that uses it, leaving the rite four devDependencies.

## Concept

- Infrastructure

## Example code

```bash
# All 107 calendars, as before.
romcal-build build --rite roman1969

# Or just the one being worked on.
romcal-build build --rite roman1969 --calendars France --locales Fr

romcal-build check --rite roman1969
romcal-build doc   --rite roman1969
romcal-build trust --rite roman1969 --sync
```

```typescript
// rites/roman1969/romcal.build.ts — everything the builder needs to know about a rite.
export default {
  calendars: calendarDefinitions,
  locales,
  martyrology,
  formats: ['cjs', 'esm', 'iife'],
} satisfies RiteBuildManifest;
```

Verified: `npm ci`, lint, build, the full suite, `data:checks`, the docs checks and a
narrowed `--calendars France` build all green, and `rites/roman1969/dist` is
byte-identical to dev's through this and the three PRs beneath it. The logger has 7
tests of its own.


packages/builder/src/commands/publish.ts
```diff
@@ -47,16 +43,25 @@ const publish = (target: Target): boolean => {
...
   const bundleDirs = fs
     .readdirSync(bundlesBasePath, { withFileTypes: true })
     .filter((dirent) => dirent.isDirectory())
     .map((dirent) => path.join(bundlesBasePath, dirent.name));
 
-  const targets: Target[] = [rootDir, ...bundleDirs].map((dir) => {
+  const targets: Target[] = [repoRoot, ...bundleDirs].map((dir) => {
     const { name, version } = JSON.parse(fs.readFileSync(path.join(dir, 'package.json'), 'utf-8'));
     return { dir, name, version };
   });
```
emagnier, line 71: With a full build on disk, `publish --calendars France` still attempts to publish every calendar package. The CLI accepts and validates the filter, but this list ignores it. Filter the targets, or reject `--calendars` for this command instead of silently publishing outside the requested selection.

packages/builder/src/commands/build.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+
+  if (dryRun) return;
+
+  /**
+   * Delete and recreate empty dist directory
+   */
+  log.step(`Cleaning up the ${manifest.outDir}/ directory`);
+  rimraf.sync(distDir);
```
emagnier: This deletes the previous build even when only one stage is selected. After a successful build, `build --emit packages` removes `dist` and then fails with ENOENT when writing the calendar package.json files. Partial runs need to keep the outputs they depend on and only replace the files they regenerate.

packages/builder/src/commands/bundle.ts
```diff
@@ -96,8 +97,15 @@ export class RomcalBuilder {
...
-  rimraf.sync(resolve('tmp/bundles'));
+export const RomcalBundler = (options: ResolvedOptions, log: Logger): void => {
+  const { manifest, riteRoot } = options;
+  const bundlesDir = resolve(riteRoot, manifest.tmpDir, 'bundles');
+
+  // The rite's own calendars inherit from its base calendar, which the engine only
+  // knows about once the rite registers it.
+  registerBaseCalendar(manifest.baseCalendar);
+
+  rimraf.sync(bundlesDir);
```
emagnier, line 118: `bundle --dry-run` still deletes `tmp/bundles` and writes new files. The dry-run check in `runBuild` does not cover the standalone `bundle` command. We should check `options.dryRun` before removing or writing anything here.

packages/builder/src/commands/build.ts
```diff
@@ -0,0 +1,302 @@
...
+      const pkgName = toPackageName(calendar);
+
+      const dir = join(distDir, 'bundles', pkgName);
+
+      const modulePkg: PackageJson = {
+        name: manifest.packageNameTemplate.replace('[calendar]', pkgName),
+        version: pkg.version,
+        description: `Localized romcal calendar for ${calendar}`,
+        module: './esm/index.js',
+        main: './cjs/index.js',
```
emagnier: With `--formats esm`, `main` and `exports.require` still point to `./cjs/index.js`, which isn't generated. These entries should match the formats actually built, or packaging should reject incompatible selections.

Also, I'd like to simplify this and support ESM only. Node 22.13+ can load ESM through `require()` without a flag or warning, as long as the module and its dependencies don't use top-level `await`. And modern browsers can load ESM with `<script type="module">`. Since v3 isn't final yet, this seems like a good time to drop both CJS and IIFE.

packages/builder/src/options.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Turning `process.argv` into a resolved set of options.
+ *
+ * Flags override the manifest, and anything not given falls back to everything the
+ * rite declares, so `romcal-build build` on its own behaves exactly like the old
+ * `npm run build`.
+ */
+
+const FORMATS: readonly BuildFormat[] = ['cjs', 'esm', 'iife'];
+const ARTIFACTS: readonly BuildArtifact[] = ['bundles', 'docs', 'packages', 'types'];
```
emagnier, line 13: `docs` is accepted by `--emit` and included by default, but `runBuild` never calls `runDoc`. `build --emit docs` exits successfully without updating the documentation. Wire up this stage, or remove it from `--emit` and keep documentation generation under the `doc` command.

packages/builder/src/options.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const [name, inline] = arg.slice(2).split('=');
+    if (inline !== undefined) {
+      flags[name] = inline;
+      continue;
+    }
+
+    const next = argv[i + 1];
+    // A flag either takes the next token or is a boolean switch.
+    flags[name] = next && !next.startsWith('-') ? next : true;
```
emagnier, line 64: `romcal-build --dry-run publish` consumes `publish` as the flag's value, then still selects it as the command. Since that value is a string, `dryRun` becomes false. `publish --dry-run=true` has the same problem. 
Parse boolean switches explicitly and reject unsupported values before running the command. A requested dry run must not silently become a real publish.

emagnier (changes requested):
Nice to see the build tools brought together under one CLI 👍

packages/builder/src/commands/build.ts
TobiTenno: do you think we should drop both? I wasn't really sure, since cjs can still have issues, and i didn't want this to be a horribly breaking set of changes. it would clean up our bundles a lot.

packages/builder/src/commands/bundle.ts
TobiTenno: fixed that 👍

packages/builder/src/commands/build.ts
TobiTenno: changed. i don't think the other was _wrong_, but this is more targeted.

packages/builder/src/commands/build.ts
emagnier: I think we can keep both for now to keep this PR focused. I'd still like us to revisit ESM-only soon though, and make a decision before the final v3 release at the latest.

packages/builder/src/options.ts
TobiTenno: Fixed in `98395747`: `runBuild` calls `runDoc` when `emit` includes `docs` (dry-run and real).

packages/builder/src/options.ts
TobiTenno: Fixed in `98395747`: `--dry-run` is a dedicated boolean flag (`BOOLEAN_FLAGS` + `parseBoolean`), so it no longer swallows the command name. Covered by `options.spec.ts`.

packages/builder/src/commands/bundle.ts
emagnier: Thanks for the fix! There is still a similar issue with `tmp/bundles`: after a full build, `build --calendars France --emit bundles` deletes the intermediate files for all other calendars. Running `build --emit packages` afterwards fails with ENOENT when copying their `index.d.ts` files. Could you limit cleanup in `RomcalBundler` to the selected calendars as well?

packages/builder/src/commands/build.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  /**
+   * Add package.json and index.d.ts files to all calendar bundles
+   */
+  if (emit.includes('packages')) {
+    log.step('Package calendar bundles as npm modules');
+    // The published calendars carry the version of `romcal` itself and take a peer
+    // dependency on it, so this is the root manifest rather than the rite's.
+    const pkg = JSON.parse(fs.readFileSync(join(repoRoot, 'package.json'), 'utf-8')) as RootManifest;
+    const entryPoints = packageEntryPoints(options.formats);
```
emagnier: After a clean `build --formats esm`, running `build --emit packages` uses the default formats and changes `main` and `exports.require` back to `./cjs/index.js`, which does not exist. Could you check that the requested entry points exist before writing the package metadata, or reject this combination with a clear error?

emagnier (changes requested):
Thanks for the fixes! Just 2 small points remaining

emagnier (approved):

PR #1236, opened 2026-09-01
refactor(package): move the calendar engine into @internal/generator and open its rite seams

TobiTenno:
Replacement for #1232. GitHub marked that PR merged when `reorg/calendar-dates` was accidentally fast-forwarded onto this branch; the commits here are the rebase onto the #1231 review fix.


packages/generator/src/models/liturgical-day.ts
```diff
@@ -19,17 +17,18 @@ import {
...
 
 import { CyclesMetadata } from './cycles-metadata';
 import { LiturgicalDayConfig } from './liturgical-day-config';
 import { LiturgicalDayDef } from './liturgical-day-def';
 
 const COMPUTED_FIELDS = ['name', 'colorNames', 'seasonNames', 'rankName', 'definition', 'config'];
 
-export class LiturgicalDay implements BaseLiturgicalDay {
-  readonly #liturgicalDayDef: LiturgicalDayDef;
+export class LiturgicalDay<V extends Vocabulary = Vocabulary> implements BaseLiturgicalDay<V> {
```
emagnier, line 28: The Roman entry point re-exports this class with Vocabulary as its default, so an explicitly typed LiturgicalDay now has a string rank. `const rank: Rank = day.rank` no longer compiles. The same affects precedence, seasons and periods. Specialize the public 1969 exports to Roman1969Vocabulary, as already done for CalendarDef and the Romcal method return types.

packages/generator/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "types": "dist/src/index.d.ts",
+  "scripts": {
+    "build": "tsc --project tsconfig.release.json",
+    "clean": "rimraf coverage dist",
+    "test": "TZ=UTC jest",
+    "test:without-coverage": "TZ=UTC jest",
+    "typecheck": "tsc --noEmit --importHelpers false -p tsconfig.json"
+  },
+  "dependencies": {
+    "@internal/calendar-dates": "file:../proper-of-time",
```
emagnier: This should point to file:../calendar-dates. packages/proper-of-time does not exist in this checkout. Regenerate the lockfile too: it still contains the old package entry and a nested dependency link pointing to that missing directory.

emagnier (changes requested):

packages/generator/package.json
TobiTenno: fixed

packages/generator/src/models/liturgical-day.ts
TobiTenno: yup, got them specialized

packages/generator/src/types/liturgical-day.ts
emagnier: This type also needs the vocabulary parameter. LiturgicalDayRoot now defaults to Vocabulary, so rank and precedence become string here, including through `day.fromExtendedCalendars`. You can propagate the vocabulary and specialize the public 1969 export as well.

emagnier (changes requested):

packages/generator/src/types/liturgical-day.ts
TobiTenno: done in [f174f3a](https://github.com/romcal/romcal/pull/1236/commits/f174f3abcc38c98f9c6d58fd7f40b938448ee713)

emagnier (approved):

